PR #1, opened 2020-02-24
Create vixenNetwork.yml

ueaslsef:


Leopere:
Good to see contributors, someone will have to test this soon.

ueaslsef:
Question for you: How do we organize scrapers? how shall they be named?

E.g. `xpathscrapers/vixennetwork/ueaslsefscraper.yml`
or `vixen_tushy_blacked_xpathonly.yml` ?

bnkai:
I think this need to be spit in two files.
For the tushy part i had to change  to this
`Details: //div[@data-test-component="VideoDescription"]//p`

vixenNetwork.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: "Vixen/Blacked/Tushy"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - "https://www.blackedraw.com/"
+      - "https://www.blacked.com/"
+      - "https://www.tushy.com/"
+      - "https://www.tushyraw.com/"
+      - "https://www.vixen.com/"
```
halorrr, line 10:
```suggestion
      - "https://www.vixen.com/"
      - "https://www.deeper.com/"
```

Deeper can be added to this list.

vixenNetwork.yml
```diff
@@ -0,0 +1,25 @@
+---
+name: "Vixen/Blacked/Tushy"
```
halorrr, line 2:
```suggestion
name: "Vixen Media Group"
```
Instead of the name of this scraper ever expanding this is probably more inclusive as the official name. Can maybe add a comment to the top of the tile though to the extent of:
```
# This currently includes Blacked, BlackedRaw, Tushy, TushyRaw,
# Vixen, and Deeper.
```

vixenNetwork.yml
```diff
@@ -0,0 +1,25 @@
...
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performer: //div[@data-test-component="VideoModels"]/a
+    scene:
+      Title: //h1[@data-test-component="VideoTitle"]
+      Performers:
+        Name: $performer
+        URL: $performer/@href
+      Details: //p[@data-test-component="VideoDescription"]
```
halorrr, line 21:
```suggestion
      Details: //div[@data-test-component="VideoDescription"]/p
```
This fixes it to pull the details on all the sites.

vixenNetwork.yml
```diff
@@ -0,0 +1,25 @@
...
+      Date:
+        selector: //span[@data-test-component="ReleaseDateFormatted"]
+        parseDate: "January 2, 2006"
```
halorrr, line 24:
```suggestion
      Date:
        selector: //span[@data-test-component="ReleaseDateFormatted"]
        parseDate: "January 2, 2006"
      Image: (//div[@data-test-component="ProgressiveImage"]/img/@src)[1]
```

This should scrape images as well

vixenNetwork.yml
bnkai: This is the same I posted above for tushy, tushy raw.

vixenNetwork.yml
bnkai: Image links are generated through javascript, the above xpath doesn't work.

vixenNetwork.yml
bnkai: `//p[@data-test-component="VideoDescription"]|//div[@data-test-component="VideoDescription"]//p`
covers both cases

vixenNetwork.yml
halorrr: Oh wow, the site doesn't load at all without javascript 
TUSHY : Exclusive Anal HD Porn Videos 2020-05-13 10-23-23

Is there a scrapping method that would work for images that are reliant on javascript?

bnkai: Have a look at the other scrapers.
First you identify the script section you need from the page source ( something like this `//script[contains(text(),'INITIAL_STATE')]` in this case) 
and then you regex replace from inside the url

halorrr: @bnkai When you get a chance can you try this one:
```
Image:
        selector: //script[contains(text(),'INITIAL_STATE')]
        replace:
          - regex: (?:{"name":"mainLandscape_1920x1080","width":1920,"height":1080,"src":")([^"]+)(?:.*)
          - with: $1
          - regex: \\u002F
          - with: /
```

halorrr: I've hit a wall in this one, I'll leave my progress here and maybe someone else can figure out the image field:

```
Image:
   selector: //script[contains(text(),'INITIAL_STATE')]/text()
   replace:
      - regex: .+(?:{"name":"mainLandscape_1920x1080","width":1920,"height":1080,"src":")(\w{5}:)(?:\\u002F\\u002F)([^\\u002F][\w-.]*)(?:\\u002F)([^\\u002F][\w]*)(?:\\u002F)([^\\u002F][\w]*)(?:\\u002F)([^\\u002F][\w]*)(?:\\u002F)([^\\u002F][\w]*)(?:\\u002F)([^\\u002F][\w-_.?=&]*).+
      with: $1//$2/$3/$4/$5/$6/$7
```

The `.+` on either end ensure the first image is always matched. In my initial testing these worked as the scenes I was testing on happened to have the main image listed first in their scripts. However on further testing I found it to be that images can appear in random quantity and order. If you were to remove the `.+` on either end you will see there are typically 1-7 matches for a scene with my regex. The main image and then any number of the suggested scenes on the page.

This is how the regex breaks down the matches:
unknown

I think the matches could be identified by the content in groups 4 or 7 as they seem to be more identifying, however I can't find data on the page that would confirm which ones of these are associated with the main scene.

bnkai:
Using CDP seems to get the images as intended
The final looking version looks like this
```yaml
name: "Vixen Media Group"
sceneByURL:
  - action: scrapeXPath
    url:
      - "https://www.blackedraw.com/"
      - "https://www.blacked.com/"
      - "https://www.deeper.com/"
      - "https://www.tushy.com/"
      - "https://www.tushyraw.com/"
      - "https://www.vixen.com/"
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $performer: //div[@data-test-component="VideoModels"]/a
    scene:
      Title: //h1[@data-test-component="VideoTitle"]
      Performers:
        Name: $performer
      Details: //p[@data-test-component="VideoDescription"]|//div[@data-test-component="VideoDescription"]//p
      Date:
        selector: //span[@data-test-component="ReleaseDateFormatted"]
        parseDate: "January 2, 2006"
      Image:
        selector: //img[@class="sc-1egln9q-1 irBaDM"]/@src
driver:
  useCDP: true
# Last Updated August 6, 2020
```

PR #10, opened 2020-04-14
Czechav

MrX292:
should work with all Czechav sites


bnkai:
Could you add some valid urls for sites you tested  instead of plain czech?
it is good to verify because for some sites some details differ eg image location or   title.
If necessary you can split to individual files.

MrX292:
added the urls.
worked with all Sites some just dont have tags or Details on the site.

PR #17, opened 2020-04-22
Add AEBN xpath scene scraper

whitty-username:
Should be working, title might need further testing.


bnkai:
From a first test title seems to need a little more tweaking indeed.
I had to change to 
`Title: //h1[@class="dts-section-page-heading-title"]|//div[@class="dts-section-page-heading-title"]/h1`
as in some scenes an extra div was added

bnkai:
AEBN uses `Sept` instead of `Sep` for September so an extra replace regex for the date is needed
```
- regex: "Sept"
  with: "Sep" 
```

whitty-username:
Thanks for your help @bnkai applied those changes and tested. Working without issue now.

PR #22, opened 2020-05-14
Add BangBros xPath scrapper

halorrr:


scrapers/BangBros.yml
```diff
@@ -0,0 +1,23 @@
...
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="ps-vdoHdd"]/h1/text()
+      Details: //div[@class="vdoDesc"]/text()
+      Tags:
+        Name:
+          selector: //div[@class="vdoTags"]/a/text()
+      Performers:
+        Name: //div[@class="vdoCast"]/a[2]
```
bnkai, line 16: why a[2] ? on scenes with multiple performers we want all the performers 
doesn't `//div[@class="vdoCast"]/a` work  ?

halorrr: Ah I took the wrong approach. In my original test there was only 1 model, forgot to test with multiple. There is a reason it starts at 2 though.

I'll change that to `//div[@class="vdoCast"]/a[position()>1]/text()`

The first `a` in that section is the website name. So in this case for example 
/a[1] = `Ass Parade`
/a[2] = `Canela Skin`
/a[3] = `Katrina Moreno`
/a[4] = `Jorge Garcia`
Anal Canela and Katrina Go Greek | Ass Parade | Bangbros 2020-05-14 17-52-55

The tags are in a separate area though

bnkai:
Is it worth it to add the Studio Name since its is available a[1] ?

halorrr:
Added the Studio Name in!

PR #26, opened 2020-05-15
Adds GloryHoleSwallows xPath scraper

halorrr:
I'd like some input on this one. I'm unclear why the date/image does not work in testing.

### DIFF
--- a/scrapers/GloryHoleSwallow.yml
+++ b/scrapers/GloryHoleSwallow.yml
@@ -1,0 +1,25 @@
+name: "GloryHoleSwallow"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://gloryholeswallow.com/newtour/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Details: //span[@class="latest_update_description"]/text()
+      Date:
+        selector: //ul/comment()[contains(.,'update_date')]
+        replace:
+          - regex: .+(?:<.+>)(.+)(?:<.+>)
+            with: $1
+        parseDate: Jan 02, 2006
+      Tags:
+        Name: //span[@class="tour_update_tags"]/a//font/text()
+      Image:
+        selector: (//div[@class="back"]/center/img/@src)[1]
+        replace:
+          - regex: ^
+            with: https://gloryholeswallow.com/newtour/
+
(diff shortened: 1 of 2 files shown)


scrapers/GloryHoleSwallow.yml
```diff
@@ -0,0 +1,25 @@
...
+          - regex: .+(?:<.+>)(.+)(?:<.+>)
+          - with: $1
+        parseDate: January 02, 2006
+      Tags:
+        Name: //span[@class="tour_update_tags"]/a//font/text()
+      Image:
+        selector: (//div[@class="back"]/center/img/@src)[1]
+        replace:
+          - regex: ^
+          - with: https://gloryholeswallow.com/newtour/
```
bnkai, line 23: you put an extra `-` to the `with:`
should be
`with: https://gloryholeswallow.com/newtour/`

bnkai:
For the date part i think it's a bug in the scraper code ( comment nodes don't have innertext so output html should be used instead ) that i'll open an issue for.
For the image part the `-` was the  issue

scrapers/GloryHoleSwallow.yml
halorrr: Ahhh there we go

halorrr:
Perfect, so only thing that doesn't work on this one is the date then (which is kinda a bonus feature anyway since the date isn't actually displayed visually on the pages, so it would be a welcome surprise that this scraper is actually able to pull a date once comments are fixed haha)

scrapers/GloryHoleSwallow.yml
```diff
@@ -0,0 +1,25 @@
...
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Details: //span[@class="latest_update_description"]/text()
+      Date:
+        selector: //ul/comment()[contains(.,'update_date')]
+        replace:
+          - regex: .+(?:<.+>)(.+)(?:<.+>)
+            with: $1
+        parseDate: January 02, 2006
```
bnkai, line 16: Dates used are in the 3 letter form
`January` should be switched to `Jan`

halorrr: Done!

halorrr:
Updated to include documentation as well

bnkai:
Changes needed to stash were merged. Merging this.

PR #27, opened 2020-05-15
Adds ManyVids xPath scrapper

halorrr:
ManyVids doesn't include the year in their dates, so the date scrapper will always end up spitting out dates like `0000-01-01` for example with a blank year. Should I leave the date scrapper in? Or remove it?


bnkai:
i would comment it out for future reference, there might be some changes in the future that support partial date

halorrr:
Perfect! Done and commented out!

PR #32, opened 2020-05-17
Add Reality Kings style xPath scraper

halorrr:
This is another mega scraper I've tested and it works with all sites listed in it. Only special configuration I had to make to make it support all was to have the Title look for `h1` or `h2` as all the sites were using one or the other with the same class name.

It was created out of my babes scraper (with the added ability now to grab studio name) and removed the need for a scene scraper inside of the mofos.yml


scrapers/Mofos.yml
```diff
@@ -1,49 +1,17 @@
...
     scraper: performerScraper
-sceneByURL:
-  - action: scrapeXPath
-    url: 
-      - mofos.com/scene
-    scraper: sceneScraper
 xPathScrapers:
   performerScraper:
     performer:
       Name: //h1
```
bnkai, line 10: Is there a use for a performer scraper only fetching the name ?
Since the main, scene functionality is in the RK mega scraper this one Mofos.yml should be removed alltogether.

halorrr: Honestly not sure, I wasn't sure if anyone was using this functionality. It looked like it was still being worked on with all the comments so I wasn't sure if I should leave it or remove it. Think I should remove it entirely then?

bnkai: Yeah, remove it. If anyone is interested they can make a new one only for the performers.

halorrr: Done!

halorrr:
An already big scraper just got bigger. Just found the sites under `Really Useful, Ltd.` are using the same template. Which nets us all the fakehub sites (faketaxi, etc) and such in this scraper

PR #33, opened 2020-05-17
Adds PlayboyTV xPath scraper

halorrr:
Still debugging the image on this one. Doesn't seem to be working.

### DIFF
--- a/scrapers/PlayboyTV.yml
+++ b/scrapers/PlayboyTV.yml
@@ -1,0 +1,29 @@
+name: "PlayboyTV"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - www.playboy.tv/watch/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h2[@class="title"]/text()
+      Date:
+        selector: //small[@class="subtitle"]/text()
+        replace:
+          - regex: .+(?:• Added on )(.+)
+            with: $1
+        # Date is not always present, but when it is, it is in a set format in
+        # the description field
+        parseDate: 02 January 2006
+      Details:
+        selector: //small[@class="subtitle"]/text()
+        replace:
+          - regex: (.+)(?:• Added on )(?:.+)
+            with: $1
+        # The field sometimes, has a description, sometimes has episode details.
+        # When it has episode details, the date can be stripped out.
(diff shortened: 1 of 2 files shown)


halorrr:
Post the issue, another javascript block. Looks like the script is hosted externally too, am I right in my assumption that it would not be possible to pull the image for this one then?

scrapers/PlayboyTV.yml
```diff
@@ -0,0 +1,29 @@
...
+        # the description field
+        parseDate: 02 January 2006
+      Details:
+        selector: //small[@class="subtitle"]/text()
+        replace:
+          - regex: (.+)(?:• Added on )(?:.+)
+            with: $1
+        # The field sometimes, has a description, sometimes has episode details.
+        # When it has episode details, the date can be stripped out.
+      Image: //div[@id="video"]/@poster
```
bnkai, line 26: shouldn't it be `//video[@id="video"]/@poster` ?

halorrr: No the id=video one is a div:
`//div[@id="video"]/@poster`

For video there is also this one:
`//video[@id="video_html5_api"]/@poster`

But neither seem to work cause the images are loaded via Javascript

bnkai:
When something doesn't work as it should always have a look at the page source directly.
From a few scenes i could find, that didn't force me to join i got something like this from source 
```
<video id="video" class="video-js vjs-16-9 vjs-big-play-centered" controls preload="none" style="width:100%;" poster="https://photo-ht.pbnetcdn.com/releases/05/72/57221/player-large.jpg">
```
thats the code that stash gets , we don't need to care about the js.
`//video[@id="video"]/@poster` for images works fine for me at least in the following scenes "stay-in-get-off/s01e10-inside-and-out-with-abigail-mac", "road-trip", "intimate-tales", "seduction-principles", "perception" ( add the ..../watch/ at the beginning).
Can you retest or provide some samples where it doesn't seem to work ?

To tell you truth given the almost 0 useful metadata from this site, i wouldn't even bother scraping any scenes from here.

halorrr:
You are right, looks like the source had things arranged differently than how it was when I was using the inspect tool. I tested and that works for the scenes I wasn't able to get before.

Definitely agree it isn't the most useful scraper though.

PR #39, opened 2020-05-18
Add mypervmom xPath scene scraper

bnkai:


halorrr:
Interesting how much different of a template this has from pervmom.com

bnkai:
Yeah a lot of the myxxxxx sites have different templates. I was "forced" to do this btw since the pervmom.com site has onle a few select scenes while the mypervmom seems to have a lot more. I think that might be true for the rest myxxxxxx sites from the same network.

halorrr:
Going on your theory I tested all of the paperstreetmedia domains with the word `my` in front to see what I could turn up.

Your script also works for:
- mybadmilfs.com
- mydaughterswap.com
- mysislovesme.com

Of note, the latest scene seems can sometimes be in non date format. i.e. `3 Days Ago`:
image

Saw this behaviour on both mypervmom.com and mysislovesme.com

I don't think there is a good way to handle that. But just noting it in case you can think of anything.

SCENE-SCRAPABLE.md
```diff
@@ -101,7 +101,8 @@ momslickteens.com|RealityKings.yml
 moneytalks.com|RealityKings.yml
 mrpov.com|FinishesTheJob.yml
 mylf.com|PaperStreetMedia.yml
-mylfdom.com|PaperStreetMedia.yml
+mylfdom.com|PaperStreetMedi
```
halorrr, line 104: Accidentally dropped the `a` in Media here

bnkai: nice catch. copy paste fail

bnkai:
Yeah i noticed that with the recent scenes. No easy way past that, except than revisiting to scrape after a week or so.
Will test the extra urls and update later, ty.

PR #40, opened 2020-05-19
Adds JulesJordan xPath scraper

halorrr:
Tested and the scraper works on all the sites.

~The only field that doesn't seem to work is the "Movie" field. This is my first time using that field though so I'm not sure if I'm missing something there.~

All working now

### DIFF
--- a/scrapers/JulesJordan.yml
+++ b/scrapers/JulesJordan.yml
@@ -1,0 +1,38 @@
+name: "JulesJordan"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.julesjordan.com/
+      - https://www.girlgirl.com/
+      - https://www.manuelferrara.com/
+      - https://www.theassfactory.com/
+      - https://www.spermswallowers.com/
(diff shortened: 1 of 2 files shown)


scrapers/JulesJordan.yml
```diff
@@ -0,0 +1,38 @@
...
+        selector: $details//div[@class="cell update_date"]/text()
+        parseDate: 01/02/2006
+      Details:
+        selector: $details//span[@class="update_description"]/text()
+        concat: " "
+      Tags:
+        Name: //span[@class="update_tags"]/a/text()
+      Performers:
+        Name: $details//span[@class="update_models"]/a/text()|$galinfo//span[@class="update_models"]/div/a/text()
+      Movie:
```
bnkai, line 28: its `Movies:` not `Movie:` thats why it wasn't processed.
With movies it works (provided you have the movie with name needed)

halorrr: Ah perfect! Updated!

PR #43, opened 2020-05-19
Adds NewSenstations Network Sites xPath scraper

halorrr:
Of note the `/tour` part of these links is important, which is why I left it in for the site list.

i.e.
`newsensations.com/tour_ns/` is completely different from `newsensations.com/tour_rs/` with how they have set things up (and thus why they need to be in separate scrapers)


bnkai:
Looks good merging.
As a side note after a few views instead of the cover image the ns join now image appears making image scraping not to work. It's a known limitation for this network, nothing can be done so no biggie.

PR #44, opened 2020-05-19
Adds HotWifeXXX xPath scraper

halorrr:


scrapers/HotWifeXXX.yml
```diff
@@ -0,0 +1,26 @@
...
+    scene:
+      Title: $info//h2/text()
+      Date:
+        selector: $info//div[@class="released2 trailerStarr"]/text()
+        replace:
+          - regex: (,.+)
+            with:
+        parseDate: 01/02/2006
+      Details: $info//div[@class="videoDetails"]/p/text()
+      Tags:
```
bnkai, line 20: since `Tags:` is not used it should be removed

halorrr: Fixed!

PR #4, opened 2020-03-23
Create 21sexturynetwork.yml

babycasio:

PR #49, opened 2020-05-21
Adds Gamma Entertainment xPath scraper

halorrr:
I was making an EvilAngel scraper when I realized the existing Wicked scraper could be merged in.

Posted in Discord looking to see when setting the URLS for a scene scraper, if there any way to add a wildcard to part of the URL? For instance:
`https://www.evilangel.com/en/video/`

Both the `https://www.evilangel.com/` and the `/video/` part are necessary to ensure we are scraping a scene page and not a movie page. However I don't care what language the page is in like `en` in my example.

I tired `https://www.evilangel.com/*/video/` but that didn't seem to work.

^ Exploring the above before committing to keeping en in the URL

Also still going thought the giant list of sites under the parent company. There are likely going to be quite a few scrapers for this one company, but I'm trying to keep all the ones of similar templates together.


bnkai:
Is there even a way to change the language? I don't see any language option in the site.
Also for the newline issue leave the replace as is ( "\n" ) since it works after a few code changes that are needed to stash ( backend + ui ). The nurumassage descriptions btw are more like an essay.

halorrr:
Hmmm you are right, looks like the `en` is a non-issue, I can't find a way to switch the language either. 

Sweet I'll leave the "\n" as is then

And lol I like to think that someone somewhere is very proud of the effort they put into those descriptions

bnkai:
From a quick look we can add 21sextury.com and close #4 if you can just add the performers from this site. They switched to a js for loading the performers ( search for window.context in the page source)

halorrr:
For anyone following this PR, I've finished the company brand list. This scraper already covers 102 sites in my testing.

Working on three remaining sites that all have the same issue in trying to pull the performer, as the performer is only listed in the JS:

- 21sextury.com
- puretaboo.com
- isthisreal.com

As much as I would prefer to get them all in this one scraper, I may end up needing to split them out into their own scraper

bnkai:
Do you want to split the sites that have performer issues to separate scrapers ?
If they need more regex replace then there is no need to do it for all other sites also that don't need it.

bnkai:
Lol talk about comments in sync...

halorrr:
I'm going to split the 3 remaining sites into a separate PR as they are blocked until we get some other post-processing options https://github.com/stashapp/stash/issues/573

bnkai:
@halorrr is this ok for merge now? I tried most sites and they look ok .

halorrr:
@bnkai Yup this one is all set

PR #52, opened 2020-05-22
Add 21naturals.com Scene Scraper

Cromseth:
Trying to upload an scene scraper for 21naturals.com.
I hope I did it right... first time at github.


halorrr:
Looks like you did it right, though 21naturals is actually included in this PR as well that I'm working on https://github.com/stashapp/CommunityScrapers/pull/49

bnkai:
Everything seems ok and from what i see it uses different selectors than #49, #4 .
I don't know though if it makes sense to have two different scrapers for the same network and how stash will handle this ( having to choose for a url from 2 different scrapers )

Cromseth:
I didn't realize it was another scraper on the way. No problem if this is refused, yours seems much more complete and it worked as a training for me XD.

I have 3 more sites to keep and eye to... but when the info is embebbed in "json" stuff I am really lost... LOL

Cromseth:
Should I close the pull request?

bnkai:
I think #49 is already enough. #4 will probably be closed in favor of that also.
For the json stuff have a look a the existing scrapers how they do the detection based on script and contains. After that it's moslty regex related.

Cromseth:
Closed for being covered at #49

PR #54, opened 2020-05-24
Adds IsThisReal xPath scraper

halorrr:
As mentioned here https://github.com/stashapp/CommunityScrapers/pull/49#issuecomment-633166300 this PR is currently in a blocked state due to the quirks of these sites and post-processing limitations.

This is the currently done work on it though.

### DIFF
--- a/SCENE-SCRAPABLE.md
+++ b/SCENE-SCRAPABLE.md
@@ -171,6 +171,7 @@ iloveblackshemales.com|GammaEntertainment.yml
 immorallive.com|GammaEntertainment.yml
 innocenthigh.com|PaperStreetMedia.yml
 intimatelesbians.com|Spizoo.yml
+isthisreal.com|IsThisReal.yml
 japanhdv.com|JapanHDV.yml
 jaysinxxx.com|GammaEntertainment.yml
 jessicajaymesxxx.com|Spizoo.yml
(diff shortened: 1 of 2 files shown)


halorrr:
Well 21sextury.com and puretaboo.com changes their site since this PR came up and the performers/tags/details are no longer in the page source. They are still on there for isthisreal.com though. So I updated the scraper to be just for that site now that we have the split function and I could at least finish it for that one.

bnkai:
Everything ok , just the studio missing.

halorrr:
So the Studio name is no where in the page source (properly formatted) and there doesn't seem to be an actually search page, just a dynamic JS search that loads on any page.

I elected for this method which gets the studio name, but if the use has spaces between the words, the studios won't match. If this gets addressed https://github.com/stashapp/stash/issues/626 it will be a non-issue. Another way to address would be once this merges https://github.com/stashapp/stash/pull/616 the map function could be used in a later PR to map all possible studios to the correct spacing. 

This is what I added for now:

```
      Studio:
        Name:
          selector: //link[@rel="canonical"]/@href
          replace:
            - regex: .+(?:\/video\/)([^\/]+).+
              with: $1
```

bnkai:
`https://www.isthisreal.com/en/videos?query="title of scene"` works for me
Needed a little guesswork to find that, the thing is though  NO data in the page source of the results :-(
I think the canonical link is a fair compromise. Shall i merge ?

halorrr:
Yup I think this is good to merge.

PR #56, opened 2020-05-25
Adds LegalPorno xPath scraper

halorrr:


scrapers/LegalPorno.yml
```diff
@@ -0,0 +1,31 @@
...
+    scene:
+      Title:
+        selector: //h1[@class="watchpage-title"]//text()
+        concat: " "
+      Date:
+        selector: //span[@title="Release date"]/a/text()
+        parseDate: 2006-01-02
+      Details: $description/div[3]/dd/text()
+      Performers:
+        Name: $description/div[1]/dd/a[position() mod 2 = 1]/text()
```
bnkai, line 20: performers are not always every two links eg https://www.legalporno.com/watch/34413/veronica_leal_roughed_up_double_penetrated_by_four_men_sz2073
```
Name: $description/div[1]/dd/a[contains(@href,'legalporno.com/model')]/text()
```
seems to work

scrapers/LegalPorno.yml
```diff
@@ -0,0 +1,31 @@
...
+      Studio:
+        Name: //div[@class="studio-director"]//a/text()
```
bnkai, line 22: same studio gets matched twice 
```
Name: //div[@class="col-md-4 col-lg-3 hide-mobile text-right"]/div[@class="studio-director"]//a/text()
```
limits that to once

scrapers/LegalPorno.yml
halorrr: Fixed!

scrapers/LegalPorno.yml
halorrr: I had noticed that, but got lazy since it was always the same studio in my testing, and only the first match would get picked up anyway. Updated though

PR #69, opened 2020-05-31
Add Performer Scraping support to ManyVids

compound-dumbo:


scrapers/ManyVids.yml
```diff
@@ -19,4 +29,69 @@ xPathScrapers:
...
+      Birthdate:
+        selector: $aboutlabel[contains(text(), 'Age')]/following-sibling::*
+        replace:
+          - regex: ' \(.*\)$'
+            with: ''
+        parseDate: 2006.01.02
+      Ethnicity:
+        selector: $aboutlabel[contains(text(), 'Ethnicity')]/following-sibling::*
+      Country:
+        selector: $aboutlabel[contains(text(), 'Nationality')]/following-sibling::*
```
bnkai, line 63: from a quick test, since we are looking for the country
`selector: $aboutlabel[contains(text(), 'Nationality')]/following-sibling::*/img/@alt` 
seems to fetch what we need and in most cases spares us from editing

scrapers/ManyVids.yml
```diff
@@ -19,4 +29,69 @@ xPathScrapers:
...
+          - regex: '.*?or '
+            with: ''
+          - regex: ' cm$'
+            with: ''
+          - regex: \D+[\s\S]+
+            with: ""
+      Measurements:
+        selector: $aboutlabel[contains(text(), 'Measurements')]/following-sibling::* | $aboutlabel[contains(text(), 'Breast Size')]/following-sibling::*
+      FakeTits:
+        selector: $aboutlabel[contains(text(), 'Breast Size')]/following-sibling::*[contains(text(), 'Cohesive Gel') or contains(text(), 'Silicon') or contains(text(), 'Saline')]
```
bnkai, line 78: an `or contains(text(), 'Natural')` seems to be missing

scrapers/ManyVids.yml
```diff
@@ -19,4 +29,69 @@ xPathScrapers:
...
+        replace:
+          - regex: '\s+?[0-9]+.*'
+            with: ''
```
bnkai, line 81: since our field is `FakeTits: ` should we replace `Natural` with `No` and either replace everything else with "Yes" or leave it as it is ? eg `Silicon`

scrapers/ManyVids.yml
```diff
@@ -19,4 +29,69 @@ xPathScrapers:
...
+      Twitter: //*[@class='mv-about__banner-social']/a/i[@class='icon twitter-icon']/../@href
+      Instagram: //*[@class='mv-about__banner-social']/a/i[@class='icon instagram-icon']/../@href
+      Birthdate:
+        selector: $aboutlabel[contains(text(), 'Age')]/following-sibling::*
+        replace:
+          - regex: ' \(.*\)$'
+            with: ''
+        parseDate: 2006.01.02
+      Ethnicity:
+        selector: $aboutlabel[contains(text(), 'Ethnicity')]/following-sibling::*
```
bnkai, line 61: i think we need some `replace` in ethnicity to keep the db consistent 
eg `Black / Ebony` -> `black` , `White / Caucasian` -> `white` , `Latino / Hispanic` -> `hispanic`
NewFreeones ,  Babepedia scrapers have most of the cases i think

scrapers/ManyVids.yml
```diff
@@ -19,4 +29,69 @@ xPathScrapers:
...
+            with: ''
+          - regex: $
+            with: ' - today'
+      Tattoos:
+        selector: $aboutlabel[contains(text(), 'Tattoos')]/following-sibling::*
+      Piercings:
+        selector: $aboutlabel[contains(text(), 'Piercings')]/following-sibling::*
+
+      Image:
+        selector: //*[contains(concat(' ',normalize-space(@class),' '),' mv-model-display ')][not(@data-current-portrait = '')]/@data-current-portrait | //*[contains(concat(' ',normalize-space(@class),' '),' mv-model-display ')][not(@data-current-avatar = '')]/@data-current-avatar
```
bnkai, line 95: is there a case where 
`selector: $pimg[not(@data-current-portrait = '')]/@data-current-portrait|$pimg[not(@data-current-avatar = '')]/@data-current-avatar`
doesn't work ?
where $pimg `$pimg: //div[@class="mv-model-display mv-model-display__block js-profile-display-container"]`
I only tried a few where it seems to work.

compound-dumbo: Not every MV star has a portrait, but everyone has an avatar.
The default avatar is a data url containing an svg logo.
Not sure if that should be filtered out

scrapers/ManyVids.yml
compound-dumbo: Agreed on your examples, but what about the other ethnicities on MV, such as Native American, Middle Eastern, Alaskan and so on?

scrapers/ManyVids.yml
compound-dumbo: Agreed that `Natural` should be picked up and replaced with `No`.
IDK how this field is currently handled, but I would like to retain as much information as possible, so for non-`Natural` values, I'd like to keep it as is (eg `Silicon`)

So for queries for performers `FakeTits` you would list all performers with non-empty, non-`No` `FakeTits`

scrapers/ManyVids.yml
compound-dumbo: Nice, I didn't even catch the alt on the img

scrapers/ManyVids.yml
bnkai: The first scraper used in stash had only four categories asian, white, black and hispanic to correlate with the asian, caucasian, black and latin from freeones. I think freeones reclassifies the rest into those four categories.
It would make sense to adopt some more ethnicities like stashdb has: `indian, middle eastern, mixed , other` to have at least some distinct values we can switch from later if needed.
That leaves out (assuming `East Indian` corresponds to `indian`) `Pacific Islander` `Alaskan`  and `Native American` from the MV site. Idk for them add them as they are and leave it to the user to edit, add to other ?

scrapers/ManyVids.yml
bnkai: We have filters that have `equals`, `not equals`, `is null`, `is not null` i think so keeping `no` for `Natural` and filtering by `equals No` or `not equals No`  would work  and we can keep the extra info you want.

scrapers/ManyVids.yml
bnkai: Imo it's ok either way default or none at all.

halorrr:
Also don't forget to add an entry for this in the performer scraper documentation part of the README.md

scrapers/ManyVids.yml
compound-dumbo: I opted to keep the extra ones (including East Indian), but lower-cased for at least a bit of consistency

PR #72, opened 2020-06-03
ManyVids: Fix Twitter and Instagram links

peolic:
Fix to get only the usernames

### DIFF
--- a/scrapers/ManyVids.yml
+++ b/scrapers/ManyVids.yml
@@ -49,8 +49,16 @@ xPathScrapers:
           - regex: Trans
             with: transgender_female
       URL: //a[@data-tab='about']/@href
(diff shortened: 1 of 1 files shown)


bnkai:
Stash (dev version) works both with url and plain usernames.
If its for cosmetic purposes though then it's ok i think.
Isn't it simpler though to just remove the starting string ?
```
replace:
   - regex: ^.*twitter\.com/
     with: ''
```

peolic:
> Isn't it simpler though to just remove the starting string ?

It might be okay for ManyVids, yes, but I encountered multiple profiles on (New)FreeOnes which had irregular links (`http://twitter.com/username?lang=en` and such)

bnkai:
Looks ok then.
Merging this, for other cases when we are not sure i think it's better to leave the url as is.

PR #78, opened 2020-06-16
Fix image scraping in BangBros scraper

halorrr:
This fixes an issue with the BangBros scraper pulling dead image links.

If you look at the page source on a video like `https://bangbros.com/video24949/the-afro-fuck` the source for the `img` isn't actually the scene image. The proper one is linked in the `video` though.

### DIFF
--- a/scrapers/BangBros.yml
+++ b/scrapers/BangBros.yml
@@ -15,7 +15,7 @@ xPathScrapers:
       Performers:
         Name: //div[@class="vdoCast"]/a[position()>1]/text()
       Image:
-        selector: //img[@id="player-overlay-image"]/@src
+        selector: //video/@poster
         replace:
           - regex: ^
             with: "https:"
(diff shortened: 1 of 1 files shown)


bnkai:
Can you update the last updated date also ?
I think we should make the "# last updated" part a recommended one for all future scrapers

bnkai:
Bangbros uses the highres image from the `img` xpath and fall backs on error to a lower res image that is also used in the `video` xpath. Using the `img` xpath we get a page with `Not Found` instead of the image hence the error `image uknown format`
We can't test for that so i guess we should settle by getting the fallback image everytime like you propose by using the `video` xpath you mention.

halorrr:
Oops forgot to update the date, fixed. Yeah I think I like the last updated part as a standard, I started doing it on the firsts ones I created and then just kept doing it from there on

PR #82, opened 2020-06-19
Adds 21Roles xPath scraper

halorrr:
I feel that the xPaths on this are all correct but when I tried to test it with this URL:
`https://21roles.com/game/DisplayPlayer/gameId/85358`

I get a GraphQL error:
`
GraphQL error: Get https://track.21roles.com/track/OC40NDEuNDEuOTEyMS4wLjAuMC4wLjA?whitesite=https://21roles.com/game/DisplayPlayer/gameId/85358: stopped after 10 redirects
`

Not sure how to get around this.

### DIFF
--- a/scrapers/21Roles.yml
+++ b/scrapers/21Roles.yml
@@ -1,0 +1,57 @@
+name: "21Roles"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - 21roles.com/game/DisplayPlayer/gameId/
+      - lifeselector.com/game/DisplayPlayer/gameId/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //title/text()
+        replace:
+          - regex: .+(?:DisplayPlayer\s-\s)(.+)(?:\s-\sDisplayPlayer)+
+            with: $1
+      Details: //div[@class="info"]/p/text()
+      Tags:
+        Name:
+          selector: //title/text()
+          replace:
+            - regex: .+(?:DisplayPlayer\s-\s)(.+)(?:\s-\sDisplayPlayer)+
+              with: https://21roles.com/site/search/keyword/$1
+            - regex: (.+)(?:\s-\sInteractive).+
+              with: https://lifeselector.com/site/search/keyword/$1
+          subScraper:
+            selector: //div[@class="details"]/div[contains(.,'Labels')]//a/text()
+            concat: ","
+          split: ","
+      Performers:
+        Name:
+          selector: //div[@class="modelBlock"]/div[@class="description"]/h1/a/text()
+          replace:
+            - regex: .+(?:\/)(\d+)+
+              with: https://21roles.com/game/DisplayPlayer/gameId/$1/view/cast
+          subScraper:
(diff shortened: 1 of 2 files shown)


bnkai:
As mentioned in discord it was an issue of cookies/redirect depth.
Even with this you need to recheck the performers and tags selectors, as js is used. Having a look at the page source shows that they are inside divs.

halorrr:
Once your PR merges I'll do some testing on this one to get it finalized.

halorrr:
I updated the performers/tags selectors and it looked like they should work based on the source, but still no luck 🤔

halorrr:
@bnkai when you get a chance could I get your help on the performers/tags on this one? I'm not seeing why they aren't working for me

bnkai:
the subscraper gets a 
```html
<!DOCTYPE html PUBLIC "-//IETF//DTD HTML 2.0//EN"><html><head>
<title>400 Bad Request</title>
</head><body>
<h1>Bad Request</h1>
<p>Your browser sent a request that this server could not understand.<br/>
</p>

</body></html>
```
response from the search site (had to later some code to see the response).
Its has some sort of cookie, referrer protection i think

The perfomers (cast) page seems to load fine and the selector looks legit, i don't know why it only gets the first. I'll investigate some more when i have some more time.I think i may be missing something.

The tags won't be retrieved though without some code changes (if even possible)

halorrr:
@bnkai I just made some slight tweaks to this one to improve it, did you ever get a chance to look at why it only gets the first performer and tags are a no-go?

bnkai:
@halorr yeah couldn't figure out why. I'll have another look though to see if anything changed.

halorrr:
@bnkai Figured this one out. The performer issue was because the subscraper only returns a single object so I have to concat in the subscraper then split it back out.

For the tags they work as soon as I turn on CDP, so there must be something that needs to actually load for them to work.

bnkai:
@halorrr  you somehow messed up the `SCENE-SCRAPABLE.md`  when resolving the conflct

halorrr:
Fixed!

halorrr:
Made a few more tweaks to this and it now supports `lifeselector.com` (which seems to be the exact same site with a different domain and logo on it)

scrapers/21Roles.yml
```diff
@@ -0,0 +1,57 @@
...
+            concat: ","
+          split: ","
+      Performers:
+        Name:
+          selector: //div[@class="signup-right-col"]//input[@id="requestUri"]/@value
+          replace:
+            - regex: .+(?:\/)(\d+)+
+              with: https://21roles.com/game/DisplayPlayer/gameId/$1/view/cast
+          subScraper:
+            selector: //div[@class="content"]//h1/a/text()
```
bnkai, line 36: this selector seems to get some performers twice in scenes with multiple performers (  `https://lifeselector.com/game/DisplayPlayer/gameId/86406` , `https://21roles.com/game/DisplayPlayer/gameId/86503` )
this 
```yaml
            selector: //div[@class="modelBlock"]/div[@class="description"]/h1/a/text()
```
seems to work ok

halorrr: Updated

PR #83, opened 2020-06-19
Adds XXXPawn xPath scraper

halorrr:
Had a bit of an interesting use of the subscraper for this one. One domain has the scene date, and one doesn't, so I used the subscraper to be able to pull the scene date regardless of which domain is used. (One site also has better images, so I doubled down and used it there too)

I'm sure I'm missing something silly here but I do seem to be having a date parsing issue:
`
Warning Error parsing date string 'March 5th, 2017' using format 'January 2nd, 2006': parsing time "March 5th, 2017" as "January 2nd, 2006": cannot parse "th, 2017" as "nd, "
`

I'm not sure what I'm doing wrong with the parse format.

### DIFF
--- a/scrapers/XXXPawn.yml
+++ b/scrapers/XXXPawn.yml
@@ -1,0 +1,40 @@
+name: "XXXPawn"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - xxxpawn.com
+      - xxxpawn.xxx
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h1/text()|//div[@id="content"]//h2/text()
+      Date:
+        selector: //link[@rel="canonical"]/@href
+        replace:
+          - regex: .+(?:\.xxx|\.com\/video\d+)\/([\w-]+)(?:.+|)
+            with: https://xxxpawn.xxx/$1
+        subScraper:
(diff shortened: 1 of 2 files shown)


bnkai:
Again there is no parse format that parses the letter s after the day. I solved that in a couple of scrapers by regex removing the non numeric part of the day eg 5th -> 5 2nd -> 2

halorrr:
That did the trick!

scrapers/XXXPawn.yml
```diff
@@ -0,0 +1,42 @@
+name: "XXXPawn"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - xxxpawn.com
+      - xxxpawn.xxx
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
```
bnkai, line 10: leftover common -> $details should be removed

bnkai:
Just a leftover $details the rest work fine.

bnkai:
hm for the titles  we get an extra one
`//h1/text()|//div[@id="content"]//h2/text()`
looks like it fixes that
remember `|` isn't an `or` its actually `union` so if both xpaths exist it will return results of both

halorrr:
Updated! That is a good clarification, I had been treating it as an OR

PR #84, opened 2020-06-20
Adds iKissGirls xPath scraper

halorrr:


bnkai:
A sample scene `http://ikissgirls.com/trailers/Afternoon-Lovers.html` that has details in it.
Pretty  rare for the site

halorrr:
Added the description selector in

PR #85, opened 2020-06-20
Adds FamilyTherapyXXX xPath scraper

halorrr:


bnkai:
```yaml
 Tags:
        Name:
          selector: //div[@class="entry-content"]/p[contains(text(),'***')]|//p[@class="post-meta"]//a/text()
          replace:
            - regex: \*\*\*|Starring\s
              with:
          split: ' & '
```
isn't the second xpath only needed ?
the first xpath with the replace: starring  can be removed and used for the performer names.
was there a case that it had tags that aren't in the post-meta part?

bnkai:
as stated above
```
Performers:
        Name: //p[@class="post-meta"]//a/text()
```
can be replaced by the Starring replace regex above as in most cases a performer is in the entry content but not in the post-meta part. Where there cases with the opposite ocurring?

scrapers/FamilyTherapyXXX.yml
```diff
@@ -0,0 +1,41 @@
...
+  sceneScraper:
+    common:
+    scene:
+      Title: //h1/text()
+      Date:
+        selector: //meta[@itemprop="uploadDate"]/@content
+        parseDate: 2006-01-02T15:04:05-07:00
+      Details:
+        selector: //div[@class="entry-content"]/p[1]/text()|//meta[@itemprop="description"]/@content
+        concat: ''
```
bnkai, line 17: Concat doesn't do anything here.
To work it needs a proper option eg  whitespace `'  '` 
What you use is `''` which is an empty char
If you do that you'll see that we get 2 times the title because `|` union returns results for both xpaths
On from the 2 xpaths in the union need either to be removed or made more explicit
Was there a sample scene that needed the `|` to work ?

halorrr:
I don't know what I was thinking with my flip flop of the tags and performer logic lol, I've fixed that. Question coming in the thread for the concat though.

scrapers/FamilyTherapyXXX.yml
halorrr: So for concat then, is there no way to concat the strings without a separator? Ex if the two sections are broken up mid word like `fam` and `ily` I don't want a whitespace in there, I just want them straight concatted.

There were some scenes that required the other xpath to work, I'll revisit tonight and see if I can make it more explicit.

bnkai:
i think when concat's value is checked  if its "" , the empty string, it gets ignored and not processed.
The strings.Join that it uses though is capable of joining with an empty string.
Make an issue with an example scraper and i'll look at it.

halorrr:
^ of course I can not find an example for the above anymore, I'll create an issue when it comes up in a scraper again. I've updated this one.

PR #87, opened 2020-06-20
Add studio field to CherryPimps

halorrr:


bnkai:
`//h1` doesn't seem to be consistent 
https://wildoncam.com/trailers/17235-alexiszara-vanetinanappi.html
https://petite.xxx/trailers/17199-aryanaamatista.html
https://cherrypimps.com/trailers/17215-naomiswann.html
sometimes it is `//h2` also

`//div[@class="container"]/a[@class="logo"]/@title` seems to work better but havent tried a lot of scenes

halorrr:
I tested 6 scenes from each site with the above version and it worked. I've updated the PR

PR #89, opened 2020-06-20
Add studio field to GammaEntertainment

halorrr:


bnkai:
Seems to work for most cases. Didn't have time to test all studios though. Which did you check? 
It also seems that girlsway.com changed layout so we should make a separate scraper for that.

halorrr:
I had only tested on a handful. Not sure what happened with Vivid, I restored the old scraper in this PR and I'm gonna audit every site on the list with a test to make sure there are no other ones like that hiding in here.

halorrr:
Ok did an audit of all sites (tested 1 scene on each site). Studio worked on all of them.

Found that prettydirty had changed up their date format, so I added regex to rearrange it nicely.

girlsway has switched to the new adulttime format (21sextury, puretaboo) and will have to wait for 625 to merge.

shemaleidol & iloveblackshemales both redirect to a new site tsfactor so I've removed and replaced them.

bnkai:
Looks ok 
```yaml
          - regex: <\\\/br>|<br\s\\\/>|<br>|<br\\/>
            with: "\n"
```
had to add an extra `<br\\/>` for the lines as the descriptions in prettydirty used that

PR #91, opened 2020-06-22
Add Puffynetwork and Vipissy xPath scraper

niemands:
The two sites are pretty similar, but some differents prevent me from making only one config


scrapers/Puffynetwork.yml
```diff
@@ -0,0 +1,39 @@
...
+    common:
+      $performer: //dl/dd/a
+    scene:
+      Title: 
+        selector: //h2[@class="title"]/span/text()
+        replace:
+          - regex: ^(.*)[—]\s+
+            with:
+      Date:
+        selector: //dl/dt[2]/span/text()
```
bnkai, line 18: imo its better to use something like `selector: //dl/dt[text()="Released on:"]/span/text()` instead of the position `dt[2]`
although it seems to work ok as it is

scrapers/Puffynetwork.yml
```diff
@@ -0,0 +1,39 @@
...
+        parseDate: Jan 2, 2006
+      Details: //div[@class="show_more"]/text()[1]|//div[@class="show_more showed_more"]/text()[1]
```
bnkai, line 20: `Details: //div[@class="show_more"]/text()[1]`
is enough , the second div doesn't exist in the page source
When viewing the scene from firefox if you click on the more button the `show more` div becomes `show_more showed_more` 
Since our scraper takes the page source directly and no js is involved this isn't needed

scrapers/Puffynetwork.yml
```diff
@@ -0,0 +1,39 @@
...
+      Tags: 
+        Name: //p[@class="tags"]/a/text()
+      Performers:
+        Name: $performer/text()
+      Studio:
+        Name: 
+          selector: //h2[@class="title"]/div/div[2]/a/text()
```
bnkai, line 27: `selector: //h2[@class="title"]//div[contains(text(),"Site:")]/a/text()`
same as for the date imo its safer to avoid positions if possible

scrapers/Vipissy.yml
```diff
@@ -0,0 +1,31 @@
...
+      $performer: //dl/dd/a
+    scene:
+      Title: 
+        selector: //section[@class="downloads"]/strong/text()
+        replace:
+          - regex: ^(.*)[—]\s+
+            with:
+      Date:
+        selector: //dl/dd[2]/text()
+        parseDate: Jan 2,2006
```
bnkai, line 19: missed a space after the comma  `parseDate: Jan 2, 2006` -> `parseDate: Jan 2, 2006`
as stated for the other scraper  imo ` selector: //dl/dd[contains(txt(),"Released on:")]/text()`  is safer than `dd[2]`

scrapers/Vipissy.yml
```diff
@@ -0,0 +1,31 @@
...
+      Details: //div[@class="show_more"]/text()[1]|//div[@class="show_more showed_more"]/text()[1]
```
bnkai: also for the reason  explained in the other scraper `Details: //div[@class="show_more"]/text()[1]` is enough

bnkai:
Just a small detail i forgot to mention
We recently started adding a comment like this
`# Last Updated  June 22, 2020`
to the end of the file to track changes among scraper versions
Its not strictly enforced but is highly recommended.
Can you update both yml files with it ?

niemands:
Good idea, done

halorrr:
Small thing, not sure how much it matters but github is flagging that there is no newline at the end of the file. When I create these in Atom it always does that part for me.

niemands:
Hm, Notepad++ didn't warn me about that. I might should consider switching to a proper editor in the future 😉

PR #93, opened 2020-06-23
Add a blank template for url scene scraper

halorrr:
I personally would find this useful to have, figured others might as well.

Currently when I create a scene scraper, I duplicate an existing one, then clear out all the existing paths. This adds a blank template that can quickly be duplicated and ready to start building on.

### DIFF
--- a/templates/SceneScraperTemplate.yml
+++ b/templates/SceneScraperTemplate.yml
@@ -1,0 +1,25 @@
+name: "SiteName"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - siteurl.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+      Details:
+      Date:
+        selector:
+        parseDate: Jan 2, 2006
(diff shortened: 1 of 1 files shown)


bnkai:
I think it could be useful as a starting template.
Can we simplify the name a bit ? eg XPathSceneTemplate or SceneScraperTemplate or TemplateSceneScraper .... The yml extension implies that its for an xpath scraper.
Adding a 0 or  0_ or some symbol as a prefix might be  a good  idea so that it  will show at the start of the file list or at the end if needed. This will also separate it better from the rest of the scrapers.

halorrr:
I updated it to be `SceneScraperTemplate` 

Regarding the `0_` to separate it, you may have missed that I put it in a new folder called `templates` so it already has separation from the actual scrapers.

I chose the new folder option so if someone takes the workflow that I do, which is to download a zip of the repo and just copy the scrapers folder to stash, they aren't copying over the templates as well.

bnkai:
Oh missed the folder. Looks ok then.

PR #96, opened 2020-06-23
Adds Firstanalquest xPath scraper

niemands:
Thanks again to hyde and halorrr for helping me with the subscraper

### DIFF
--- a/scrapers/Firstanalquest.yml
+++ b/scrapers/Firstanalquest.yml
@@ -1,0 +1,35 @@
+name: firstanalquest
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - firstanalquest.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performer: //ul[@class="list-inline"][contains(text(),"Models:")]/li/a
+      $title: //div[@class="badge-content"]/preceding-sibling::span/text()
+    scene:
+      Title: $title
+      Details: //div[@class="text-desc"]/p/text()
+      Tags:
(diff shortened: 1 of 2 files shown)


scrapers/Firstanalquest.yml
```diff
@@ -0,0 +1,36 @@
...
+      - firstanalquest.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performer: //ul[@class="list-inline"][contains(text(),"Models:")]/li/a
+      $title: //div[@class="badge-content"]/preceding-sibling::span/text()
+    scene:
+      Title: $title
+      Details: //div[@class="text-desc"]/p/text()
```
bnkai, line 14: In scenes that the description has links inside 
`http://www.firstanalquest.com/videos/first-time-anal-sex-with-the-inexperienced-teen-lita-black-361/`
`http://www.firstanalquest.com/videos/hardcore-anal-sex-with-horny-inexperienced-teen-girlfriends-346/`
the details won't work  because the paragraph gets split and also the anchor text is missing
Leaving the details selector as `Details: //div[@class="text-desc"]` is safer and works ok for a couple of scenes i tested

niemands: Good catch and also good to know, that you can leave out the /text() selector. Otherwise I would have tried messing with the concat postprocessing

bnkai: By default the innertext is exracted from the resulting fields, It should work ok in most cases without the need for text()

If the specific div has some unrelated info then selectively concating would indeed be the only choice.

PR #97, opened 2020-06-24
Add tags to xxxpawn

halorrr:
Realized after the fact that the .com version has useful tags in the meta.


bnkai:
Good catch!
Update the  `Last updated date`

halorrr:
Updated!

PR #98, opened 2020-06-24
Update PaperStreetMedia

peolic:
- [x] Add FosterTapes & DadCrush
- [x] Update Performers selector to fix empty values

### DIFF
--- a/SCENE-SCRAPABLE.md
+++ b/SCENE-SCRAPABLE.md
@@ -92,6 +92,7 @@ czechstreets.com|Czechav.yml
 czechsupermodels.com|Czechav.yml
 czechtaxi.com|Czechav.yml
 czechwifeswap.com|Czechav.yml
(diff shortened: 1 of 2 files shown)


bnkai:
The 2 extra sites seem to work ok.
Just noticed though that due to some js comments in the anchor text we get two extra empty performers for   every  performer entry.
Switching to 
```yaml
      Performers:
        Name: //p[@class="model-name"]/a
```
leaves the "cleaning" to stash and we get the expected number of performers.
I tried a few scenes from every site and it seems to work ok. 
Can we verify that @peolic @halorrr ?

peolic:
@bnkai 
I see what you mean.
I tried a few sites with that change, seems to work fine.

halorrr:
Looks good in my tests too. Small thing but you updated the date number but not the month in the "last updated" section.

peolic:
@halorrr 
Thanks

Issue #73, opened 2020-06-08
Scrapers Request


iamjen023:
**updated on April 22, 2024**

This issue is to remain open as a ongoing list.
I will ~~crossout~~ once we get it made.

Any of this can be worked on at any time.

Please ask for site with this style.
name of site | url | free/paid/premium(pay for more) | requires subscription to view (yes/no) | notes 

Site | URL | Free/Paid/Premium| Sub Req (Yes/No) | Notes|Worked On By
----|-----------------------|--------------------------------------|---------------------|------------------------------|-------
abdreams |  https://abdreams.com/ | paid | yes | none
abplayhouse | http://www.abplayhouse.com/ | paid | no | none
Allover30 | https://www.allover30.com/ | paid | yes | none
bailey jay |  https://www.ts-baileyjay.com/ | paid | no | none
diapermess |  https://www.diapermess.com/ | paid | no | none
erodougazo | https://erodougazo.com/ | free | no subscription | detailed Performer & film metadata, JAV only
Gelbooru | https://gelbooru.com | free | no | mostly images, but also some scenes
Max Hardcore | https://www.max-hardcore.com/ | Paid | no | Scenes and models
natalie mars | https://nataliemars.com/ | paid | no | none |
Pascals Sub Sluts | https://www.pascalssubsluts.com/submissive/ | paid | yes | Scenes and models
punishedindiapers | http://punishedindiapers.com/ | paid | no | none
redtube |  https://www.redtube.com/ | premium| no | none
Suicide Girls | https://www.suicidegirls.com | paid | yes | probably galleries and scenes
Triga Films | https://trigafilms.com/ | paid | no | Doesn't require a subscription to view details, but does require a free login
youtube | https://youtube.com | premium | no | none
~21roles~ |~https://www.21roles.com/~| ~paid~ | ~no~ | ~none~|halorrr
~adult time studios~|~https://www.adulttime.com/~|~paid~ |~yes~ |~none~|Belleyy
~angela white~ | ~http://angelawhite.com/~ | ~paid~ | ~no~ | ~none~ | halorrr
~avmoo~|~https://avmoo.online/~|~free~|~no subscription~|~Scenes and Performers, contain images for both, change domain name frequently, JAV only~|MortonBridges
~Broke Straight Boys~|~https://brokestraightboys.com/~|~paid~|~no~|~Mostly clips~|Maista6969
[...]

PR #99, opened 2020-06-25
Adds xtube xPath scraper

niemands:
This one was requested in #73.
In some premium scenes the uploader seems to be the performer or you can at least uniquely identify her/him, so I added the uploader for the performer. I have also added the categories to the tags, as stash does not distinguish between them and they are often times more concise than the actual tags.

### DIFF
--- a/scrapers/Xtube.yml
+++ b/scrapers/Xtube.yml
@@ -1,0 +1,30 @@
+name: xtube
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - xtube.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //form/h1/text()
+      Date:
+        selector: //script[contains(text(), "dimension10")]/text()
+        replace:
+          - regex: ^.+'(\d{4})(\d{2})(\d{2})'.+$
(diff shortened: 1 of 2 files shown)


bnkai:
In these kind of sites you get what you can.The only thing  i could comment on is
```yaml
      Performers:
        Name: //input[@name="contentOwnerId"]/@value[not (contains(.,"sponsor_"))]
```
changing the above seems to reduce some false performer entries. Usually sponsored uploads are from studios not actual performers.

scrapers/Xtube.yml
```diff
@@ -0,0 +1,30 @@
...
+        parseDate: 01 02, 2006
+      Studio:
+        Name:
+          selector: //title/text()
+          replace:
+            - regex: .+
+              with: xtube
+      Details: //meta[@property="og:description"]/@content
+      Performers:
+        Name: //form[@id="postCommentForm"]/input[@name="contentOwnerId"][not (contains(.,"sponsor_"))]/@value
```
bnkai, line 26: the `[not (contains(.,"sponsor_"))]` there doesnt do anything
it shouls be in the `@value`
like this
`//form[@id="postCommentForm"]/input[@name="contentOwnerId"]/@value[not (contains(.,"sponsor_"))]`

PR #104, opened 2020-07-01
Create iafd-scene.yml

xiosensei:
tested on several scenes. I do not like the fact that the scraper name is iafd-scene but the iafd.yml name is taken by iafd performer scraper.


bnkai:
It is recommended to combine the scraper (scene+performer). You can edit the existing iafd and add the relevant sections.
A couple of examples you can check are the `manyvids` and the `pornhub` scrapers.

xiosensei:
https://github.com/stashapp/CommunityScrapers/pull/105

PR #105, opened 2020-07-02
Update Iafd.yml

xiosensei:
added scene scraper for iafd

### DIFF
--- a/scrapers/Iafd.yml
+++ b/scrapers/Iafd.yml
@@ -19,11 +19,11 @@ xPathScrapers:
     scene:
       Title: //h1/text()
       Date: 
-        selector: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]/text()
-        parseDate: January 2, 2006
+        selector: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]
+        parseDate: Jan 2, 2006
       Details:  //div[@id="synopsis"]/div[@class="padded-panel"]//text()
       Studio:
-        Name: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Studio']/following-sibling::p[1]/text()
+        Name: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Studio']/following-sibling::p[1]
       Performers:
         Name: //div[@class="castbox"]/p/a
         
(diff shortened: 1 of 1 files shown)


scrapers/Iafd.yml
```diff
@@ -10,6 +15,18 @@ performerByURL:
     scraper: performerScraper
 
 xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h1/text()
+      Date: 
+        selector: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]/text()
+        parseDate: January 2, 2006
```
bnkai, line 23: the date format should be changed to `Jan 02, 2006`

scrapers/Iafd.yml
```diff
@@ -10,6 +15,18 @@ performerByURL:
     scraper: performerScraper
 
 xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h1/text()
+      Date: 
+        selector: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]/text()
```
bnkai, line 22: `//div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]/text()`
the trailing text `/text()` should be removed as it gives an extra empty date match  
`//div[@class="col-xs-12 col-sm-3"]//p[text() = 'Release Date']/following-sibling::p[1]` is enough

scrapers/Iafd.yml
```diff
@@ -10,6 +15,18 @@ performerByURL:
...
+        parseDate: January 2, 2006
+      Details:  //div[@id="synopsis"]/div[@class="padded-panel"]//text()
+      Studio:
+        Name: //div[@class="col-xs-12 col-sm-3"]//p[text() = 'Studio']/following-sibling::p[1]/text()
```
bnkai, line 26: the studio doesn't work
you have to either do
`//div[@class="col-xs-12 col-sm-3"]//p[text() = 'Studio']/following-sibling::p[1]/a/text()`
since the text belonged to the `anchor` and not the `p`
or
`//div[@class="col-xs-12 col-sm-3"]//p[text() = 'Studio']/following-sibling::p[1]`
which would leave it to stash to evaluate the node text of the `p`  and get everything inside he `p`  including the `anchor `
i would recommend the 2nd option

scrapers/Iafd.yml
xiosensei: this one is weird... I will test to make sure that you are right before changing

xiosensei:
Thanks for the feedback @bnkai learned few things doing this.

PR #106, opened 2020-07-02
Adds TheNude performer search and xPath scraper

hyde231:


scrapers/TheNude.yml
```diff
@@ -0,0 +1,76 @@
...
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.thenude.com
+    scraper: performerScraper
+xPathScrapers:
+  performerSearch:
+    performer:
+#      Name: //a[@class="model-name"]/text()
+      Name: //a[@class="model-name"]/../../a/@title
```
bnkai, line 15: another alternative for the names shown when searching
```yaml
      Name:
        selector: //figcaption/span
        replace:
          - regex: "^AKA:"
            with: ""
```
i don't know which is better though

scrapers/TheNude.yml
```diff
@@ -0,0 +1,76 @@
...
+      URL: //a[@class="model-name"]/@href
+  performerScraper:
+    performer:
+      Name: //meta[@itemprop="name"]/@content
+      URL: //meta[@itemprop="url"]/@content
```
bnkai, line 20: `URL: //meta[@itemprop="url"]/@content[1]` 
gets only the first url
as it was we get an extra cover url

scrapers/TheNude.yml
```diff
@@ -0,0 +1,76 @@
...
+        concat: "-"
+      Aliases: 
+        selector: //meta[@itemprop="additionalName"]/@content
+        concat: ", "
+      Tattoos: 
+        selector: //li/span[@class="list-quest"][contains(text(),'Tattoos')]/../text()
+      Piercings: 
+        selector: //li/span[@class="list-quest"][contains(text(),'Piercings')]/../text()
+      Image:
+        selector: //meta[@itemprop="image"]/@content
```
bnkai, line 73: we should get only the first image here 
`selector: (//meta[@itemprop="image"])[1]/@content`

scrapers/TheNude.yml
```diff
@@ -0,0 +1,76 @@
...
+    scraper: performerScraper
+xPathScrapers:
+  performerSearch:
+    performer:
+#      Name: //a[@class="model-name"]/text()
+      Name: //a[@class="model-name"]/../../a/@title
+      URL: //a[@class="model-name"]/@href
+  performerScraper:
+    performer:
+      Name: //meta[@itemprop="name"]/@content
```
bnkai, line 19: `(//meta[@itemprop="name"])[1]/@content`
gets rid of the second matched name (usually studio)

scrapers/TheNude.yml
hyde231: agreed, I had that in mind, but didn't know about the trick with the parenthesis

scrapers/TheNude.yml
hyde231: I can't decide as well between the three possibilities:
1. `//a[@class="model-name"]/text()`  gets only the name
2. `//a[@class="model-name"]/../../a/@title`  gives more details on the label / studio 
3. `//figcaption/span` gives a whole bunch of aliases as well.

So what would be the expected behavior for which workflow?

hyde231:
What still is missing is handling of the paginated results of a search. Currently only the first page is taken into account

scrapers/TheNude.yml
hyde231: If you don't exactly know what you are looking for, option 3 seems to be the best

bnkai: only the name is consistent with the rest of the performer scrapers, but doesn't help much
label and studio can be helpful to get the exact performer you want 
all aliases help when you don't know the studio,label

so all are useful
pick anyone you want and leave the others as a comment, then the user can select based on his usage patterns.

bnkai:
The simple brute force solution would be to add `&filter_per_page=all` to the url and so get only one page for all results
```
https://www.thenude.com/index.php?page=search&action=searchModels&__form_name=navbar-search&m_aka=on&filter_per_page=all&m_name={}
```
I don't know how that will impact performance though
The default is `24` for this site

I would imagine if the user types only a part of the name for example and gets  a lot of results (more than say 24) he would type something more so that he doesn't need to scroll down. Doing that would probably limit the number enough so that only a single page is enough.

As i think about it choosing a per page limit that we think is enough ( probaby a little bigger than the default) seems ok  but i would like to have some feedback on that.

PR #108, opened 2020-07-05
Add Javlibrary xPath scene scraper

Belleyy:
Tested for the EN version, but should work for other (CN,TW,JA) since Div keep same name.

Using :
- ID as Title
- Title in Detail
- Maker as Studio

Wall preview:

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -2,7 +2,7 @@ name: javlibrary
 sceneByURL:
   - action: scrapeXPath
     url:
-      - https://www.javlibrary.com/
+      - www.javlibrary.com/
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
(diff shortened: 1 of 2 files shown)


scrapers/javlibrary.yml
```diff
@@ -0,0 +1,29 @@
+name: javlibrary
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.javlibrary.com/
```
bnkai, line 5: can we relax the url a bit ?
```yaml
    url:
      - www.javlibrary.com/
```
it seems to me that in some cases javlibrary redirects to http as well and i wouldn't want to manually edit the urls to change to https

bnkai:
Can you also add a comment with the last updated date at the end of the file ?
Something like this
```yaml
# Last Updated July 5, 2020
```
is  highly recommended on new scrapers to keep track of changes across versions

PR #115, opened 2020-08-02
Update Brazzers.yml

fustios:
New Site layout. I used the RealityKings.yml as a template, they are quite similar.

### DIFF
--- a/scrapers/Brazzers.yml
+++ b/scrapers/Brazzers.yml
@@ -1,29 +1,21 @@
-name: brazzers
+name: "Brazzers"
 sceneByURL:
   - action: scrapeXPath
     url:
-      - https://www.brazzers.com/scenes
+      - https://www.brazzers.com/video/
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
-    common:
-      $sceneinfo: //div[@class="scene-main-info"]
     scene:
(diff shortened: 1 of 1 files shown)


scrapers/Brazzers.yml
```diff
@@ -1,29 +1,21 @@
...
       Performers:
-        Name: $sceneinfo/p[@class="related-model"]/a
-      Image:
-        selector: //div[@id="trailer-player"]/img/@src
-        replace:
-          - regex: ^
-            with: "https:"
+        Name: //div[@class="sc-1b6bgon-2 jywgHS"]/span/a/text()
+      Studio:
+        Name: //div[@class="vdkjux-5 gIliaY"]/text()
```
bnkai, line 20: This should be 
```yaml
     Studio:
        Name: //a[@class="vdkjux-5 gIliaY"]/text()
```
instead

bnkai:
Can you also add a comment like this at the end ?
```yaml
# Last Updated August 4, 2020
```

scrapers/Brazzers.yml
fustios: Ah yes, i changed it.

fustios:
Added the update info at the end of the file and changed the Studio path.

PR #116, opened 2020-08-02
Update RealityKings.yml

fustios:
Some sites have a movies categorie. Scraper won't work for these movies with the restriction to scene in the url.

### DIFF
--- a/scrapers/RealityKings.yml
+++ b/scrapers/RealityKings.yml
@@ -48,6 +48,7 @@ sceneByURL:
       - https://www.iconmale.com/
       - https://www.propertysex.com/scene/
       - https://www.devianthardcore.com/scene/
+      - https://www.familyhookups.com/scene
     scraper: sceneScraper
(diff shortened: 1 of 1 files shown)


bnkai:
The `scene` part was left there i think to note that the scraper was made/tested with the videos/scenes in mind not the movies part.
For most fields it could work both  ways  but it needs to be verified first. Images for example seem to have different xpath.

The image xpath has changed btw
```yaml
Image: //img[@class="tg5e7m-1 lSdde"]/@src
```
is needed now

fustios:
Yes I tried it with doghouse, but wasn't sure if the image tag changed or if it's just not the same on each site. Seems to have changed everywhere. 
I updated the image xPath for scenes and added the movie image xPath.

schneckenschnabel:
- https://www.familyhookups.com/scene
 works with this too

fustios:
Yes it works, I added it.

PR #118, opened 2020-08-05
Improve LegalPorno. Add AssTraffic.

woodgen:
This pull request adds performer support to the LegalPorno scraper. It implements a fragment scraper for scenes that retrieves the correct scene title and URL if a valid scene ID was found. This can be then scraped with the scene by URL parser.

Furthermore this PR contains a scene and performer scraper for AssTraffic.com.

### DIFF
--- a/scrapers/AssTraffic.yml
+++ b/scrapers/AssTraffic.yml
@@ -1,0 +1,62 @@
+name: "AssTraffic"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - asstraffic.com/movies/
+    scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - asstraffic.com/models/
+    scraper: performerScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h2
+      Date:
+        selector: //div[@class="container"]//span[contains(.,"Added")]/text()
+        replace:
+          - regex: "Added "
+            with:
+        parseDate: January 2, 2006
+      Details: //p[@class="mg-md"]/text()
+      Performers:
+        Name: //div[@id="video-info"]//a[contains(@href,'/models')]/text()
+      Studio:
+        Name:
+          fixed: AssTraffic
+      Image: //video[@id='video']/@poster
+      Tags:
+        Name: //div[contains(@class,"tag-container")]/a/text()
+
+  performerScraper:
+    performer:
+      Name: //h3[@class="mg-md"]
+      Country:
+        selector: //div[@class='col-md-4 modeldetail']/strong[text() = 'Nationality:']/following-sibling::text()[1]
+        postProcess:
+          - map:
+              Hungarian: Hungary
+              French: France
+              Czech: Czech Republic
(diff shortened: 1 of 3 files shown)


scrapers/AssTraffic.yml
```diff
@@ -0,0 +1,60 @@
...
+        replace:
+          - regex: "Added "
+            with:
+        parseDate: January 2, 2006
+      Details: //p[@class="mg-md"]//text()
+      Performers:
+        Name: //div[@id="video-info"]//a[contains(@href,'/models')]/text()
+      Studio:
+        Name:
+          Fixed: AssTraffic
```
bnkai, line 29: should be `fixed` not `Fixed`

scrapers/AssTraffic.yml
```diff
@@ -0,0 +1,60 @@
...
+  - action: scrapeXPath
+    url:
+      - asstraffic.com/models/
+    scraper: performerScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h2
+      Date:
```
bnkai, line 18: Date imo is safer like this
```yaml
      Date:
        selector: //div[@class="container"]//span[contains(.,"Added")]/text()
        postProcess:
          - replace:
            - regex: "Added "
              with:
          - parseDate: January 2, 2006
```

scrapers/AssTraffic.yml
```diff
@@ -0,0 +1,60 @@
...
+        selector: /html[1]/body[1]/div[1]/div[1]/div[1]/div[1]/div[2]/span[1]//text()
+        replace:
+          - regex: "Added "
+            with:
+        parseDate: January 2, 2006
+      Details: //p[@class="mg-md"]//text()
```
bnkai, line 24: the `//` before text() isn't needed `/` is enough
```yaml
      Details: //p[@class="mg-md"]/text()
```

scrapers/AssTraffic.yml
```diff
@@ -0,0 +1,60 @@
...
+      Performers:
+        Name: //div[@id="video-info"]//a[contains(@href,'/models')]/text()
+      Studio:
+        Name:
+          Fixed: AssTraffic
+      Image: //video[@id='video']/@poster
+
```
bnkai, line 31: tags can also be added for the scene scraper
```yaml
      Tags:
        Name: //div[@class="col-sm-8 col-md-8 no-padding-side tag-container"]/a/text()
```

scrapers/AssTraffic.yml
woodgen: Makes total sense. I guess this was one of my first scrapers where I just copied the absolute path of the element from my chromium plugin here.

scrapers/AssTraffic.yml
woodgen: The reason why I didn't do this initially was that AssTraffic uses dictionaries for tags in contrast to our flat list model. This will result in garbage tags sometimes with `big`, `average` etc values for the various body related tags. But there are many tags that make sense on their own, so it's better than nothing.

PR #121, opened 2020-08-07
Add performerByURL and movieByURL scrapers to Gamma Entertainment and Jules Jordan

woodgen:
This patch set adds performer scrapers to Gamma Entertainment and Jules Jordan.
GE was only tested with Evil Angel, but works well. I also just scrapes the photo and name, which is enough for my needs.

Jules Jordan currently gets the wrong height in imperial units. To fix this I am going to write a post process in go that converts that at some point.

### DIFF
--- a/scrapers/JulesJordan.yml
+++ b/scrapers/JulesJordan.yml
@@ -8,6 +8,13 @@ sceneByURL:
       - https://www.theassfactory.com/
       - https://www.spermswallowers.com/
     scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - julesjordan.com/trial/models/
+    scraper: performerScraper
+
 xPathScrapers:
   sceneScraper:
     common:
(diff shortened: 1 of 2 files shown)


woodgen:
This now requires my `toMetric` post process which can be found here:
https://github.com/stashapp/stash/pull/711

bnkai:
Adding the blocked label till 711 is sorted out.

woodgen:
This is not blocked anymore, https://github.com/stashapp/stash/pull/711 was merged.

scrapers/JulesJordan.yml
```diff
@@ -29,10 +43,54 @@ xPathScrapers:
...
+              with: https://www.julesjordan.com/
+
+  performerScraper:
+    performer:
+      Name: //span[@class='title_bar_hilite']/text()
+      Birthdate:
+        selector: //div[@class='model_bio model_bio_info']//text()[contains(.,'Age')]
+        postProcess:
+          - replace:
+            - regex: "Age:"
```
bnkai, line 60: In some performers birthdate is missing and only the age eg `18` is present.
Not really much can be done about it though and constructing a regex to only match birth dates is probably a waste of time.

PR #122, opened 2020-08-07
RealityKings: Update to match site again.

woodgen:
RealityKings.com was apparently updated. I needed these changes to make
it work again.

This patch uses the contains function to only look for relevant
obfuscated class, and be more fault tolerant.

### DIFF
--- a/scrapers/RealityKings.yml
+++ b/scrapers/RealityKings.yml
@@ -50,6 +50,13 @@ sceneByURL:
       - https://www.devianthardcore.com/scene/
       - https://www.familyhookups.com/scene
     scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - realitykings.com/model/
+    scraper: performerScraper
+
 xPathScrapers:
   sceneScraper:
     scene:
@@ -65,5 +72,30 @@ xPathScrapers:
       Studio:
         Name: //a[contains(@class, "kXGSit")]/text()
       Image: //img[contains(@class, "lSdde")]/@src
+
+  performerScraper:
+    performer:
+      Name: //div[contains(@class, "cNKuKg")]/text()
+      Birthdate:
+        selector: //span[contains(.,"Date of Birth")]/following-sibling::span[1]/text()
+        postProcess:
+          - replace:
+            - regex: "nd"
+              with:
+            - regex: "th"
+              with:
+            - regex: "rd"
+              with:
+            - regex: "([\\d]+)st"
+              with: $1
+          - parseDate: January 2, 2006
+      Measurements: //span[contains(.,"Measurements")]/following-sibling::span[1]/text()
+      Country: //span[contains(.,"Birth Place")]/following-sibling::span[1]/text()
+      Height:
+        selector: //span[contains(.,"Height")]/following-sibling::span[1]/text()
+        postProcess:
+          - feetToCm: true
+      Image:
+        selector: //img[contains(@class, "ibyLSN")]/@src[contains(.,"model")]
 
 # Last Updated August 05, 2020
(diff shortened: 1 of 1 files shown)


scrapers/RealityKings.yml
```diff
@@ -50,22 +50,48 @@ sceneByURL:
       - https://www.devianthardcore.com/scene/
       - https://www.familyhookups.com/scene
     scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - realitykings.com/model/
+    scraper: performerScraper
+
```
bnkai, line 59: adding
```yaml
performerByName:
  action: scrapeXPath
  queryURL: https://www.realitykings.com/models?q={}
  scraper: performerSearch
```
allows for search by name

scrapers/RealityKings.yml
```diff
@@ -50,22 +50,48 @@ sceneByURL:
...
 xPathScrapers:
```
bnkai, line 60: needed by the performerByName
```yaml
  performerSearch:
    common:
      $result: //section[@id="73374444684"]//a[@class="aq1tgu-1 hupvSc"]

    performer:
      Name: $result/@title
      URL:
        selector: $result/@href
        replace:
          - regex: ^
            with: https://www.realitykings.com
```

scrapers/RealityKings.yml
```diff
@@ -50,22 +50,48 @@ sceneByURL:
...
+    performer:
+      Name: //div[contains(@class, "cNKuKg")]/text()
+      Birthdate:
+        selector: //span[contains(.,"Date of Birth")]/following-sibling::span[1]/text()
+        postProcess:
+          - replace:
+            - regex: "nd"
+              with:
+            - regex: "th"
+              with:
```
bnkai, line 86:
```yaml
 - regex: "rd"
              with:
            - regex: "([\\d]+)st"
              with: $1
```
needed for 1st and 3rd 
(st matches with August so a more precise regex was needed)

scrapers/RealityKings.yml
```diff
@@ -50,22 +50,48 @@ sceneByURL:
...
+          - parseDate: January 2, 2006
+      Measurements: //span[contains(.,"Measurements")]/following-sibling::span[1]/text()
+      Country: //span[contains(.,"Birth Place")]/following-sibling::span[1]/text()
```
bnkai, line 89: country  is almost never what we want and not consistent 
`https://www.realitykings.com/model/4979/jessa-rhodes`
`https://www.realitykings.com/model/503/veronica-rodriguez`
`https://www.realitykings.com/model/56999/naomi-swann`

scrapers/RealityKings.yml
```diff
@@ -50,22 +50,48 @@ sceneByURL:
...
+      Height:
+        selector: //span[contains(.,"Height")]/following-sibling::span[1]/text()
+        postProcess:
+          - feetToCm: true
+      Image:
+        selector: //img[contains(@class, "ibyLSN")]/@src[contains(.,"model")]
```
bnkai, line 96: In some performers `https://www.realitykings.com/model/65102/brooklyn-gray` we can probably get the `FakeTits` or the `Ethnicity` field `https://www.realitykings.com/model/503/veronica-rodriguez`. Not sure if its worth the trouble though

scrapers/RealityKings.yml
woodgen: Since stash doesn't misbehave when it doesn't recognize the country I would leave this in, since it's better than nothing.

scrapers/RealityKings.yml
woodgen: I will leave it as something "for the future".

PR #130, opened 2020-08-11
Add ThePornDB (metadataapi.net) JSON scraper

bnkai:
withoutpant's JSON scraper for the  ThePornDB
Changed the scene query to search for `oshash` as well as the `filename`


woodgen:
Just tried this with a scene that just has the file name in the title and got this log:
```
DEBU[0026] Processing scene:                            
WARN[0026] Could not find json path 'data.0.url' in json object 
WARN[0026] Could not find json path 'data.0.date' in json object 
WARN[0026] Could not find json path 'data.0.description' in json object 
WARN[0026] Could not find json path 'data.0.background.small' in json object 
WARN[0026] Could not find json path 'data.0.title' in json object 
```

Besides the log it didn't do anything. Am I doing something wrong here? Do I need to switch to `oshash` to search by hash? I'm still on MD5, which was default when I started.

bnkai:
@woodgen that is when no result is found.
It searches using the filename and the oshash. If you don't have an oshash or it isn't present it should return results from the filename.

woodgen:
@bnkai That makes sense. The ThePornDB needs more hashes then :) I also was able to find entries with this.

PR #132, opened 2020-08-12
Add movie scraping to AEBN

whitty-username:


scrapers/AEBN.yml
```diff
@@ -30,4 +35,37 @@ xPathScrapers:
           - regex: ^
             with: "https:"
           - regex: "\\?s=293w"
-            with: 
+            with:
+
+  movieScraper:
+    movie:
+      Name: //h1[@class="dts-section-page-heading-title"]|//div[@class="dts-section-page-heading-title"]/h1
+      Director: //li[@class='section-detail-list-item-director']//span//a
```
bnkai, line 43: In scenes with multiple directors , eg `` we should concat them since there is only one director field in the movie section.
```yaml
      Director:
        selector: //li[@class='section-detail-list-item-director']//span//a
        concat: ", "
```

PR #137, opened 2020-08-17
TransSensual own scraper and removed from RK

ZZenvo:
TransSensual didn't seem to work with the RealityKings scraper anymore.
So made it's own and deleted it from the RK scraper.


ZZenvo:
Updated with your feedback.
Also decided to just add moviescraper part as well, but it is kinda shitty/pointless, because the movie section on their site don't make too much sense.

scrapers/TransSensual.yml
```diff
@@ -0,0 +1,63 @@
...
+  # of some scenes they released with same theme? 
+  # So a movie has date as the earliest released scene, and other scenes have their own.
+  # Description is just name of the 'movie', so gonna go with whole 'desc & cat' as synopsis.
+  # So i'd say movies on TransSensual aren't 'proper'. But here goes:
+  movieScraper:
+    movie:
+      Name: //h2[@class='wxt7nk-4 fSsARZ']/text()
+#      Director: 
+#      Duration: 
+      Aliases:
```
bnkai, line 49: Instead of aliases it gets the performers. Is this intentional?

scrapers/TransSensual.yml
```diff
@@ -0,0 +1,63 @@
...
+      Date:
+        selector: //span/div[contains(.,"Release Date")]/../text()
+        parseDate: January 2, 2006
+      Synopsis: 
+        selector: //div[@class='tjb798-2 flgKJM']//text()  # just to have something in synopsis
+        concat: " "
+      Studio:
+        Name: //meta[@name="dti.network"]/@content
+      FrontImage: //img[contains(@src, "cover")]/@src
+      BackImage: //img[contains(@src, "cover")]/@src
```
bnkai, line 61: Since there is no backimage shouldn't we just leave it out?

scrapers/TransSensual.yml
ZZenvo: Yes, that was intentional.
Some movies got title-aliases?
Just liked having names in that field myself, but we can change it.
I guess with a movie section overhaul, some of these things will change anyways?

scrapers/TransSensual.yml
ZZenvo: Yeah, forgot that movie-tab view only needed/shows frontimage, so we can leave it out.

ZZenvo:
Seems like site has changed(classes names) because it is part of RK network.

ZZenvo:
Updated with your feedback & retested with a few scenes and movies.
Scenescraper:
Supported Site|Scraper|Content
------------- | ------------- | -------------
transsensual.com/scene|TransSensual.yml|Trans

Moviescraper:
Supported Site|Scraper|Content
------------- | ------------- | -------------
transsensual.com/movie|TransSensual.yml|Trans

PR #138, opened 2020-08-18
Add Private xPath scene and movie scraper

niemands:


bnkai:
Duration is not ideal but nothing can be done now
A minutes to duration function would be helpful here.
Merging as is

PR #139, opened 2020-08-19
Add HimerosTV & TimTales xPath scene scraper

whitty-username:
HimerosTV image scrape doesn't appear to be working for some reason. Might need some help on that one.

TimTales appears to work as expected.


scrapers/HimerosTV.yml
```diff
@@ -0,0 +1,33 @@
...
+      Tags:
+        Name: 
+          selector: //div[@class='scene-tags']//ul//a
+      Image:
+        selector: //div[@class='mejs-poster mejs-layer']//img/@src
+        replace:
+          - regex: ^
+            with: "https://himeros.tv"
+          - regex: "-1x.jpg"
+            with: "-3x.jpg"
```
Belleyy, line 33: Forget to use PostProcess.

```    
        postProcess:
          - replace:
              - regex: ^
                with: "https://himeros.tv"
              - regex: "-1x.jpg"
                with: "-3x.jpg"
```

scrapers/TimTales.yml
```diff
@@ -0,0 +1,31 @@
...
+        Name: 
+          selector: //div[@class='text']//p[@class='categories'][contains(text(),'Men:')]/a
+      Tags:
+        Name: 
+          selector: //div[@class='text']//p[@class='categories'][contains(text(),'Categories:')]/a
+      Image:
+        selector: //*[@class='video player no-volume play-button']/@style
+        replace:
+          - regex: .+(https[^')]+).+
+            with: $1
```
Belleyy, line 31: Same as above, PostProcess for the replace

Belleyy:
- You forget to edit the SCENE-SCRAPABLE.md
- You can add at the end of file `#Last Updated August 20, 2020`
- HimerosTV scraper don't seem to work for me (not only the image), can you give a link maybe i take a wrong one ? (`https://himeros.tv/tour/trailers/Dream-Dick.html`)

whitty-username:
@Belleyy Apologies, I'm a total noob so forgive me for those oversights. It looks like HimerosTV was coincidentally updated **literally** within the last 24 hours. I'll try and update the scraper as its broken now on my end too.

whitty-username:
Okay so HimerosTV is fixed now, but still having trouble with the image. Essentially the domain needs to be injected at the start of the string to complete the image URL. Not sure if I'm using the right command for that. 🤷🏻‍♂️

Belleyy:
For the image : `//div[@class='player-thumb']/img[@class="update_thumb thumbs stdimage"]/@src0_1x`

Why? Because your current selector appear after the page is loaded (javascript change thing). 
When creating scraper, try disabling javascript (if you use Ublock you can block it) then you find the XPath.

Other solution would be to use CDP. Scraper would wait the page to load then getting the XPath  (Check CDP support in Wiki for more info)
```
driver:
  useCDP: true
```

whitty-username:
Took me a while but I figured it out 😂

PR #141, opened 2020-08-19
Adding XsList Performer Scraper (JAV)

Belleyy:
### XsList Performer Scraper (JAV)

Writed "(JAV)" in name to warn people it's for JAV Performers.

Not sure about the format of Measurements.

### DIFF
--- a/scrapers/Xslist.yml
+++ b/scrapers/Xslist.yml
@@ -23,7 +23,7 @@ xPathScrapers:
         selector: //div[@id="layout"]/div/h2[1]/text()
         postProcess:
           - replace:
-            - regex: (.+\()(.+)(\/.+)
+            - regex: (.+\()(\D\W+)([/)].+)
               with: $2
 
       URL: //head/meta[@property="og:url"]/@content
--- a/scrapers/Xslist.yml
+++ b/scrapers/Xslist.yml
@@ -30,10 +30,12 @@ xPathScrapers:
         concat: ", "
         postProcess:
           - replace: 
-            - regex: \,\s\w+\s\w+\s\(.+
+            - regex: "(.+)( \\b[a-zA-Z]+\\s\\b[a-zA-Z]+)(.+?)([\\p{Han}\\p{Hiragana}\\p{Katakana}ー]+)(.+)"
+              with: $4, $1
+            - regex: "(\\b.+?|)([\\p{Han}\\p{Hiragana}\\p{Katakana}ー]+)(.+)(Profile)(.+)|(.+)"
+              with: $2$6
+            - regex: ^,|,$
               with: 
-            - regex: (.+\()(\D\W+)([\/)].+)
-              with: $2
       URL: //head/meta[@property="og:url"]/@content
       Birthdate:
         selector: //div[@id="layout"]/div/p[1]/text()[not(contains(.,"n/a")) and contains(.,"Born")]
--- a/scrapers/Xslist.yml
+++ b/scrapers/Xslist.yml
@@ -20,7 +20,7 @@ xPathScrapers:
     performer:
       Name:
         selector: //span[@itemprop="name"]/text()
-        # To have same name as JavLibrary so the scraper will found the performer.
+        #Uncomment below to convert to Surname Name (JavLibrary compatible)
         #postProcess:
         #  - replace: 
         #    - regex: (.+)(\s)(.+)
--- a/scrapers/Xslist.yml
+++ b/scrapers/Xslist.yml
@@ -41,15 +41,15 @@ xPathScrapers:
             - regex: "cm"
               with: ""
       Measurements: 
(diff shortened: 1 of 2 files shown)


scrapers/Xslist.yml
```diff
@@ -0,0 +1,65 @@
...
+  performerSearch:
+    performer:
+      Name: //li[@class="clearfix"]/h3/a/@title
+      URL: 
+        selector: //li[@class="clearfix"]/h3/a/@href
+  
+  performerScraper:
+    performer:
+      Name: //span[@itemprop="name"]/text()
+      Aliases:
```
bnkai, line 22: Aliases seems weird `https://xslist.org/en/model/91.html` , `https://xslist.org/en/model/94586.html`
This looks better
```yaml
      Aliases:
        selector: //span[@itemprop="additionalName"]/text()
        concat: ", "
```

scrapers/Xslist.yml
```diff
@@ -0,0 +1,65 @@
...
+      Aliases: 
+        selector: //div[@id="layout"]/div/h2[1]/text()
+        postProcess:
+          - replace:
+            - regex: (.+\()(\D\W+)([/)].+)
+              with: $2
+
+      URL: //head/meta[@property="og:url"]/@content
+      Birthdate:
+        selector: //div[@id="layout"]/div/p[1]/text()[not(contains(.,"n/a")) and position()=1]
```
bnkai, line 31: position is not needed if we do it like this
```yaml
      Birthdate:
        selector: //div[@id="layout"]/div/p[1]/text()[contains(.,"Born") and not(contains(.,"n\a"))]
```

scrapers/Xslist.yml
```diff
@@ -0,0 +1,65 @@
...
+                - regex: (Born:\s)(.+)
+                  with: $2
+            - parseDate: January 2, 2006
+      Height: 
+        selector: //span[@itemprop="height"]/text()[not(contains(.,"n/a"))]
+        postProcess:
+          - replace:
+            - regex: "cm"
+              with: ""
+      Measurements:
```
bnkai, line 43: also position not needed here 
```yaml
      Measurements: 
        selector: //div[@id="layout"]/div/p[1]/text()[(contains(.,"Measurements")) and not(contains(.,"n/a"))]
```

scrapers/Xslist.yml
```diff
@@ -0,0 +1,65 @@
...
+              with: ""
+      Measurements: 
+        selector: //div[@id="layout"]/div/p[1]/text()[not(contains(.,"n/a")) and position()=3]
+        postProcess:
+          - replace:
+            - regex: (Measurements:\s)(.+)
+              with: $2
+            - regex: \s+\/\s+
+              with: "-"
+      CareerLength:
```
bnkai, line 51: same as above
```yaml
 selector: //div[@id="layout"]/div/p[1]/text()[contains(.,"AV Activity") and not(contains(.,"n/a"))]
```

scrapers/Xslist.yml
Belleyy: Sometimes performer don't have this line, so your selector don't work. `https://xslist.org/en/model/5.html`
Manage to put your selector + mine with a new regex:

```      Aliases: 
        selector: //span[@itemprop="additionalName"]/text()|//div[@id="layout"]/div/h2[1]/text()
        concat: ", "
        postProcess:
          - replace: 
            - regex: \,\s\w+\s\w+\s\(.+
              with: 
            - regex: (.+\()(\D\W+)([\/)].+)
              with: $2```

Comment (bnkai) [reply scrapers/Xslist.yml:28]: By weird I meant that it's not actually the alias but the name written in Kanji.
If you want to keep that too as an alias then imho your selector would need to come before mine in the union operator so that we end up with : name in kanji , alias1 , alias2 , alias3 ....

Comment (Belleyy) [reply scrapers/Xslist.yml:28]: Her name in Kanji can be useful so I like having it.
I made a new regex to get  `name in kanji , alias1 , alias2 , alias3 ...`.

File: scrapers/Xslist.yml [hunk 5]
```diff
@@ -0,0 +1,75 @@
...
+    performer:
+      Name: //li[@class="clearfix"]/h3/a/@title
+      URL: 
+        selector: //li[@class="clearfix"]/h3/a/@href
+  
+  performerScraper:
+    performer:
+      Name:
+        selector: //span[@itemprop="name"]/text()
+        # To have same name as JavLibrary so the scraper will found the performer.
```
bnkai, line 23: you mean same name convention as JavLibrary ? `Surname Name`   as opposed to `Name Surname` that is now?
I would rephrase it as 
`#Uncomment  below to convert to Surname Name (JavLibrary compatible)`

Belleyy: Thanks for correct me, English is not my native language so sometimes my explanation is bad 👍

PR #150, opened 2020-09-01
Trans500 normal and espanol scraper of their site

ZZenvo:
Found that if you go to the espanol version. e.g. `/tourespanol/` instead of `/tour/` or `/tour3/` in the URL, there is more metadata showing at the scene-page. Here's both versions, since studio is then missing on the espanol part...
Supported Site|Scraper|Content
------------- | ------------- | -------------
trans500.com/tour/|Trans500.yml|Trans
trans500.com/tour3/|Trans500.yml|Trans
trans500.com/tourespanol|Trans500espanol|Trans

Perhaps using Subscraper, could combine them into one, and support the "standard url path" with all fields from both.

### DIFF
--- a/scrapers/Trans500.yml
+++ b/scrapers/Trans500.yml
@@ -1,4 +1,4 @@
-name: trans500
+name: Trans500
 sceneByURL:
   - action: scrapeXPath
     url: 
--- a/scrapers/Trans500.yml
+++ b/scrapers/Trans500.yml
@@ -1,0 +1,20 @@
+name: trans500
+sceneByURL:
+  - action: scrapeXPath
+    url: 
+      - trans500.com/tour/
+      - trans500.com/tour3/
+    scraper: sceneScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="col-sm-12"]/p/../h2/text()
+      Details: //p[@class="description"]
+      Image:
+        selector: //video[@id="my-video"]/@poster
+        replace:
+          - regex: ^
+            with: "http://www.trans500.com"
+      Studio: 
+        Name: //p[@class="pull-right"]/b/text()
--- a/scrapers/Trans500espanol.yml
+++ b/scrapers/Trans500espanol.yml
@@ -36,8 +36,21 @@ xPathScrapers:
         # replace: 
           # - regex: http 
             # with: https
-      Studio: 
+      Studio:
         Name: 
-          fixed: Trans500
+          selector: //div[@class="scene-infobrick"][contains(text(), "Categories:")]/a/text()
+          concat: ","
+          postProcess:
+            - replace:
+                - regex: .*(superramon).*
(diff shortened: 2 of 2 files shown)


bnkai:
@ZZenvo i'll have a look. The only thing prohibiting the use of `subscraper` is that we can't get the `base` url from anywhere in the scene page.

ZZenvo:
> @ZZenvo i'll have a look. The only thing prohibiting the use of `subscraper` is that we can't get the `base` url from anywhere in the scene page.

Oh, yeah! Ofc... ;/

bnkai:
After some digging it turns out that we can get the studio from the tourespanol urls
The site/studio is present in the categories part
A specific site has a respective category added to the scene 
```yaml
      Studio:
        Name: 
          selector: //div[@class="scene-infobrick"][contains(text(), "Categories:")]/a/text()
          concat: ","
          postProcess:
            - replace:
                - regex: .*(superramon).*
                  with: "Super Ramon"
                - regex: .*(tsgirlfriendexperience).*
                  with: "TS Girlfriend Experience"
                - regex: .*(ikillitts).*
                  with: "I Kill It TS"
                - regex: .*(transatplay).*
                  with: "Trans at Play"
                - regex: .*(behindtrans500).*
                  with: "Behind Trans500" 
```
The only catch is if a scene doesn't belong to these sites/studios (from a quick look didn't spot any) or a new site/studio is added
@ZZenvo if this works with your samples we could keep only the tourespanol scraper as it holds all the data

ZZenvo:
@bnkai Yeah, i remember we talked about studio/subsites being in the categories/tags part of the site with the espanol urls.
Nice that you made the regex mapping them out. Will test your changes with with some more samples, and update when possible.

ZZenvo:
@bnkai I finally got around to testing some more, and found that some scenes were missing the specific studioname in the categories part, which is used in the regex. So i just added more of the variations/abbreviations i found, using `|`, to the regex. 
Sometimes a scene had both `ikillitts `and `superramon` in its tags, but then the first regex replace will be used, and they seemed to be in the proper order as how you would want it. So thats good.

Tested on a good amount of scenes.
Only thing is that it feels a bit slower to scrape, but don't know if it's just the site or the regex. Or both. Other than that it seems to work pretty fine.

bnkai:
Seems ok.
Do you want to merge both scrapers ? If yes then just rename one of them
```yaml
name: trans500
```
The `name:` field is the same for both.

The scraping speed seems the same to me, the only thing slowing things down is the image download part afaik.

ZZenvo:
Yeah okay, it is probably just the image download / site.

Kept both for now. Since when on the site it's hard to stay in Espanol mode when navigating around. So you kinda have to put the espanol part of the url, in the url everytime yourself.
You could just get all the 'normal links', and then script/sql to rename the normal urls to espanol ones, but probably not many going to do that.

PR #152, opened 2020-09-02
Fix RK sites broken by PR 122

halorrr:
This PR adds back the classes removed in https://github.com/stashapp/CommunityScrapers/pull/122

The class names were not changed on all sites so the PR broke sites like PropertySex that are still on the old classes


bnkai:
@halorr i merged another scraper file for the old layout to keep things cleaner 
anything that doesnt work shoukd be moved there

PR #163, opened 2020-09-08
Adding Lustery xpath scraper

Tweeticoats:
Adding Lustery xpath scraper.

### DIFF
--- a/SCENE-SCRAPABLE.md
+++ b/SCENE-SCRAPABLE.md
@@ -260,6 +260,7 @@ littleasians.com|PaperStreetMedia.yml|
 littlehellcat.com|Teencoreclub.yml|
 lookathernow.com|RealityKings.yml|
 lubed.com|AMAMultimedia.yml|
+lustery.com|Lustery.yml|Amature
 m45e.com|javlibrary.yml|JAV
 maketeengape.com|Teencoreclub.yml|
 maledigital.com|GammaEntertainment.yml|Gay
--- a/SCENE-SCRAPABLE.md
+++ b/SCENE-SCRAPABLE.md
@@ -260,7 +260,7 @@ littleasians.com|PaperStreetMedia.yml|
 littlehellcat.com|Teencoreclub.yml|
 lookathernow.com|RealityKings.yml|
 lubed.com|AMAMultimedia.yml|
-lustery.com|Lustery.yml|Amature
+lustery.com|Lustery.yml|Amateur
 m45e.com|javlibrary.yml|JAV
 maketeengape.com|Teencoreclub.yml|
 maledigital.com|GammaEntertainment.yml|Gay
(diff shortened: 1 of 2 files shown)


scrapers/Lustery.yml
```diff
@@ -0,0 +1,19 @@
+name: Lustery
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://lustery.com
```
bnkai, line 5: should we limit the url to
```yaml
      - https://lustery.com/video-preview/
```
?

if we use a link like `https://lustery.com/videos#video-442` it wont work it needs `https://lustery.com/video-preview/442`

scrapers/Lustery.yml
```diff
@@ -0,0 +1,19 @@
...
+    scene:
+      Title:
+        selector: //div[@class='video-popup-info']//div[@class='title']/h3/text()
+      Performers:
+        Name: 
+          selector: //div[@class='couple-and-categories']//div[@class='couple']/a/text()
+      Details: //div[@class='video-popup-info']/div[@class='description']/text()
+      Tags:
+        Name:  //div[@class='tags']/a/text()
+
```
bnkai, line 18:
```yaml
      Image: //meta[@property="og:image"]/@content
```
seems to work for the image part

scrapers/Lustery.yml
Tweeticoats: Yes that is probably a good idea.

halorrr: Worth noting the canonical URL exists on the pages via

`<link rel="canonical" href="https://lustery.com/video-preview/442"/>`

Instead of filtering we also could grab canonical URL on the first scrape, and then once the URL is updated the user would be able to scrape properly. Might avoid user confusion figuring out why their link isn't working.

bnkai: If you use the `https://lustery.com/videos#video-442` link the canonical you get is `<link rel="canonical" href="https://lustery.com/videos"/>` and that's the page that stash downloads and why the scraper doesn't work with these links. So no it wouldn't work.

scrapers/Lustery.yml
bnkai, line 18:
```yaml
      Studio:
        Name:
          fixed: Lustery
```

SCENE-SCRAPABLE.md
```diff
@@ -260,6 +260,7 @@ littleasians.com|PaperStreetMedia.yml|
 littlehellcat.com|Teencoreclub.yml|
 lookathernow.com|RealityKings.yml|
 lubed.com|AMAMultimedia.yml|
+lustery.com|Lustery.yml|Amature
```
bnkai, line 263: typo
```
lustery.com|Lustery.yml|Amateur
```

PR #164, opened 2020-09-08
Adding VNAGirls series of sites containing websites for individual models

Tweeticoats:
The same scraper seems to work for everything in my testing.


scrapers/VNAGirls.yml
```diff
@@ -0,0 +1,68 @@
...
+  sceneScraper:
+    scene:
+      Performers:
+        Name:
+          selector: //div[@class='customcontent']/h3/text()
+          split: ","
+      Title:
+        selector: //div[@class='customcontent']/h1/text()
+      Details:
+        selector: //div[@class='customcontent']/div[1]/text()
```
bnkai, line 54: haven't tested on all sites but on some the `div` is missing and instead `h2` is used
the following seems to work
```yaml
     Details:
        selector: //div[@class='customcontent']/div[1]/text()|//div[@class='customcontent']/h2/text()
```

scrapers/VNAGirls.yml
```diff
@@ -0,0 +1,68 @@
...
+      Tags:
+        Name:
+          selector: //div[@class='customcontent']/h4/text()
+          split: ", "
+#     Image:
```
bnkai, line 59: studio and image ( not fully tested on all sites )
```yaml
      Studio:
        Name: 
          selector: //p[@class="copyright"]
          postProcess:
            - replace:
                - regex: .*\s(\w+)\.(?i)com\s.+
                  with: $1 
      Image:
        selector: //p[@class="copyright"]|//center/img/@src
        concat: "|"
        postProcess:
          - replace:
              - regex: .*\s(\w+\.(?i)com)\s.+\|(.+)
                with: https://www.$1/$2
```

PR #127, opened 2020-08-09
Adding several Scrapers 

Belleyy:
Adding multiple scrapers

_Some scrapers was made by people in discord (DDFNetwork, pornworld, scoreland, cumlouder) but didn't make PR_

New Site|Scraper
------------- | -------------
18eighteen.com|TheScoreGroup.yml
1by-day.com|DDFNetwork.yml
40somethingmag.com|TheScoreGroup.yml
50plusmilfs.com|TheScoreGroup.yml
60plusmilfs.com|TheScoreGroup.yml
abuseme.com|AbuseMe.yml
aliciasgoddesses.com|Andomark.yml
ariellynn.com|Andomark.yml
bffs.com|PaperStreetMedia.yml
cherrypimps.com|trafficpimps.yml
cumlouder.com|cumlouder.yml
ddfbusty.com|DDFNetwork.yml
ddfnetwork.com|DDFNetwork.yml
eurogirlsongirls.com|DDFNetwork.yml
euroteenerotica.com|DDFNetwork.yml
family.xxx|trafficpimps.yml
filf.com|filf.yml
handsonhardcore.com|DDFNetwork.yml
hotlegsandfeet.com|DDFNetwork.yml
houseoftaboo.com|DDFNetwork.yml
legsex.com|TheScoreGroup.yml
lethalhardcore.com|lethalhardcore.yml
nympho.com|nympho.yml
onlyblowjob.com|DDFNetwork.yml
pimp.xxx|trafficpimps.yml
plumperpass.com|PlumperPass.yml
pornmegaload.com|TheScoreGroup.yml
pornworld.com|pornworld.yml
r18.com|r18.yml
ravenswallowzxxx.com|Andomark.yml
rawattack.com|RawAttack.yml
reidmylips.elxcomplete.com|Andomark.yml
scoreland.com|TheScoreGroup.yml
sexuallybroken.com|sexuallybroken.yml
sheseducedme.com|Andomark.yml
submissived.com|PaperStreetMedia.yml
sugarcookie.xxx|sugarcookie.yml
sweetfemdom.com|SweetFemdom.yml
toughlovex.com|toughlovex.yml
wankz.com|wankz.yml
xlgirls.com|TheScoreGroup.yml

_Edit 16/08/2020: Updated list_

PR #165, opened 2020-09-09
Add BFFs xPath Scene Scraper

halorrr:


bnkai:
@halorrr there is a pending PR ( https://github.com/stashapp/CommunityScrapers/pull/127 ) that adds the bffs site to the papermedia scraper. I haven't tested that yet but i think its close enough to yours and should work. If that doesn't work now we can merge yours.

halorrr:
Didn't see that one there! We can close this PR then.

PR #166, opened 2020-09-09
Add LittleCapriceDreams Scene Scraper #166

budislov:


scrapers/LittleCapriceDreams.yml
```diff
@@ -0,0 +1,24 @@
...
+      Details: //meta[@property="og:description"]/@content
+      Image: //meta[@property="og:image"]/@content
+      Studio:
+        Name: //meta[@property="og:site_name"]/@content
+      Date:
+        selector: //div[@class="vid_date"]
+        parseDate: January 2, 2006
+      Performers:
+        Name: //div[@class="vid_info_content"][1]//a
+driver:
```
bnkai, line 20: why is useCDP needed ?
it works ok without it.

scrapers/LittleCapriceDreams.yml
```diff
@@ -0,0 +1,24 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.littlecaprice-dreams.com/project
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //meta[@property="og:title"]/@content
+      Details: //meta[@property="og:description"]/@content
```
bnkai, line 11: imo details is more accurate (close to what the site shows) like this
```yaml
      Details: //div[@class="vid_desc"]/p
```
the og:description is more like a synopsis

scrapers/LittleCapriceDreams.yml
budislov: You are correct, it works without it. In my initial testing when I was attempting to scrap several fields and it seemed to not work without the useCDP. I should have tested the final without the useCDP. That was an oversight on my part. Let's pull that out.

scrapers/LittleCapriceDreams.yml
budislov: I was debating between og:description and the "vid_desc" class. I'm fine with making the change to the details to what you propose.

PR #168, opened 2020-09-11
Update Brazzers.yml Xpath Scraper

Belleyy:
Trying a other way to get the Xpath so if the site update **classname** if will still work (i hope).
I take the Svg icon of the "❤️", then go back (parent) to get the Title, Performer, Date.

Btw: It look like this scraper can work with RealityKings too

### DIFF
--- a/scrapers/Brazzers.yml
+++ b/scrapers/Brazzers.yml
@@ -7,26 +7,43 @@ sceneByURL:
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
+    common:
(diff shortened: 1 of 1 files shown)


bnkai:
Unfortunately it doesn't seem that `thumbnailUrl` is present in all scenes
eg `https://www.brazzers.com/video/3788301/brazzers-house-2` `https://www.brazzers.com/video/3788311/the-exxxception` `https://www.brazzers.com/video/3792731/brazzers-house-3-series-trailer`

PR #169, opened 2020-09-11
Update RealityKingsOL.yml xpath scraper

budislov:
Updated the @class on Title, Name, and Image to reflect changes on the sites.

### DIFF
--- a/scrapers/RealityKingsOL.yml
+++ b/scrapers/RealityKingsOL.yml
@@ -17,7 +17,7 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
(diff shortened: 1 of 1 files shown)


bnkai:
I tried avoiding the classes alltogether.
These seem to work for all included sites , need to verify against more samples though

```yaml
      Title: //h1/text()
```
```yaml
      Performers:
        Name: //h1/following-sibling::div//a[contains(@href,"/model")]/text()
```
```yaml
      Image: //img[@alt=""][contains(@src,"poster")]/@src
```

budislov:
I will test out your proposed changes with the samples that I have.

bnkai:
thanks @budislov i checked with some scenes but some more verification is needed

budislov:
@bnkai I've tested this with a bunch of different samples on these various sites with no issues.

PR #186, opened 2020-09-23
Update Kink.yml

Belleyy:
Fixing details

### DIFF
--- a/scrapers/Kink.yml
+++ b/scrapers/Kink.yml
@@ -11,7 +11,9 @@ xPathScrapers:
       Date:
         selector: //span[@class="shoot-date"]/text()
         parseDate: January 2, 2006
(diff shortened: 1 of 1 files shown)


scrapers/Kink.yml
```diff
@@ -11,7 +11,7 @@ xPathScrapers:
       Date:
         selector: //span[@class="shoot-date"]/text()
         parseDate: January 2, 2006
-      Details: //p[@class="description-text"]/text()
+      Details: //span[@class="description-text"]/p
```
bnkai, line 14: From a test url `https://www.kink.com/shoot/20662` it seems that there scenes with multiple paragraphs so a concat is needed
```yaml
      Details:
        selector: //span[@class="description-text"]/p
        concat: "\n\n"

```

PR #187, opened 2020-09-23
Adding MeanaWolf xPath Scraper

Belleyy:


scrapers/MeanaWolf.yml
```diff
@@ -0,0 +1,31 @@
...
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h3/text()
+      Date:
+        selector: //span[text()="ADDED:"]/following-sibling::text()
+        postProcess:
+          - parseDate: January 2, 2006
+      Details: 
+        selector: //div[@class="trailerContent"]/p/text()
```
bnkai, line 16: a lot of scenes (`https://www.meanawolf.com/scenes/Slutty-Sisters-Slumber-Party_vids.html` , `https://www.meanawolf.com/scenes/Go-Go-Stop_vids.html`, `https://www.meanawolf.com/scenes/Its-Been-Awhile_vids.html` )  have <br> in the details that breaks the text() in many parts

```yaml
      Details: 
        selector: //div[@class="trailerContent"]/p/text()
        concat: "\n\n"
```

Belleyy: I saw it but for me the details was only the first part.
The second part 'Clip Contains:' is not really the description...
Anyways i will add it

bnkai: In some scenes the `clip contains` looks better than the first part imo. A user can edit the text afterwards and remove what he doesnt like from the details anyway since like you said in some other scenes the extra text doesn't make sense.

PR #188, opened 2020-09-24
Update SceneScraperTemplate.yml

Belleyy:
Using PostProcess and adding the cheatsheet (Great ressource).

Maybe we should show with regex / map too ?
Like:
```
      Studio:
        Name:
          selector: 
          postProcess:
            - replace:
                - regex: [Regular Expression]
                  with:
            - map:
                studio: Studio
```

### DIFF
--- a/templates/SceneScraperTemplate.yml
+++ b/templates/SceneScraperTemplate.yml
@@ -11,7 +11,8 @@ xPathScrapers:
       Details:
       Date:
         selector:
-        parseDate: Jan 2, 2006
+        postProcess:
+          - parseDate: January 2, 2006
       Image:
       Studio:
         Name:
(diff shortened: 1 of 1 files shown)


bnkai:
The cheatsheet is already present in the wiki or the help section <https://github.com/stashapp/stash/blob/develop/ui/v2.5/src/docs/en/Scraping.md>. Imho the template is just a quick skeleton to start a scraper, it doesn't need to have all the cases (although map should probably be added)  
The cheatsheet or the regex checker https://regex101.com/?flavor=golang are useful but imho shouldn't be in the template. A separate file in the repo containing useful links or just a bookmark in the user's browser is better suited for them imho again.
Any more feedback is welcome on the matter

PR #189, opened 2020-09-24
Update Iafd.yml (Fixing country)

Belleyy:
### Using this list from Wikipedia: https://en.wikipedia.org/wiki/List_of_adjectival_and_demonymic_forms_for_countries_and_nations
Adding a giant map to have the country.
Maybe some country don't have Flag on Stash, i don't tested every country...

Note: 
Still have problem with some performer since they can have multiple Nationality (Ariana Marie, Ariella Ferrera,..)
I was thinking the `Country` field is for Birthplace or Nationality ?
Birthplace will be easier since you can only born to 1 place.

Btw, there is a issues about it in Stash repo : https://github.com/stashapp/stash/issues/810


bnkai:
I think Country is for Nationality ( not  convenient for our case ).
Multiple nationalities are a problem since we don't handle multiple countries well ( we can probably ignore them for the time being  given the small nuber of performers with dual+ nationalities)

Even if there are some flags missing i think your map is the way to go.

Belleyy:
Right, you can merge it if you don't have issue with it 😃

PR #194, opened 2020-09-30
Adds GirlsRimming xPath scraper

halorrr:
Things of note:

- The performer names are listed below each video in related videos and on the performer pages like `https://www.girlsrimming.com/tour/models/alessa-savage.html`, however I could not find a want with a subscraper to get there

- Anything that is part of a series has a date listed on here `https://www.girlsrimming.com/tour/series/series.html` but again, couldn't find a way to get what I need with a subscraper

- Most video has their tags properly listed in the keywords metafield. There are a few poorly entered scenes though like `https://www.girlsrimming.com/tour/trailers/Switch-Me-On-Ep2-Rim-the-Realtor.html` where for some reason the tags field contains a copy of the scene description instead.

### DIFF
--- a/scrapers/GirlsRimming.yml
+++ b/scrapers/GirlsRimming.yml
@@ -1,0 +1,32 @@
+name: "GirlsRimming"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - girlsrimming.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class='updatesBlock']/h2[@class='title']/text()
+      Details: //meta[@name='description']/@content
+      Image: //div[@class='player-thumb']/img/@src0_1x
+      Studio:
+        Name:
+          fixed: GirlsRimming
+      Movies:
+        Name:
+          selector: //div[@class='updatesBlock']/h2[@class='title']/text()
+          postProcess:
+            - replace:
+              - regex: (.+)(?:\sEp\d).*
+                with: $1
+      Tags:
+        Name:
+          selector: //meta[@name='keywords']/@content
+          postProcess:
+            - replace:
+                - regex: "[^,]*Id\\s(\\d+)[^,]*"
+                  with:
+          split: ","
+
+# Last Updated September 30, 2020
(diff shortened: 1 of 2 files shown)


scrapers/GirlsRimming.yml
```diff
@@ -0,0 +1,28 @@
...
+      Image: //div[@class='player-thumb']/img/@src0_1x
+      Studio:
+        Name:
+          fixed: GirlsRimming
+      Movies:
+        Name:
+          selector: //div[@class='updatesBlock']/h2[@class='title']/text()
+          postProcess:
+            - replace:
+              - regex: (.+)(?:\sEp\d:\s).+
```
bnkai, line 21: In some series like `https://www.girlsrimming.com/tour/dvds/yoga-session.html` or `https://www.girlsrimming.com/tour/dvds/fidelity.html` there is no episode title (and thus no `: \s`part after the ep ) in the movie name so the regex doesnt match
This
```yaml
                - regex: (.+)(?:\sEp\d).*
```
should be enough

scrapers/GirlsRimming.yml
```diff
@@ -0,0 +1,28 @@
...
+                with: $1
+      Tags:
+        Name:
+          selector: //meta[@name='keywords']/@content
```
bnkai, line 25:
```yaml
          postProcess:
            - replace:
                - regex: "[^,]*Id\\s(\\d+)[^,]*"
                  with:
          split: ","
```
filters out from the tags some performer names that have ids (eg in `https://www.girlsrimming.com/tour/trailers/Fidelity-Ep1.html`)

bnkai:
For the dates i couldn't find a way also ( i hate sites without a proper search function )

For the performers in some scenes (`https://www.girlsrimming.com/tour/trailers/Fidelity-Ep1.html`) in the keywords you can find the id  ( eg `[^,]*Id\\s(\\d+)[^,]*`) and then access the performer page from there (something like `"https://www.girlsrimming.com/tour/models/popular/?id=$1"` ) using a subscraper to get the full name but
- not all scenes have the performer id (many have only names and some not even performers)
- there is no way to split the meta keyword content ( for the tags ) and then postprocess it. Split is the last action to be executed and there is no easy way to change that.

Given the above i think we probably can't do anything more for this site.

halorrr:
Resolved the review items, and yeah I agree, don't think there is anything more we can do for this one

PR #196, opened 2020-10-03
Remove protocol, www and language subdomains from URL matching

halorrr:
I've experienced this myself with some of the scrapers, so after the chat in #the-scraping-initiative today I decided to open up a PR to clean up the URLs used to match scrapers. Some sites do not employ best practices when it comes to their design so this is a much safer approach for the scrapers.

- Removing Protocols like http and https as some sites don't employ HSTS so both versions are accessible
- Removing www as some sites have not used primary URLs are can be accessed through both versions
- Remove language subdomains like `.en` as a future proofing member for users who might approach from a different language


bnkai:
While the protocol and www is probably safe to remove in all scrapers the language subdomain isn't
Various sites use different layout or localization that can break the scraper.
Eg heyzo and carrineancom scrapers won't work as intended if you don't use the specified subdomain.
(Heyzo uses different localization , while carribeancom an entirely different page layout)

halorrr:
Updated the PR to no longer drop the language subdomains and updated the `Updated At` lines on the ones I edited.

PR #198, opened 2020-10-03
Add SpankBang xPath scraper

halorrr:
Knocking off one of the scrapers from https://github.com/stashapp/CommunityScrapers/issues/73

### DIFF
--- a/scrapers/SpankBang.yml
+++ b/scrapers/SpankBang.yml
@@ -1,0 +1,21 @@
+name: "SpankBang"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - spankbang.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class='left']/h1/@title
+      Details: //meta[@name='description']/@content
+      Image: //div[@class='play_cover']/img/@src
+      Studio:
+        Name: //li[@class='us']/a[@class='ul'][contains(@href,'channel')]/text()
+      Tags:
+        Name: //div[@class='ent']/a[@href[contains(.,'tag')]]/text()|//li[@class='p']/text()
+      Performers:
+        Name: //div[@class='ent']/a[@href[contains(.,'pornstar')]]/text()|//div[@class='left']/h1/a/text()
+      URL: //link[rel='canonical']/@href
+
+# Last Updated October 6, 2020
--- a/SCENE-SCRAPABLE.md
+++ b/SCENE-SCRAPABLE.md
@@ -475,6 +475,7 @@ sneakysex.com|MindGeek.yml|
 soapymassage.com|GammaEntertainment.yml|
 sophiedeelive.com|VNAGirls.yml|
 sororitysluts.com|Hustler.yml|
+spankbang.com|SpankBang.yml|Tube Site
 SpearTeenPussy.com|Teencoreclub.yml|
 spermantino.com|Teencoreclub.yml|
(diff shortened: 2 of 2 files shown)


scrapers/SpankBang.yml
```diff
@@ -0,0 +1,31 @@
...
+            - regex: ^
+              with: https://spankbang.com
+          subScraper:
+            selector: //h1
+            postProcess:
+              - replace:
+                - regex: (.+)\sOfficial.+
+                  with: $1
+      Tags:
+        Name: //div[@class='ent']/a[@href[contains(.,'category') or contains(.,'tag')]]/text()|//li[@class='p']/text()
```
bnkai, line 26: From the samples i tested it seems that tags always contain the keywords from category and we get a lot of dupes as a result. Was there a case where the `category` was needed ?

scrapers/SpankBang.yml
```diff
@@ -0,0 +1,31 @@
...
+      Performers:
+        Name: //div[@class='ent']/a[@href[contains(.,'pornstar')]]/text()|//div[@class='left']/h1/a/text()
```
bnkai, line 28: The second part of the union seems to add a lot of noise and usually the scenes that don't have the first part do not contain `pro` performers. Should we remove the second part and lose some performers from some scenes but make it more clean ? (`https://spankbang.com/2nma9/video/uncle+i+need+your+help` In some cases like this the performer listed from left class , gina gerson here, isnt even present in the scene).

scrapers/SpankBang.yml
```diff
@@ -0,0 +1,31 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class='left']/h1/@title
+      Details: //meta[@name='description']/@content
+      Image: //div[@class='play_cover']/img/@src
+      Studio:
+        Name:
+          selector: //li[@class='us']/a[@class='ul']/@href[contains(.,'channel')]
```
bnkai, line 15: do we need the subscraper ?
it seems to me that 
```yaml
      Studio:
        Name: //li[@class='us']//a[@class='ul'][contains(@href,'channel')]
```
is enough

scrapers/SpankBang.yml
halorrr: I didn't notice the repetition on that one, looking at the examples I was referencing that seems to stay true. I'll remove the category

scrapers/SpankBang.yml
halorrr: It is really a mixed bag with pornstars when I looked around. There are a ton of scenes with `pro` performers that are only listed in the 2nd area like this one for example `https://spankbang.com/47tsa/video/squirting+down+under` however it also contains performers who aren't in the scene. There doesn't seem to be a good grasp on data accuracy across the site, and at the end of the day, the scraper is as accurate as the site it is scraping from.

Here are two examples where the performers are accurate and only in the 2nd area:
`https://spankbang.com/3glmz/video/how+i+met+my+girlfriend+elsa`
`https://spankbang.com/4f5px/video/pov+cute+blonde+in+2+scenes+swallow+and+creampie`

But to really demonstrate the great (sarcasm) data accuracy, here is one where the pro performers in both sections are entirely wrong:
`https://spankbang.com/32kx6/video/best+friend+bet`

Only so much we can do about that 🤷‍♂️

scrapers/SpankBang.yml
halorrr: You are correct! Fixing it

PR #202, opened 2020-10-06
Fixing IAFD

Belleyy:
They use Https now.

Still have minor issue... (maybe this exist before ?)
Sometime it won't find `Lexi Belle` , but `Lexi Bell` work. Sometimes everything work 🤷 

#123


bnkai:
With the fix everything seems to work ok. Even your example @Belleyy works fine for both searches. Maybe a query timed out ?
From <https://www.iafd.com/tips.asp> it seems that not much be done anyway.

PR #210, opened 2020-10-09
Adds 21sextury and girlsway to puretaboo xPath scraper

halorrr:
Girlsway could have been added without any changes, but 21sextury needed different selectors to get the title and description. Luckily the new selector works on all sites though.


Belleyy:
You should remove `girlsway.yml`

halorrr:
I didn't notice that yml there since it didn't have a reference in the scene scrapable list. I've removed it now though!

PR #214, opened 2020-10-10
Update Mindgeek (TransSensual & Movie)

Belleyy:
- Adding Movie scraper for Digital playground + TransSensual 
- Fixing TransSensual by adding into Mindgeek scraper (#123)

Using `ancestor` instead of `../../..`


halorrr:
I was actually just about to PR MindGeek to add `kinkyspa.com`, could you add it in in this PR too?

bnkai:
Looks ok.
You can also add 
```yaml
      - kinkyspa.com/model/
      - transsensual.com/model/
```
in the  `performerByURL:` so that the performer by url scraper works.
The transensual performers page though has only the name and image for most performers.

PR #218, opened 2020-10-12
Adds BoxTruckSex xPath scene scraper

halorrr:


scrapers/BoxTruckSex.yml
```diff
@@ -0,0 +1,29 @@
...
+    scene:
+      Title: //h2/text()
+      Details:
+        selector: //div[@id[contains(.,'viddesc')]]/p/text()
+        concat: "\n"
+      Date:
+        selector: $details/p[@class='mb0']/text()
+        postProcess:
+          - parseDate: Jan 2, 2006
+      Image: //meta[@property='og:image']/@content
```
bnkai, line 20: The image from the meta property is not the same as the one from the scenes. ( the image shown in many samples i tried while relevant doesn't seem to  appear in the scene page)
It seems to me that the image provided by the player is more accurate.
```yaml
      Image:
        selector: //video[@id="jwbox_video"]/@poster
        postProcess:
          - replace:
              - regex: ^
                with: https://www.boxtrucksex.com
```

halorrr: Updated!

PR #220, opened 2020-10-12
Update Javlibrary

Belleyy:
Look like the mirror u44r.com became b47w.com.

- Added sceneByFragment

Use the title to try to find the scene on Javlibrary (you can use mirror too, just replace javlibrary.com). 
**The title must have no extension.** (Like SSNI-000). 
So just remove the extension in Scene edit then save then scrape...

- Added Movie scraper

Sometimes scenes are split in different file since JAV is sometimes more than 4 hours. So the movie scraper can be used to create a link between these files.

-------

Since the `sceneByFragment` need some instruction to use it, i am not sure if it's a good idea to made it...

Note: Cloudflare block the Javlibrary.com atm, try with b47w

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -9,7 +9,6 @@ sceneByURL:
     url:
       - javlibrary.com/
       - m45e.com/
-      - u44r.com/
       - g46e.com/
       - b47w.com/
     scraper: sceneScraper
@@ -18,7 +17,6 @@ movieByURL:
     url:
       - javlibrary.com/
       - m45e.com/
-      - u44r.com/
       - g46e.com/
       - b47w.com/
     scraper: movieScraper
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -1,4 +1,9 @@
-name: javlibrary
+name: JavLibrary
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.javlibrary.com/en/vl_searchbyid.php?keyword={title}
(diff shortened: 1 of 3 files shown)


scrapers/javlibrary.yml
```diff
@@ -1,33 +1,77 @@
-name: javlibrary
+name: JavLibrary
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.javlibrary.com/en/vl_searchbyid.php?keyword={title}
+  #Note: Use the title of your scene. So if your filename is SSNI-000.mp4, remove the extension, save, then use the scraper
```
bnkai, line 5: cleaned up the note part a bit , the rest work fine
```yaml
  # Note: The scraper uses the title from your scene to search for the movie id.
  # Make sure that the title is saved  in the correct form e.g. SSNI-000 ( no extensions or other extra words/characters  should be present) before you use the scrape with function.
  # If javlibrary doesn't work you can replace javlibrary with b47w in the queryURL and sceneScraper scene URL part
```

Belleyy: > If javlibrary doesn't work you can replace javlibrary with **any mirrors** in the queryURL ~~and sceneScraper scene URL part~~

I think it's better not to name a mirror since they can change their name at any moment.
For the URL part, i think it's better to always have javlibrary on it because: 
- Javlibrary link will always work, maybe not the scraper but if you got the URL, you already succeed scraping it.
- If you used a mirror, you wanted to use javlibrary first, so your first thought was to have a javlibrary URL.

PR #221, opened 2020-10-12
add Karissa-Diamond.com xPath scene scraper

SpedNSFW:
Adds a scene scraper for Karissa-Diamond.com. Uses fixed studio and performer as this is the performer's solo site.

### DIFF
--- a/scrapers/Karissa-Diamond.yml
+++ b/scrapers/Karissa-Diamond.yml
@@ -22,11 +22,12 @@ xPathScrapers:
       Performers:
         Name:
           fixed: "Karissa Diamond"
-      Image:
+      Image:      
         selector: //meta[@property="og:image"]/@content
-        postProcess:
-          - replace:
-            - regex: cover
-              with: cover_720
+# Uncomment if you prefer a 720p cover over 1080p
+#        postProcess:
+#          - replace:
+#            - regex: cover
+#              with: cover_720
 
 # Last Updated October 12, 2020
--- a/scrapers/Karissa-Diamond.yml
+++ b/scrapers/Karissa-Diamond.yml
@@ -18,7 +18,7 @@ xPathScrapers:
           - parseDate: January 02, 2006
       Details: //div[@class="well"]/text()
       Tags:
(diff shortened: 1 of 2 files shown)


halorrr:
I'm partial to always including the URL when available to clean up URLs that might have UTM parameters or such trailing at the end, depending how the user got to the scene page. This will grab that:

```
URL: //meta[@property='og:url']/@content
```

halorrr (commented):
_Ignore this comment, GH won't let me delete it. It is a duplicate of the code comment._

scrapers/Karissa-Diamond.yml
```diff
@@ -0,0 +1,32 @@
...
+          - parseDate: January 02, 2006
+      Details: //div[@class="well"]/text()
+      Tags:
+        Name: //div[@class="col-md-5 col-md-offset-1 col-md-push-6 hidden-xs"]//div[@class="col-xs-12 tagList"]/a/text()
+      Performers:
+        Name:
+          fixed: "Karissa Diamond"
+      Image:
+        selector: //meta[@property="og:image"]/@content
+        postProcess:
```
halorrr, line 27: I have to wonder if this is better to have commented out as an option for the user to turn on but not on by default.

scrapers/Karissa-Diamond.yml
```diff
@@ -0,0 +1,32 @@
...
+      Studio:
+        Name:
+          fixed: "Karissa Diamond"
+      Date:
+        selector: //div[@id="setDate"]/text()
+        postProcess:
+          - parseDate: January 02, 2006
+      Details: //div[@class="well"]/text()
+      Tags:
+        Name: //div[@class="col-md-5 col-md-offset-1 col-md-push-6 hidden-xs"]//div[@class="col-xs-12 tagList"]/a/text()
```
halorrr, line 21: This would be shorter and more future proofed for this this class names change a little:

```
//div[@class[contains(.,'tagList')]]/a/text()
```

SpedNSFW: Ahh thanks. Still relatively new to XPaths.

PR #226, opened 2020-10-17
Update javlibrary.yml (Note)

Belleyy:
I don't changed the note on my previous PR... 😅

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -5,7 +5,7 @@ sceneByFragment:
   # Note: The scraper uses the title from your scene to search for the movie id.
   # Make sure that the title is saved in the correct form e.g. SSNI-000 ( no extensions or other extra words/characters should be present) before you use the scrape with function.
   # If javlibrary doesn't work, you can replace javlibrary with any mirrors in the queryURL.
-  # If the movie have a Blu-Ray version, the scraper will set the URL of the DVD version. So you need to use the sceneScraper by URL after.
+  # The scraper matches the DVD version of a title. If you want the Blu-Ray version you'll need to add the correct URL manually and scrape using that afterwards.
   scraper: sceneScraper
 sceneByURL:
   - action: scrapeXPath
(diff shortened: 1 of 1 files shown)


scrapers/javlibrary.yml
```diff
@@ -21,7 +24,7 @@ movieByURL:
       - b47w.com/
     scraper: movieScraper
 xPathScrapers:
-  sceneScraper:
+  sceneScraper by URL:
```
bnkai, line 27: this should be left as it was
```yaml
  sceneScraper:
```
why was it changed ?

Belleyy: euh idk, maybe i mess up with something in github 😢

scrapers/javlibrary.yml
```diff
@@ -2,7 +2,10 @@ name: JavLibrary
 sceneByFragment:
   action: scrapeXPath
   queryURL: https://www.javlibrary.com/en/vl_searchbyid.php?keyword={title}
-  #Note: Use the title of your scene. So if your filename is SSNI-000.mp4, remove the extension, save, then use the scraper
+  # Note: The scraper uses the title from your scene to search for the movie id.
+  # Make sure that the title is saved in the correct form e.g. SSNI-000 ( no extensions or other extra words/characters should be present) before you use the scrape with function.
+  # If javlibrary doesn't work, you can replace javlibrary with any mirrors in the queryURL.
+  # If the movie have a Blu-Ray version, the scraper will set the URL of the DVD version. So you need to use the sceneScraper by URL after.
```
bnkai, line 8: typo
If the movie **has** a Blu-Ray version,  the scraper will set the URL of the DVD version.

If i understand correctly what you mean though  this could be rephrased to


`The scraper matches the DVD version of a title. If you want the Blu-Ray version you'll need to add the correct URL manually and scrape using that afterwards.`

From a quick look it seems to me that most bluray versions have an extra tag `Blu-ray` added and the `Risky Mosaic` one removed ( uncensored version ? ).

Belleyy: ### Exemple: 
If your title is SSNI-888, the scraper will get this page:
`https://www.javlibrary.com/en/vl_searchbyid.php?keyword=SSNI-888`
There is 2 results, one is the Blu-Ray and DVD.
So by default the scraper will take the URL of the movie without "Blu-Ray" in title. (Then you will use the icon near URL to scrape it)

PR #228, opened 2020-10-18
Add FamilyStrokes scraper

aussiemale02:


Belleyy:
FamilyStrokes scraper already exist in `PaperStreetMedia.yml`
Still working for me, so i am not sure about this PR.

```
# Switched to CDP due to Cloudflare errors
driver:
  useCDP: false
```
You need to set it to true if you want use CDP. 🤷‍♂️

PR #232, opened 2020-10-21
Update FreeonesCommunity.yml

Belleyy:
- Adding PostProcess
- Changed xPath (Using name instead of attribute)
- Fixing "Bio" in name


bnkai:
Upon testing i can't seem to access freeones.xxx anymore. I get redirected to the .com site (maybe geolocation or something)
The .com site seems to have now  the same layout with the .xxx as replacing all `freeones.xxx` references  with `freeones.com` seems to work.

PR #241, opened 2020-10-22
queryURLReplace for JavLibrary

Belleyy:
https://regex101.com/r/0JCwtY/2

Not exactly same regex that Bnkai posted in comment of the PR

Note: You will need the latest develop version (October 22)

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -6,9 +6,7 @@ sceneByFragment:
     filename:
       - regex: (.*[^a-zA-Z0-9])*([a-zA-Z-]+\d+)(.+)
         with: $2
-  # Note: The scraper uses the title from your scene to search for the movie id.
-  # Make sure that the title is saved in the correct form e.g. SSNI-000 ( no extensions or other extra words/characters should be present) before you use the scrape with function.
-  # If javlibrary doesn't work, you can replace javlibrary with any mirrors in the queryURL.
+  # Note: If javlibrary doesn't work, you can replace javlibrary with any mirrors in the queryURL.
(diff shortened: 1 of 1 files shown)


bnkai:
Most of my files had the `alpanum-code`form so i thought that the `-` was mandatory.
After looking at the scearch tips though it seems it wasnt necessary to have the `-`
```
You can find the video ID on the jacket of your DVD box. It usually looks like [alphabet]-[number] (e.g. TT-013); in some cases, it may also look like [alphabet]-[alphabet][number] (e.g. ADV-VSR0407). You can type the complete ID or the starting portion of the ID to search.

Search “tt-013”, “tt 013”, “tt-13”, or “tt13” will return video that has the ID “TT-013”.

Search “adv” will return videos whose ID starts with ADV, this is also useful if you are looking for a particular type of videos.
```
The tricky part ( it is easier when there is a `-` in the filename) is to keep only the keyword from the filename

Belleyy:
> 
> 
> Most of my files had the `alpanum-code`form so i thought that the `-` was mandatory.
> After looking at the scearch tips though it seems it wasnt necessary to have the `-`
> 
> ```
> You can find the video ID on the jacket of your DVD box. It usually looks like [alphabet]-[number] (e.g. TT-013); in some cases, it may also look like [alphabet]-[alphabet][number] (e.g. ADV-VSR0407). You can type the complete ID or the starting portion of the ID to search.
> 
> Search “tt-013”, “tt 013”, “tt-13”, or “tt13” will return video that has the ID “TT-013”.
> 
> Search “adv” will return videos whose ID starts with ADV, this is also useful if you are looking for a particular type of videos.
> ```
> 
> The tricky part ( it is easier when there is a `-` in the filename) is to keep only the keyword from the filename

Exactly why i changed your regex because some/most of my file don't have the "-", this regex don't care about it and search Letter with number after it.

PR #242, opened 2020-10-24
Hentai site Galleries Scraper (xPath)

Belleyy:
nhentai & hentai2read scraper

For hentai2read, the release date is only the Year, so i added 01-01 so we got a correct date.

Not sure about details (Alternative title ?) and studio (Artist or group ?)

Btw: there is nothing about galleries scraping in Stash wiki/help
Btw 2: In scraper page in Settings, there in galleries scraper table

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,9 @@ xslist.org|Xslist.yml
 ## Galleries Scrapers
 This list is meant to keep track of which sites are already supported by existing community scrapers. And which scrapers support them. When introducing a new scraper, add the sites your scraper supports to this list in your PR. Please keep the site list in alphabetical order to keep the list tidy.
 
-Supported Site|Scraper|Content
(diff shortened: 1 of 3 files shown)


bnkai:
The nhentai.net site is a mess. They have everything by chapters not by title? And their titles are #$@#$ .
@Belleyy  I thinks its okay with the details and studio you chose, if you don't have anything else to change i think its ok to merge as is

Belleyy:
> 
> 
> The nhentai.net site is a mess. They have everything by chapters not by title? And their titles are #$@#$ .

I never used nhentai.net so i don't really know about this 😄 , but someone in discord asked for it so it was a good reason for me to try to make a gallery scraper.

> @Belleyy I think its okay with the details and studio you chose, if you don't have anything else to change i think its ok to merge as is

Go for it 👍

PR #247, opened 2020-10-25
add sceneByFragment for Kink

bnkai:
Add sceneByFragment support that gets the scene id from filename.
Currently detects ids from filenames with this form (lets say 12345 is the scene id)
```
12345_performer_other_data.mp4
studio - 2009-11-30 - performer - other - data (12345).mp4
````

### DIFF
--- a/scrapers/Kink.yml
+++ b/scrapers/Kink.yml
@@ -4,10 +4,24 @@ sceneByURL:
     url:
       - kink.com
     scraper: sceneScraper
+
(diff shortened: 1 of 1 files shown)


aaron-typo:
Its working. But behind any scraped title I get this sign: 

bnkai:
@aaron-typo oops i knew the `text()` part was there for a reason....
Fixed that , the extra character is the heart after the title

PR #252, opened 2020-11-01
adjust vixen cover selector

SpedNSFW:
The second class is no longer present on images. The  selector appears to be their primary selector for images.

### DIFF
--- a/scrapers/vixenNetwork.yml
+++ b/scrapers/vixenNetwork.yml
@@ -22,7 +22,9 @@ xPathScrapers:
         selector: //span[@data-test-component="ReleaseDateFormatted"]
         parseDate: "January 2, 2006"
       Image:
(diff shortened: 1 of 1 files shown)


bnkai:
@SpedNSFW  can you provide some samples (urls) were the class is not present? I've gone through most of my stuff and the image part works ok. The `irBaDM` is not present in the page source but gets added by js.
Since the scraper uses cdp make sure to add the debug part when testing
```yaml
debug:
  printHTML: true
```
set the log level to debug and check the log not the page source for the selectors.

Even with that the proposed change
```yaml
        selector: //img[@class[contains(., "sc-1egln9q-1")]]/@src
```
which can also  be written as 
```yaml
        selector: //img[contains(@class, "sc-1egln9q-1")]/@src
```

sometimes returns multiple results that would need to be checked to see if  we always want the 1st.

Issue #114, opened 2020-07-21
! Stash xPath scraper code refactoring !


bnkai:
As of commit  **2b92157** ( PR https://github.com/stashapp/stash/pull/616 ) in the dev branch of stash the scraper code was refactored adding new functionality and enhancing the existing one.
Although the xpath scraper code is backwards compatible, scrapers added from now on should use the new `postProcess` field as stated in the PR. Extra functionality like  the `map` post-processing action and the `fixed` field along with some examples are also explained in the PR.

bnkai:
Closing since a validator check is in place now.

PR #254, opened 2020-11-03
Update all scrapers to use `postProcess`

peolic:
as per #114
feel free to reject if this is too big of a PR

I am working on a [scraper schema (+validator)](https://github.com/stashapp/CommunityScrapers/compare/master...peolic:scraper-schema) so spotting the deprecated attributes was rather easy

the changes were made manually and I kept the order of execution as defined in the docs:
- `concat`
- `replace`
- `subScraper`
- `parseDate`
- `split`

### DIFF
--- a/scrapers/21Roles.yml
+++ b/scrapers/21Roles.yml
@@ -12,18 +12,18 @@ xPathScrapers:
         selector: //title/text()
         postProcess:
           - replace:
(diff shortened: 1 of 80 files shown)


bnkai:
@peolic i will have a quick look through all of them, can you update the `# Last Updated Date` part or add it if its missing?

peolic:
@bnkai done

PR #255, opened 2020-11-03
Add infernalrestraints xPath scene scrapper

HeavyDinosaur:
Add infernalrestraints scene scrapper

### DIFF
--- a/scrapers/infernalrestraints.yml
+++ b/scrapers/infernalrestraints.yml
@@ -1,0 +1,40 @@
+name: infernalrestraints
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - infernalrestraints.com/device
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //tbody/tr[1]/td[1]/span[1]/a[1]/text()
+      Date:
+        selector: //tbody/tr[1]/td[1]/span[2]/text()
+        postProcess:
+          - parseDate: "January 02, 2006"
+          - parseDate: "January 2, 2006"
+      Details:
+        selector: //*[@id="articleText"]/tbody/tr/td/text()
+        concat: "<br>"
+        postProcess:
+          - replace:
+            - regex: <br>
+              with: "\n"
+      Performers:
+        Name: //*[@id="articleWrapper"]/tbody/tr[1]/td/span[1]/a[2]/text()
+      Tags:
+        Name:
+          selector: //td[contains(text(),"Tags:")]/text()
+          # To remove the Tags: text from the tags
+          postProcess:
+            - replace:
+              - regex: "Tags:"
+                with:
+          split: ","
+      Studio:
+        Name:
+            fixed: Infernal Restraints
+      Image: //*[@id="my-video"]/@poster
+      URL: //meta[@name='og:url']/@content
+
+# Last Updated Novemeber 2, 2020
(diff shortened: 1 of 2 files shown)


scrapers/infernalrestraints.yml
```diff
@@ -0,0 +1,40 @@
...
+      - infernalrestraints.com/device
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //a[@class="articleTitleText"][1]/text()
+      Date:
+        selector: //span[@class="articlePostDateText"][1]/text()
+        postProcess:
+          - parseDate: "January 02, 2006"
```
bnkai, line 14: this `          - parseDate: "January 02, 2006"` isnt needed
`January 02, 2006` is for sites that use 0 padding for single digit days eg April 01, 2020
`January 2, 2006` parses sites with April 1, 2020 instead
Both parse April 11, 2020

scrapers/infernalrestraints.yml
```diff
@@ -0,0 +1,40 @@
...
+          - parseDate: "January 2, 2006"
+      Details:
+        selector: //td[@class="articleCopyText"]/text()
```
bnkai, line 17: the <br> 's in the details are already parsed so you shouldn't need to do all that
```yaml
      Details:
        selector: //td[@class="articleCopyText"]/text()
        concat: "\n\n"
```
should be enough

scrapers/infernalrestraints.yml
```diff
@@ -0,0 +1,40 @@
...
+          postProcess:
+            - replace:
+              - regex: "Tags:"
+                with:
+          split: ","
+      Studio:
+        Name:
+            fixed: Infernal Restraints
+      Image: //*[@id="my-video"]/@poster
+      URL: //meta[@name='og:url']/@content
```
bnkai, line 38: is the og:url meta present in any scenes ? couldn't find a sample with it.

scrapers/infernalrestraints.yml
```diff
@@ -0,0 +1,40 @@
...
+      Performers:
+        Name: //a[@class="articleTitleText"][2]/text()
+      Tags:
+        Name:
+          selector: //td[contains(text(),"Tags:")]/text()
+          # To remove the Tags: text from the tags
+          postProcess:
+            - replace:
+              - regex: "Tags:"
+                with:
```
bnkai, line 32: The tags seem to have some extra spaces included  after the comma. It doesn't show up in the UI but is shows in the console log if you set it to debug. 
Adding an extra regex after the tags replace fixes that
```yaml
            - replace:
              - regex: "Tags:"
                with:
              - regex: ",\\s*([^,]+)"
                with: ",$1"
```

scrapers/infernalrestraints.yml
HeavyDinosaur: forgot to remove, inherited from another scrapper.

scrapers/infernalrestraints.yml
HeavyDinosaur: fixed

scrapers/infernalrestraints.yml
HeavyDinosaur: fixed

scrapers/infernalrestraints.yml
HeavyDinosaur: updated

PR #256, opened 2020-11-03
Add movieporn xPath scene scrapper

HeavyDinosaur:
Add movieporn scene scrapper


HeavyDinosaur:
so i just realized that 4 other related studios can use the same scrapper, but I cant find a parent studio name

- unrealporn.com
- horrorporn.com
- creativeporn.com
- perversefamily.com

If anyone can help with a naming structure i can add these sites also.

peolic:
Parent studio (webmasters) seems to be CzechCash `http://nats.czechcash.com/external.php?page=sites`

By the looks of it, [`Czechav.yml`](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/Czechav.yml) already contains these sites.
Haven't tested, but maybe you can simply add these 5 sites to the existing scraper.

And in case you don't/can't combine them - you can get the studio name using `//h1/img/@alt` (same selector `Czechav.yml` uses).

HeavyDinosaur:
thanks, ill have a look at it

peolic:
one more tip, make sure you disable JavaScript (or use "view source", or use `debug/printHTML`) when building out an XPath selector.
for example, you don't need CDP to get the scene cover (on `perversefamily.com` at least),
as `//video/@poster` works fine on the "raw" page

HeavyDinosaur:
thanks for the tip. I integrated the 5 sites and verified that they work.

PR #259, opened 2020-11-05
update image and tags for hucows scraper

HeavyDinosaur:
Changes:
Images: Grab the banner image instead of the screenshot 
Tags: Grab relevant tags instead of all the tags

### DIFF
--- a/scrapers/Hucows.yml
+++ b/scrapers/Hucows.yml
@@ -19,7 +19,7 @@ xPathScrapers:
               - regex: ^\s+
                 with:
           - parseDate: 2 Jan 2006
-      Image: //div[@class="imghoverclass postfeat post-single-img"]//img//@src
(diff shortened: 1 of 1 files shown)


scrapers/Hucows.yml
```diff
@@ -19,12 +19,12 @@ xPathScrapers:
               - regex: ^\s+
                 with:
           - parseDate: 2 Jan 2006
-      Image: //center/a[1]/img[@class="lightboxhover"]/@src
+      Image: //div[@class="imghoverclass postfeat post-single-img"]//img//@src
```
bnkai, line 22: shouldnt it be `//img/@src` ?
we could also future proof it a bit by using
`//div[contains(@class,"post-single-img")]//img/@src` assuming `post-single-img` will last longer than the rest of the class name

HeavyDinosaur: fixed

PR #265, opened 2020-11-08
Create .gitattributes

peolic:
Make `*.yml` files count towards the repository's language statistics.

image


peolic:
Could also add the following to 'force' an EOL character to all text files.
```gitattributes
* text=auto eol=lf
```

bnkai:
Can `text=auto`  and `eol=lf` be used together ?
I see mostly `text=auto` or `text eol=lf` 

Also is it safe to use `text=auto` for all files (eg python scipts ) ? Shouldnt we limit that to `*.yml text=auto` ?

peolic:
> Can `text=auto` and `eol=lf` be used together ?
> I see mostly `text=auto` or `text eol=lf`

Hmm, interesting.
It worked for me but that could still be incorrect.
It could be overriding the `text=auto` and applying LF to every file, even non-text.

> Also is it safe to use `text=auto` for all files (eg python scipts ) ? Shouldnt we limit that to `*.yml text=auto` ?

https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvalueauto

> When `text` is set to "auto", the path is marked for automatic end-of-line conversion. **If Git decides that the content is text**, its line endings are converted to LF on checkin. When the file has been committed with CRLF, no conversion is done.

But that just means `core.autocrlf=true` for any git user that has not set that in their local config.

bnkai:
we could also add 
```gitattributes
*.py    diff=python
*.md    diff=markdown
```

@peolic do you want to update the pr with text auto option ?

bnkai:
it doesn't seem <https://github.com/git/git/blob/master/userdiff.c> that `diff=yaml` is supported/valid , well we can try it anyway.

PR #267, opened 2020-11-09
Remove Cloudflare errors from JAVLibrary

halorrr:
Since more and more scrapers have been running into issues with Cloudflare I did some reading up on how their scraper protection actually functions.

You can see the sample error JAV was getting here: https://github.com/stashapp/CommunityScrapers/issues/123#issuecomment-706771764
Basically Cloudflare spends 5 seconds testing javascript to make sure it works as a scraper prevention method.

So anytime we get a response like this on a scraper:
image

It can be fixed by adding CDP with a 5 second sleep to the scraper.

```YAML
driver:
  useCDP: true
  sleep: 5
```

If a CAPTCHA is hit, that is a whole other story, but this Javascript challenge page is apparently the most common Cloudflare page that scrapers run into. (Info source https://github.com/Anorov/cloudflare-scrape)

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -28,10 +28,10 @@ movieByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: 
+      Title:
         selector: //div[@id="video_id"]/table/tbody/tr/td[@class="text"]/text()
       # Using URL to still get a URL with sceneByFragment if a duplicate exist.
-      URL: 
+      URL:
         selector: //div[@class="videos"]/div/a/@title[not(contains(.,"(Blu-ray"))]/../@href|//meta[@property="og:url"]/@content
         postProcess:
           - replace:
@@ -55,6 +55,8 @@ xPathScrapers:
         selector: //div[@id="video_jacket"]/img/@src
         postProcess:
           - replace:
+            - regex: (http:|https:)
+              with:
             - regex: ^
               with: "https:"
       Studio:
@@ -65,7 +67,7 @@ xPathScrapers:
       Director: //div[@id='video_director']/table/tbody/tr/td[@class="text"]/span/a/text()
       Duration: //div[@id="video_length"]/table/tbody/tr/td/span[@class="text"]/text()
       Date: //div[@id="video_date"]/table/tbody/tr/td[@class="text"]/text()
-      Synopsis: 
+      Synopsis:
         selector: //div[@id="video_title"]/h3/a/text()
         postProcess:
           - replace:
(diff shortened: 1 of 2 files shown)


halorrr:
CC @Belleyy Since you have done most of the work on this scraper.

Belleyy:
> 
> 
> CC @Belleyy Since you have done most of the work on this scraper.

I don't remember correctly but i think i tested this way, but this didn't work. 🤷‍♂️
- maybe it was the captcha that block it.
- maybe it was a other site, i try with a sleep 10 sec and was still stuck...

I can't test right now since it's been a while since i got this cloudflare block.
##### Kind of annoying to wait 5sec if you don't have issue 😋

bnkai:
@halorrr i can confirm what @Belleyy mentioned above. I already tried that . The only thing i noticed is that by using CDP and maybe 5 or more secs of sleep time it takes more time for cloudflare to block.
Changing the sleep to 5 though doesn't hurt apart from the user having to setup CDP and waiting a few more secs for every scraper.
@halorrr  can you add some comments to the scraper explaining why CDP is needed there and update the `needs` part of the scraper in the scrapers list ?

halorrr:
Yeah I did more testing and in my own tests it look me 37 tries of spamming the scraper to get Captcha blocked. The Captcha blocks are based on machine learning patterns that Cloudflare's systems come up with so they can be very different for everyone based on unique circumstances (VPN, country, usage speed, device, etc). The 5 second sleep _should_ always allow us to at least pass the Javascript challenge though.

scrapers/javlibrary.yml
```diff
@@ -77,7 +79,19 @@ xPathScrapers:
         selector: //div[@id="video_jacket"]/img/@src
         postProcess:
           - replace:
+            - regex: (http:|https:)
+              with:
             - regex: ^
               with: "https:"
 
-# Last Updated October 22, 2020
+# Switched to CDP due to Cloudflare errors Javascript errors.
```
bnkai, line 87:
```suggestion
# Switched to CDP due to Cloudflare and Javascript errors.
```

scrapers/javlibrary.yml
```diff
@@ -77,7 +79,19 @@ xPathScrapers:
...
+# The 5 second sleep in combination with the CDP driver should prevent
+# Standard Cloudflare errors in regards to their Javascript check.
+# There is still the possibliy of getting a CAPTCHA block based on Cloudflare's
```
bnkai, line 90:
```suggestion
# There is still the possibility of getting a CAPTCHA block based on Cloudflare's
```

scrapers/javlibrary.yml
```diff
@@ -77,7 +79,19 @@ xPathScrapers:
...
+# analasys of your traffic pattern. Nothing can be don't about that besides
```
bnkai, line 91:
```suggestion
# analysis of your traffic pattern. Nothing can be done about that besides
```

PR #270, opened 2020-11-10
Clips4Sale: Update

peolic:
- add studio
- use `common` for `.clipInfo` selector
- optimize date
- concat details by two empty lines

✅ tested

### DIFF
--- a/scrapers/Clips4Sale.yml
+++ b/scrapers/Clips4Sale.yml
@@ -6,28 +6,36 @@ sceneByURL:
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
+    common:
+      $clipInfo: //div[@class="clipInfo clip_details"]
+      $studio: //span[@class="clip_detail_label"][contains(text(),"From:")]/following-sibling::a
     scene:
       Title: //div[@class="clipTitle"]/h3/text()
       Details:
         selector: //div[@class="clip"]/div/p//text()
-        concat: " "
+        concat: "\n\n"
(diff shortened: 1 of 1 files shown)


bnkai:
Look ok.
Should we also replace the concat in the details section ?
```yaml
      Details:
        selector: //div[@class="clip"]/div/p//text()
        concat: "\n\n"
        postProcess:
          - replace:
              - regex: "Description:"
                with:
```
I think concating by `\n\n` instead of the space maintains better the original form of the description text.

peolic:
>I think concating by \n\n instead of the space maintains better the original form of the description text.

Yeah I agree.

Also, I found that the `//text()` selector sometimes discards some of the description.
I changed mine locally to `//div[@class="clip"]/div/p//*`.

To be fair, studios on Clips4Sale can put 'more' than plain text in their video description, which means I can't be sure that the new selector isn't worse for studios other than the ones I regularly scrape.

bnkai:
>To be fair, studios on Clips4Sale can put 'more' than plain text in their video description, which means I can't be sure that the new selector isn't worse for studios other than the ones I regularly scrape.

Yeah i noticed that too in another scraper. 

> Also, I found that the `//text()` selector sometimes discards some of the description.
> I changed mine locally to `//div[@class="clip"]/div/p//*`.

Do you want to update that also or should i merge?

peolic:
> Do you want to update that also or should i merge?

You can go ahead and merge

PR #276, opened 2020-11-14
Create HouseOFyre.yml

trocolodito:


scrapers/HouseOFyre.yml
```diff
@@ -0,0 +1,23 @@
+name: houseofyre
+sceneByURL:
+  - action: scrapeXPath
+    scraper: sceneScraper
+    url:
+      - houseofyre.com/updates
```
peolic, line 6: FYI,
The `/updates/*.html` urls don't seem to have truncated video descriptions in them.
`houseofyre.com/access/scenes` has the full description.

bnkai: @peolic  Are all the rest of data ok?
From  a couple i checked they should be fine?
It might be easier to just use a subscraper for the video description using the canonical than having to construct the urls manually. The site doesnt seem to provide the accesss urls, unless i am mistaken

peolic: @bnkai 
Yeah the data looks alright. Scene cover could be bigger, but it's probably not worth another `subScraper`.
I don't remember how I got to the `/access/scenes/` links even (it was a while ago, maybe they updated the site since)

the regex I would use:
```yml
          - replace:
              - regex: /updates/(.+)\.html$
                with: /access/scenes/$1_vids.html
```

bnkai: @peolic i am getting some weird replace with your replace
unless i use `with: /access/scenes/${1}_vids.html` instead
I traced that to re.ReplaceAllString <https://play.golang.org/p/FyvcrGFcKdt> , does it work as is for you with latest dev ?

peolic: Oh sorry, I didn't try it in Stash, only in JS.
Well, this seems to be the reason:
https://golang.org/pkg/regexp/#Regexp.Expand
> In the $name form, name is taken to be as long as possible: $1x is equivalent to ${1x}, not ${1}x, and, $10 is equivalent to ${10}, not ${1}0.

trocolodito:
Thanks for all your work guys!

PR #277, opened 2020-11-14
Add dates,tweak nympho.yml

bnkai:


SpedNSFW (approved):
All looks good to me. Just need the network to add scene tags now.

PR #288, opened 2020-11-22
Update Nubiles scraper

### LINKED ISSUE #287: Nubiles.yml - Missing url
Add **- familyswap.xxx/** _url_ to _sceneByURL_

I've tested it locally and it parses fine.

peolic:
Add missing sites + sort:
- `familyswap.xxx` (scene + gallery) - Closes #287 
- `girlsonlyporn.com` (scene + gallery)
- `hotcrazymess.com` (scene) - Supersedes #128
- `nubilesporn.com` (scene + gallery)

### Nubiles-Porn network sites vs 'other' Nubiles sites
The latter need slightly different Studio Name selectors,
because the non Nubiles-Porn network sites don't have the site name below the video / gallery
and the page title selector can be incorrect for Nubiles-Porn.
I think the same selector for scenes needs to be used for galleries on the Nubiles-Porn but I haven't tested it yet.

### DRY scraper
The "DRY" (Don't Repeat Yourself) scraper is **an idea**.
I asked for objections on Discord and got none, but I do feel like this **might be** a bit too complex.
The idea behind it is to have less duplicate selectors for sites that are close enough but not exactly the same, which makes updating it easier, IMO.

YAML also supports merging (using `<<` keys), however, while working on this scraper I encountered errors in Stash and it only partially worked, as I noticed Tags and Performers did not scrape.
I suspect deep merging is not fully supported in the Go YAML implementation.
(Should be able to use `<<` without quotes, but it's not supported either)

```yml
xPathScrapers:
  nubileFilmsMultiScraper:
    scene:
      "<<": *sceneScraper
      Studio:
        Name: //div[@class="studio"]  # example
```

### Tests
https://pastebin.com/hxNThKxy

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -223,6 +223,7 @@ famedigital.com|GammaEntertainment.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 family.xxx|trafficpimps.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 familyhookups.com|MindGeek.yml|:heavy_check_mark:|:x:|:x:|:heavy_check_mark:|-|-
 familystrokes.com|PaperStreetMedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
+familyswap.xxx|Nubiles.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|-|-
 familytherapyxxx.com|FamilyTherapyXXX.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 familyxxx.com|FamilyXXX.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fantasyhd.com|AMAMultimedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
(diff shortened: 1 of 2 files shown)


Belleyy:
Could you add `nubilesporn.com/` (It's the same as `nubiles-porn.com`). 
When i go to their site, i never write "-" so the scraper don't work. (I fixed it in local but never did a PR about it)

peolic:
@Belleyy
Added

bnkai:
_The whole problem with these kind of multi scrapers is that sometimes the networks start to migrate to newer layouts and they do it site by site. So some sites start to have problems forcing us to split the scraper again. It already happened a couple of times with Mindgeek / RK being the last one.
Also making a change to an xpath means you have to double check on all the urls present for that scraper.
If the sites have the same exact layout is not a problem but a span here a paragraph there and so on things can get messy.
Thats my main concern with mega scrapers...._

Now to the point while more complex it is also more compact and could be used a lot for scrapers that have scenes/galleries/movies sections with the same layout. I am not that convinced as stated above :-) on creating more of these megascrapers but this can serve as a template / sample and it was already a big one..

All above of course IMHO.

peolic:
Alright, this feels ready to me.
- Tested one scene and one gallery from **each** site manually in Stash: https://pastebin.com/hxNThKxy
- Updated the PR description about why merge keys could not be used (it's a shame, really).
- Split "other scrapers" into Nubile Films and Standalone, because the Nubile Films site contains scenes and galleries from its network, and required a different Studio selector.
- Updated the scene URLs to match `site.com/video/watch/`
  There are `site.com/tube/watch/` URLs, but they seem to lack correct studio info and scene categories (tags).

PR #289, opened 2020-11-23
Add Hegre xPath Scraper

niemands:
As requested in the discord channel

### DIFF
--- a/scrapers/Hegre.yml
+++ b/scrapers/Hegre.yml
@@ -21,6 +21,8 @@ performerByName:
 
 xPathScrapers:
   sceneScraper:
+    common:
+      $actor: //a[@class="record-model"]
     scene:
       Title: //meta[@property="og:title"]/@content
       Details:
@@ -37,9 +39,17 @@ xPathScrapers:
       Tags:
         Name: //div[@class="approved-tags"]/a
       Performers:
-        Name: //a[@class="record-model"]/@title
+        Name: $actor/@title
+        URL:
+          selector: $actor//a/@href
+          postProcess:
+            - replace:
+                - regex: ^
+                  with: https://hegre.com
 
   galleryScraper:
+    common:
+      $actor: //a[@class="record-model"]
     gallery:
       Title: //div[@class="record-toolbar clearfix"]//h1
       Date:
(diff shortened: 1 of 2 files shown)


scrapers/Hegre.yml
```diff
@@ -0,0 +1,96 @@
...
+        postProcess:
+          - parseDate: January 2, 2006
+      Image: //meta[@property="og:image"]/@content
+      Studio:
+        Name:
+          fixed: Hegre
+      Tags:
+        Name: //div[@class="approved-tags"]/a
+      Performers:
+        Name: //a[@class="record-model"]/@title
```
bnkai, line 40: We can add the URL also here eg
```yaml
        URL:
          selector: //a[@class="record-model"]//a/@href
          postProcess:
            - replace:
                - regex: ^
                  with: https://www.hegre.com
```
or also make a common var for the anchor selector as it uses the same with the name.

Adding the URL here makes it easier when you create the performer from the scrape scene/gallery menu as apart from the Name the URL will also be populated and then you just have to use that to scrape using the included performer scraper.

scrapers/Hegre.yml
```diff
@@ -0,0 +1,96 @@
...
+        selector: //span[@class="date"]/text()
+        postProcess:
+          - parseDate: January 2, 2006
+      Studio:
+        Name:
+          fixed: Hegre
+      Tags:
+        Name: //div[@class="approved-tags"]/a
+      Performers:
+        Name: //a[@class="record-model"]/@title
```
bnkai, line 55:
```yaml
        URL:
          selector: //a[@class="record-model"]/@href
          postProcess:
            - replace:
                - regex: ^
                  with: https://www.hegre.com
```
a common part can also be used for the url, name anchor selector

bnkai:
Looks good. 
What is the verdict for single named performers specific to a site? Shoud we add the studio in some form with the name or wait till stash performers alias functionality is more complete and revisit the issue then?

niemands:
> 
> 
> Looks good.
> What is the verdict for single named performers specific to a site? Shoud we add the studio in some form with the name or wait till stash's performers alias functionality is more complete and revisit the issue then?


I would not include the studio in the actor name, except maybe in the alias field, as this makes matching more difficult. I have also added the actor selector rather optionally for actors who are exclusive to the site, for whom you can't find information elsewhere, so that you have at least the little information the site has to offer.

PR #291, opened 2020-11-25
Update SARJ-LLC to pull wide scene covers and allow galleryByURL

DampToast:
This pull request does the following two things:

1. Previously for scenes it pulled the portrait cover.  After talking to a few people in Discord it sounded like wide covers would be preferred.  The URL for the wide cover is embedded inside an attribute with other text, but if you take the 'cover' url and replace 'cover' with 'wide' it points to the same place.  So I just used a regex to do that.

2. With 0.4.0 galleryByURL was added.  I've tested this on several galleries both recent and from way back at the sites inception and it seems to work.  I've only tested it on one of the listed sites but it stands to reason that if the scene code works across sites the gallery likely will as well since it's all the same backend.

I did run the validator and it checked out as well.

### DIFF
--- a/scrapers/SARJ-LLC.yml
+++ b/scrapers/SARJ-LLC.yml
@@ -41,7 +41,12 @@ xPathScrapers:
         Name: $data//a[@href[contains(.,'/model/')]]/text()
       Tags:
         Name: //div[@class='tag-list']//a/text()
-      Image: //div[@class='img-container']/img/@src
+      Image:
+        selector: //div[@class='img-container']/img/@src
+        postProcess:
+          - replace:
+            - regex: cover
+              with: "wide"
       Studio:
         Name: //meta[@property='og:site_name']/@content
 
--- a/scrapers/SARJ-LLC.yml
+++ b/scrapers/SARJ-LLC.yml
@@ -60,16 +60,19 @@ xPathScrapers:
         Name: //div[@class='tag-list']//a/text()
       Image:
         selector: //div[@class='img-container']/img/@src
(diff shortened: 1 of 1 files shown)


scrapers/SARJ-LLC.yml
```diff
@@ -41,8 +58,31 @@ xPathScrapers:
...
+            - regex: cover
+              with: "wide"
       Studio:
         Name: //meta[@property='og:site_name']/@content
-
-# Last Updated October 9, 2020
+  galleryScraper:
+    gallery:
+      Title: //div[@class='info-container']/h3/a/text()
+      Details: //div[@class='description-text']/p/text()
```
bnkai, line 72: in some galleries there are multiple paragraphs eg `https://www.thelifeerotic.com/model/zeo/gallery/20200920
/DRY_LEAVES_1` a concat should be used.

```yaml
        selector: //div[@class='description-text']/p/text()
        concat: "\n\n"
```

bnkai:
The `wide` trick doesnt seem to work with most of the `https://www.lovehairy.com` scenes i tried. There is no `cover` in the image. Since  it doesnt break them though it seems ok to me. Can you add a comment in the replace regex so anyone that wants the covers can spot it easier ?

DampToast:
It looks like lovehairy has 'clean' in the cover image filename instead.  I can add a regex to look for (cover|clean) if you like.  I can also update with the comment you suggested.

DampToast:
@bnkai I've made the changes you suggested.  thanks for your input.

PR #293, opened 2020-11-29
add shagmag gallery scraper

SpedNSFW:
Some things to note for this one.

 - The title is always a date in the format of `MMMM YYYY` (eg. `January 2020`)
 - There is only one upload per month, so we assume the first of the month so we can get a usable date (in reality it can come at any point in the month)
 - "Julia Rose" is present on all uploads however her name is not listed on any except the very first so we manually add her
   - This causes a small problem where she gets added twice to that very first one but it doesn't break anything
 - None of the other content on this site has dedicated pages so none of the videos are scrape-able.

### DIFF
--- a/scrapers/shagmag.yml
+++ b/scrapers/shagmag.yml
@@ -1,0 +1,33 @@
+name: 'SHAGMAG'
+galleryByURL:
+  - action: scrapeXPath
+    url:
+      - shagmag.com/issues/
+    scraper: galleryScraper
+
+xPathScrapers:
+  galleryScraper:
+    gallery:
+      Title: //h1[@class="t-Page-Heading-1"]/text()
+      URL: //link[@rel="canonical"]/@href
+      Studio:
+        Name:
+          fixed: 'SHAGMAG'
+      Date:
+        selector: //h1[@class="t-Page-Heading-1"]/text()
+        postProcess:
+          - replace:
+            # There's one post per month so just assume the 1st so we can get a usable date
+            - regex: ^
+              with: '1 '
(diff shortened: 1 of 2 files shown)


scrapers/shagmag.yml
```diff
@@ -0,0 +1,33 @@
...
+        Name:
+          fixed: 'SHAGMAG'
+      Date:
+        selector: //h1[@class="t-Page-Heading-1"]/text()
+        postProcess:
+          - replace:
+            # There's one post per month so just assume the 1st so we can get a usable date
+            - regex: ^
+              with: '1 '
+          - parseDate: 2 January 2006
```
peolic, line 23: Sorry I didn't notice this until it was merged,
but I found the official published date on the page source,
though I don't have an example where it's **not** the first day of the month.
```yml
      Date:
        selector: //script[@type="application/ld+json"]
        postProcess:
          - replace:
              - regex: ^.+[,{]"datePublished":"(.+?)T[\d:+-]+"[},].+$
                with: $1
          - parseDate: 2006-01-02
```

bnkai: Do you want to make a PR about it ? i kinda overlooked it too since its a magazine and they usually have fixed release dates anyway.

SpedNSFW: Oh nice. I couldn't see any dates on the page and didn't think to check within the script elements.

peolic: If it's mostly the first day of each month anyway it doesn't need the extra complexity.
So no PR from me.

PR #296, opened 2020-12-01
Add StudioFOW xPath scene scraper

bnkai:
as requested from #73


Belleyy:
For content for Scraper list, you could say SFM Porn or 3D Hentai

bnkai:
@Belleyy yeah i wasnt sure if it classifies as hentai as the movies are in english. Is 3D Animation an option ? SFM is also a good one.

Belleyy:
> 
> 
> @Belleyy yeah i wasnt sure if it classifies as hentai as the movies are in english. Is 3D Animation an option ? SFM is also a good one.

3D Animation could work and it's more general term than SFM.

PR #299, opened 2020-12-03
Update KellyMadisonMedia.yml (Cookie Update)

Belleyy:
Current XPath was incorrect, site changed recently ?

- Look like you don't need CDP anymore, should i remove it from the scraper and in `SCRAPERS-LIST.md` ?
   But if you use it headless chrome, you already don't have problem with cookie so i guess it should stay.
   So `useCDP:` should be in `False` or `True` by default ?

- I am not sure about the comment on top of file. Don't really know how to explain correctly.


Belleyy:
I am not sure about the cookies so need people to test it 😃

bnkai:
Seems to work ok without CDP. 
Cookies seem to work with nothing changed (value=test) :-)  so need to mention any instructions...

Can you remove the `useCDP` altogether since its not used anymore ( CDP from the scrapers needs list also ) ?

PR #301, opened 2020-12-03
Mgstage Scraper (Xpath + Cookie 🍪)

Belleyy:
JAV scraper, most of text is japanese.


bnkai:
@Belleyy i've been trying to  test this for a couple of days but the mstage site is down for me. 
Does it work for you now? (I remember the site was ok some time ago)

PR #307, opened 2020-12-09
added czechcasting.com

bonobosexparty:


scrapers/czechcasting.yml
```diff
@@ -0,0 +1,23 @@
...
+      Title: /html/head/title/text()
+      Details: //*[@id="girl-info"]/div/div/div[2]/p/text()
+      Tags:
+        Name: //*[@id="girl-info"]/div/div/ul[@class='tags']/li/a[@class='tag']/text()
+      Studio:
+        Name: $studio
+      Image:
+        selector: //video/@poster
+      Date:
+        selector: //p[@class='updated']/text()
```
bnkai, line 21: the date need to be in the format that stash expects so we need some extra postprocess
```suggestion
      Date:
        selector: //p[@class='updated']/text()
        postProcess:
          - parseDate: 01/02/2006
```

scrapers/czechcasting.yml
```diff
@@ -0,0 +1,23 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $studio: //*[@id="logo"]/a/img/@alt
+    scene:
+      Title: /html/head/title/text()
+      Details: //*[@id="girl-info"]/div/div/div[2]/p/text()
+      Tags:
+        Name: //*[@id="girl-info"]/div/div/ul[@class='tags']/li/a[@class='tag']/text()
```
bnkai, line 15:
```yaml
        Name: //ul[@class='tags']/li/a[@class='tag']/text()
```
is enough i think

scrapers/czechcasting.yml
```diff
@@ -0,0 +1,23 @@
...
+    url:
+      - czechcasting.com/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $studio: //*[@id="logo"]/a/img/@alt
+    scene:
+      Title: /html/head/title/text()
+      Details: //*[@id="girl-info"]/div/div/div[2]/p/text()
```
bnkai, line 13:
```yaml
      Details: //div[@class="description show-desc"]
```
a bit cleaner

bnkai:
Idk if it makes sense to add the performer names also in a page like this
```yaml
      Performers:
        Name:
          selector: //div[@id="girl-info"]//span[@class="name"]
          split: "/"
```
something like this seems to work ( for more than one performers in a scene  i only found one case that are seperated by `/` )

Again i am not sure if it makes sense to add this.

scrapers/czechcasting.yml
bonobosexparty: Ah that makes sense. not sure why i was being overly specific

scrapers/czechcasting.yml
bonobosexparty: added your suggestion

scrapers/czechcasting.yml
```diff
@@ -0,0 +1,25 @@
+name: czechav
```
bnkai, line 1: Just noticed that `czechav` name is used from the `czechav.yml` scraper.
Can you rename to `czechcasting` or something else?

PR #310, opened 2020-12-11
Fix TrafficPimps Scraper

SpedNSFW:
There are quite a few changes here. CDP is required to get the cover image.

Just a couple notes regarding the URL changes:

 - The pimp.xxx, family.xxx, and petite.xxx URLs redirect to cherrypimps.com and are thus pointless.
 - The wildoncam.com URL got commented out as the site has not updated to the new layout that is present on cherrypimps.com and thus none of the selectors will work. Wild On Cam scenes _are_ present on cherrypimps.com however so if they need to be scraped, use the CherryPimps site.

### DIFF
--- a/scrapers/trafficpimps.yml
+++ b/scrapers/trafficpimps.yml
@@ -2,43 +2,30 @@ name: trafficpimps
 sceneByURL:
   - action: scrapeXPath
     url:
-      - pimp.xxx/trailers/
-      - family.xxx/trailers/
-      - petite.xxx/trailers/
       - cherrypimps.com/
-      - wildoncam.com/
+      # - wildoncam.com/
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
-    common:
-      $infoblock: //div[@class="info-block"]
-      $infodata: //div[@class="info-block_data"]
     scene:
       Title:
-        selector: //h1[@class="trailer-block_title"]
-        postProcess:
-          - replace:
-              - regex: "LIVE$"
-                with:
+        selector: //div[@class="item-title"]/h1/text()
       Date:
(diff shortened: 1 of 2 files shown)


SpedNSFW:
I have removed the three `.xxx` sites from the scrapers list. I have left Wild On Cam there as I suspect the site will eventually be updated to the new layout at some point.

scrapers/trafficpimps.yml
```diff
@@ -2,43 +2,30 @@ name: trafficpimps
...
       Studio:
-        Name: //div[@class="container"]/a[@class="logo"]/@title
-      Image: 
-        selector: //script[contains(.,"contentthumbs")]/text()|//img[@class="lazyload update_thumb thumbs stdimage"]/@src
-        postProcess:
-          - replace:
-              - regex: (.+)(https.+contentthumbs.+)(" width="100%)(.+)
-                with: $2
+        Name: //a[@class="series-item-logo"]/img/@alt
+      Image: //div[@class="player-thumb"]/img/@src
```
peolic, line 27: I found this, does it work without CDP?
```suggestion
      Image: //div[@class="player-thumb"]/img/@src0_1
```

SpedNSFW: No. Tried both.

peolic: Sorry! Stupid typo (missed the `x` at the end). Also made sure it worked in Stash this time.
```suggestion
      Image: //div[@class="player-thumb"]/img/@src0_1x
```

SpedNSFW: I had tried both `@src` and `@src0_1x` in Stash and neither worked until I added the CDP. See image below of my returned results using `@src0_1x`
image

PR #319, opened 2020-12-18
added gallery scraping support for Brazzers.yml

venturevoom:


scrapers/Brazzers.yml
```diff
@@ -1,4 +1,7 @@
+###############################################################################################
```
bnkai, line 1: Can you remove these `######` comments?
We don't use them in the rest of the scrapers so its probably better to keep the structure the same for all.

scrapers/Brazzers.yml
```diff
@@ -205,4 +217,23 @@ xPathScrapers:
...
+  galleryScraper:
+    common:
+      $titlebar: //*[name()='path'][1]/../../../../h2/..
+    gallery:
+      Title: $titlebar/h2/text()
+      Studio:
+        Name:
+          selector: //div[text()="Subsite"]/../a/text()
+      Date:
+        selector: $titlebar/div/text()
```
bnkai, line 229: you forgot to parse the date
```yaml
        postProcess:
          - parseDate: January 2, 2006
```

scrapers/Brazzers.yml
```diff
@@ -205,4 +217,23 @@ xPathScrapers:
 #        fixed: "Female"
       URL: //link[@rel="canonical"]/@href
 
-# Last Updated November 8, 2020
+  galleryScraper:
```
bnkai, line 220: do you mind moving the `galleryScraper` section before the performer one and after the scene ? just for cosmetic reasons

PR #320, opened 2020-12-20
data18

MrX292:
wrote it in April and forgot about it.
not sure why i commented some of it out


Belleyy:
- You need to update `SCRAPERS-LIST.md`
- Use PostProcess ( #114 )

I don't think you should do a performer scraper if you only scrape the name. But look like you can get more information (Height, Country, Ethnicity...)

scrapers/data18.yml
```diff
@@ -0,0 +1,48 @@
...
+    url: 
+      - data18.com/
+    scraper: performerScraper
+sceneByURL:
+  - action: scrapeXPath
+    url: 
+      - data18.com/content
+    scraper: sceneScraper
+xPathScrapers:
+  performerScraper:
```
bnkai, line 13: like what @Belleyy mentioned we should either remove the performer scraper (since it only gets the name) or add more data to it

scrapers/data18.yml
```diff
@@ -0,0 +1,48 @@
...
+        #selector: //li[text()='Date of Birth: ']//span
+        #replace:
+          #- regex: ","
+            #with:
+        #parseDate: Jan 2, 2006
+      #Measurements: //li[text()='Measurements: ']//span
+  sceneScraper:
+    common:
+      $performer: //p[b[text()='Starring:']]/a
+      $studio: //p[b[text()='Site:']]/a
```
bnkai, line 26: in some scenes instead of site we have studio
```suggesttion
      $studio: //p[b[text()='Site:' or text()='Studio:']]/a
```

scrapers/data18.yml
```diff
@@ -0,0 +1,48 @@
...
+    scene:
+      Title: //h1/text()
+      Date: 
+        selector: //span[contains(text(),'Release date:')]/a
+        replace: 
+          - regex: ","
+            with:
+        # reference date is: 2006/01/02
+        parseDate: January 2 2006
```
bnkai, line 35: this doesnt really work
this is better
```yaml
      Date:
        selector: //span[contains(text(),'Release date:')]
        postProcess:
          - replace:
              - regex: "^Release date:"
                with:
              - regex: ","
                with:
          - parseDate: January 2 2006
```
but in many scenes the `day` part of the date is missing so should we regex replace that also or leave it alone ?

scrapers/data18.yml
```diff
@@ -0,0 +1,48 @@
...
+      Details: //p[b[text()='Story:']]
```
bnkai, line 36: this needs a bit of pp
```suggestion
      Details:
        selector: //p[b[text()='Story:']]
        postProcess:
          - replace:
              - regex: "^Story:"
                with:
              - regex: �
                with: "'"
```

scrapers/data18.yml
```diff
@@ -0,0 +1,48 @@
...
+      Performers:
+        Name: $performer
+        URL: $performer/@href
+      Studio:
+        Name: $studio/text()
+        URL: $studio/@href
+      #Image:
+        #selector: //div[p[span[b[text()='Photo Set']]]]//div[1]/a/@href
+        #subScraper:
+          #selector: //div[@id='post_view']/a/img/@src
```
bnkai, line 48:
```yaml
      Image:
        '//img[@style="width: 640px; height: 360px;"]/@src'
```
this seems to work ok but might need a bit more testing for false positives

bnkai:
Also a
```yaml
# Last Updated December 20, 2020
```
comment at the end of the file is recommended

scrapers/data18.yml
MrX292: did that to get one from the Photo Set they seem to be higher resolution

MrX292:
removed the performer part and added the date comment

scrapers/data18.yml
bnkai: the photo set is not available in mostl scenes

peolic: I would also try `//div[@id="moviewrap"]/img/@src` which targets the same element.

PR #322, opened 2020-12-22
Update R18 (sceneByFragment + censor)

Belleyy:
Same regex as Javlibrary but added 1 more step because R18 search is worst than Javlibrary
- It doesn't know PPPD148 but PPPD-148.

For the xpath, it compare the search text to the name of the Movies to avoid mistake
-  HNDS-057 != HND-057 (https://www.r18.com/common/search/searchword=HND-057/)

Changed the name to warn people it's for JAV, because "R18" is not really explicit.


Belleyy:
Have problem with the compare of string, i don't really know if this work. (Finding HND-057 don't work but for other scene don't have any issue...)
Replacing by `string-length` look working but can be inaccurate.

Belleyy:
- Added a regex to replace censored word.

PR #326, opened 2020-12-27
Update DorcelClub.yml

venturevoom:
Repaired the broken Scraper. Thanks to Belley for helping on IMG support.

### DIFF
--- a/scrapers/DorcelClub.yml
+++ b/scrapers/DorcelClub.yml
@@ -23,5 +23,6 @@ xPathScrapers:
           fixed: DorcelClub
       Movies:                                      
         Name: //div[@class="left"]/span[@class="movie"]/a/text()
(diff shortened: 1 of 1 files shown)


scrapers/DorcelClub.yml
```diff
@@ -1,36 +1,27 @@
...
       Performers:
-        Name: $info/div[@class="actors"]/a
+        Name: //div[@class="actress"]/a
       Studio:
         Name:
           fixed: DorcelClub
-      Movies:
-        Name: //div[@class="movie"]/a/text()
+      Movies:                                      
+        Name: //div[@class="left"]/span[@class="movie"]/a/text()
```
bnkai, line 26: The movie URL can also be added since we are updating the scraper
```suggestion
        URL: //div[@class="left"]/span[@class="movie"]/a/@href
```

PR #328, opened 2020-12-28
Added PerformerByURL scraper, feminized.com to FemdomEmpire

spacepilothannah:
feminized.com is run by the same people as femdomempire, using the same
CMS.

From their logo & its title attribute it's clear the site's name is Femdom Empire,
not FemDom Empire, so I renamed the yml file too.

I thought about adding performer stats to the scraper, but it's pretty
hard due to the way their tags work. Possible with XPath 2.0 (and heavy
use of regexes) but only present on Femdom Empire - and I'm not really
bothered about it, so I haven't looked into whether the XPath parser in
the scraper supports XPath 2.

A little bit more rationale - I didn't add a performerByName as there's no
performer search feature. The only reason for adding PerformerByURL
is that there's a few performers on feminized.com that aren't on stash-box,
nor freeones. It does make for a slightly tedious import workflow, but it still
beats going and finding images by hand.

The Gender thing is a little weird - on FemdomEmpire only the dommes are
featured under the performers section of the site, whereas on Feminized all
performers are listed, but all the men have their name as "Sissy <name>".
The title field of the performer info box always has "Mistress" at the top - so
the regex matches for "Mistress Sissy <name>" as a kind of double-check.
It is pretty unnecessary and it is only a proxy for the performer's gender (in
particular, on feminized half the performers are either men or trans women,
so "Female" doesn't strictly fit them given the existence of the 
"Transgender Woman" category.) It's a good enough heuristic for now though.
I double-checked and no one who has performed for Femdom Empire has 
"Sissy" in their name.

**Quasi-unrelated soapbox time**

All that said, as a trans woman myself I'd rather that "gender" was free text,
that tags were used to indicate peoples' trans status, and that the Gender
field was filled in based on performers' self-identified gender, rather than the
databases we have putting people into narrow boxes. ¯\_(ツ)_/¯

To be fair, I am also happy that I haven't seen the word "Shemale" anywhere
in Stash. So that's a plus 😄


scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,63 @@
...
+          selector: //img[@class="update_thumb thumbs stdimage"]/@src0_1x
+          postProcess:
+            - replace:
+              - regex: .*/tour/content//
+                with: "https://feminized.com/tour/content/"
+  performerScraper:
+    common:
+      $performerInfo: "//div[@class='contentBox'][1]"
+      # if stash scraper supports xpath 2 we might be able to parse measurements etc from femdomempire stats lists.
+      # $profileAbout: "$performerInfo/div[@class='profile-about']/p"
```
bnkai, line 43: I wouldnt even bother with how the tags are inserted in the stats list.
I would also remove the comment

scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,63 @@
...
+    performer:
+      Name:
+        selector: "$performerInfo/div[@class='profile-details']/h3"
```
bnkai, line 46: the selector need  an exra `/` after the $performerInfo to work
```yaml
  selector: "$performerInfo//div[@class='profile-details']/h3"
```

scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,63 @@
...
+          - replace:
+            - regex: "^Mistress Sissy .*$"
+              with: Male
+          - replace:
+            - regex: "^Mistress (.*)$"
+              with: Female
+      Image:
+        selector: "//li[@class='flex-active-slide']/img/@src"
+      URL:
+        selector: "//link[@rel='canonical']/@href"
```
bnkai, line 59: there is no canonical link in the page so it doesnt do anything

scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,63 @@
...
+        selector: "$performerInfo/div[@class='title']/h3"
+        postProcess:
+          - replace:
+            - regex: "^Mistress Sissy .*$"
+              with: Male
+          - replace:
+            - regex: "^Mistress (.*)$"
+              with: Female
+      Image:
+        selector: "//li[@class='flex-active-slide']/img/@src"
```
bnkai, line 57: The image selector isnt present in the page
Something like this
```yaml
      Image:
        selector: //base/@href|(//div[@class="flexslider-carousel"]//img/@src0_1x)[1] # we use the union | and concat to get the base url for the image
        concat: "|"
        postProcess:
          - replace:
            - regex: "(^[^|]+)\\|/tour/([^\\.]+\\.jpg).*"
              with: $1$2
```
seems to work (havent tested thoroughly)

scrapers/FemdomEmpire.yml
spacepilothannah: Yeah. I was under the impression from some other piece of documentation that the `URL` has to exist - this is my first time adding a scraper so I wasn't sure where to look for the ByURL-type scrapers. I'll remove it :-)

scrapers/FemdomEmpire.yml
spacepilothannah: good spot! Will fix

scrapers/FemdomEmpire.yml
spacepilothannah: fair enough. I've no real interest in the measurements thing anyway.

scrapers/FemdomEmpire.yml
spacepilothannah: It is, on feminized - I'll double check.

scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,63 @@
...
+        Name: //ul/li[@class="label" and contains(.,"Categories:")]/..//a
+      Studio:
+        Name:
+          selector: //div[@id='logo']/h1/a/span
+        Image:
+          selector: //img[@class="update_thumb thumbs stdimage"]/@src0_1x
+          postProcess:
+            - replace:
+              - regex: .*/tour/content//
+                with: "https://feminized.com/tour/content/"
```
bnkai, line 38: the image was shifted a couple of spaces breaking the scraper
and it also need to be adjusted to work with both sites
```yaml
      Image:
        selector: //base/@href|//img[@class="update_thumb thumbs stdimage"]/@src0_1x
        concat: "|"
        postProcess:
          - replace:
              - regex: "(^[^|]+)\\|/tour/([^\\.]+\\.jpg).*"
                with: $1$2
```
seeems to do the job

spacepilothannah: Ah yeah - my bad, I half-baked it.

scrapers/FemdomEmpire.yml
spacepilothannah: Aaah, yep it's a javascript thing - that selector isn't present in the page until the JS fires, I get it 😄 

Thanks for the suggested fix!

scrapers/FemdomEmpire.yml
```diff
@@ -0,0 +1,66 @@
...
+        # we use the union | and concat to get the base url for the image
+        selector: //base/@href|(//div[@class="flexslider-carousel"]//img/@src0_1x)[1] 
+        concat: "|"
+        postProcess:
+          - replace:
+            - regex: "(^[^|]+)\\|/tour/([^\\.]+\\.jpg).*"
+              with: $1$2
+
+
+# Last Updated October 10, 2020
```
bnkai, line 65: A last change to update the `Last Updated` date to today

spacepilothannah: 🤦‍♀️  I feel a fool now

bnkai: it happens to all of us :-P

PR #329, opened 2020-12-31
Updating GirsOutWest scraper.

Tweeticoats:
Adding gallery scraper. To use go to the performer's page and find the gallery url.
This site now has an ok button before accessing the site so this needs the single click support.


bnkai:
The click added to the site sets a single coookie.
Using this seems to work for me
```yaml
  cookies:
    - CookieURL: "https://tour.girlsoutwest.com"
      Cookies:
        - Name: "numheader"
          Domain: "tour.girlsoutwest.com"
          Value: "5"
          Path: "/"
```

Using cookies should be preffered  imo since 
1. no CDP is required
2. it is a lot faster ( time to load the CDP instance + time to wait after click)

Tweeticoats:
Thanks I did not notice what cookie it used.

PR #331, opened 2021-01-06
Add dickdrainers xPath scraper

niemands:
Requested in #73

### DIFF
--- a/scrapers/Dickdrainers.yml
+++ b/scrapers/Dickdrainers.yml
@@ -1,0 +1,34 @@
+name: Dickdrainers
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - dickdrainers.com/tour/trailers
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="videoDetails clear"]//h3
+      Details:
+        selector: //div[@class="videoDetails clear"]/p//text()[normalize-space(.)]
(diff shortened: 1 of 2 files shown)


scrapers/Dickdrainers.yml
```diff
@@ -0,0 +1,34 @@
...
+                with: $1
+              - regex: ^
+                with: http://dickdrainers.com
+      Studio:
+        Name:
+          fixed: Dick Drainers
+      Tags:
+        Name: //li[@class="label"]/following-sibling::li/a/text()
+      Performers:
+        Name: //li[@class="update_models"]//a/text()[normalize-space()]
```
bnkai, line 32: normalize-space shouldnt be empty it spits a warning as is and performers arent included because of that

scrapers/Dickdrainers.yml
```diff
@@ -0,0 +1,34 @@
...
+          - replace:
+              - regex: ^.+poster="(.+?)".+$
+                with: $1
+              - regex: ^
+                with: http://dickdrainers.com
+      Studio:
+        Name:
+          fixed: Dick Drainers
+      Tags:
+        Name: //li[@class="label"]/following-sibling::li/a/text()
```
bnkai, line 30: the li[@class="label"]  includes the performers along with the tags
```
        Name: //li[@class="label"][contains(.,"Tags:")]/following-sibling::li/a/text()
```
should limit that to only tags

PR #337, opened 2021-01-14
Insex update

nocrad349:
Added some URLs to insex.yml. 
I've tested it on a few scenes for each added URLs and didn't see any problems. 

insex.yml passes the project's validation.

PR #338, opened 2021-01-14
scraper added for topgrl.com

nocrad349:
Basically a copy of insex.yml. Just a few changes to deal with html that differs slightly from the rest of the Insex sites.
Only tested it on a few scenes, but it seems to work. Passes the validator.


bnkai:
For cases like this where most of the selectoes are the same i think its better to edit the original file and use yaml anchors and aliases.
You can use Nubiles.yml  as  a reference, i'll also have a look when possible.

nocrad349:
Changes moved to PR #337

PR #341, opened 2021-01-17
Update WowPorn.yml

jeremymeyers:
added .com versions. they are the same site as .xxx but some videos show .com in the intro so should include.


bnkai:
Are you sure about that?
The .com and  the .xxx  sites use entirely different layout. 
None of the selectors work for the .com sites. The com site only has an image in the scene page and the scene title in the page title. Could it be a dns issue or some use of geolocation ?

bnkai:
Cross checked and site layout seems different. Closing

PR #343, opened 2021-01-17
Update ManyVids.yml

echo6ix:
Added tag scraping for sceneScraper


bnkai:
@echo6ix i already mad a PR with your changes #340 , although i have them commented out as default.
Do you mind closing this and posting in the relevant PR if you think that tags,CDP should be enabled by default?

PR #345, opened 2021-01-18
Update censor list for R18

Belleyy:
Some words are not sure...

For Javlibrary:
- Changed the regex for JG
- Edited the regex for the date (I don't know why but yesterday the regex stopped working)

_RIP your debug log file if you use this scraper because 200 regex replace take some places 😋_

### DIFF
--- a/scrapers/javlibrary.yml
+++ b/scrapers/javlibrary.yml
@@ -313,7 +313,7 @@ xPathScrapers:
       #      - regex: HumB\*{2}d\b
       #        with: "Humbled"
       #      - regex: K\*{2}ler\b
-      #        with: "Keeler" # Not Sure
+      #        with: "Killer" # Not Sure
       #      - regex: J\*{1}\b
       #        with: "Jo" # Not Sure
       #      - regex: J\*{1}s\b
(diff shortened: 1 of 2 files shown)


scrapers/javlibrary.yml
```diff
@@ -122,42 +122,216 @@ xPathScrapers:
...
+      #      - regex: D\*{6}eful\b
+      #        with: "Disgraceful"
+      #      - regex: B\*{3}d\b
+      #        with: "Blood"
+      #      - regex: U\*{7}g\b
+      #        with: "Unwilling"
+      #      - regex: HumB\*{2}d\b
+      #        with: "Humbled"
+      #      - regex: K\*{2}ler\b
+      #        with: "Keeler" # Not Sure
```
nocrad349, line 316: Does Killer make more sense?

Belleyy: Tbh i have no idea, here is the source: 
`https://www.r18.com/videos/vod/movies/detail/-/id=nsps00929/`

As a non native english, if this make more sense for you i will change 😄

bnkai: With a google search all cases with the id presentm as comment should be ok except this one , in various results `killer` is mentioned so that is probably better.

PR #347, opened 2021-01-20
Deviante (+SubSite) XPath Scraper

Belleyy:
I think it's part of MindGeek Network, the layout is mostly the same (But selector is different).

The studio name is general but there is no information of where the scene is really come from (ForgiveMeFather, SexWorking, PrettyDirtyTeens, LoveHerAss, EroticSpice)

Subsite need CDP

~~_Got blocked by the network (Too many request)... Draft atm_~~


bnkai:
Yeah i i've done a scrape on the  deviante site.There is no way to get the studios from the scene page. Even when in a list view the scenes don't always have the studio anchor `icon` so a subscraper won't work.

scrapers/deviante.yml
```diff
@@ -0,0 +1,33 @@
...
+        Name: //h2//a[contains(@href,"/model/")]/text()
+      Studio:
+        Name: 
+          selector: //link[@rel="canonical"]/@href
+          postProcess:
+            - replace: 
+              - regex: .*www\.
+                with: ""
+              - regex: \.com.+
+                with: ""
```
bnkai, line 30: I'm guessing this was mainly used in the subsites scraper? Here its only deviante so a `fixed` name should be used

scrapers/devianteNetwork.yml
```diff
@@ -0,0 +1,49 @@
...
+                forgivemefather: Forgive Me Father
+                loveherass: Love Her Ass
+                prettydirtyteens: Pretty Dirty Teens
+                sexworking: Sex Working
+      Image:
+        selector: //div[contains(@class,"player-poster")]/@style
+        postProcess:
+          - replace:
+            - regex: (.+url\(")(.+jpg)(.+)
+              with: $2
```
bnkai, line 46: this doesnt seem to work with some of the subsites
the script seems to work though
```yaml
      Image:
        selector: //script[@type="application/ld+json"]/text()
        postProcess:
          - replace:
            - regex: '.*"thumbnailUrl": "([^"]+)".*'
              with: $1
```

PR #351, opened 2021-01-27
added scraper for chastitybabes.com

nocrad349:
added xpath scraper for chastitybabes.com

### DIFF
--- a/scrapers/chastitybabes.yml
+++ b/scrapers/chastitybabes.yml
@@ -1,0 +1,47 @@
+name: "ChastityBabes"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - 'chastitybabes.com/archives'
+    scraper: sceneScraper
+    
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performerTitle: //h1[@id="post-title"]/text()
+    scene:
+      Title:
+        selector: $performerTitle
+        postProcess:
+          - replace:
+            - regex: (.*) – (.*)
+              with: $2
+      Date: 
+        selector: //div[@class="post_info"]/text()[1]
+        postProcess:
+          - replace:
+            - regex: '.*Posted on (.*) in.*'
+              with: $1
+          - parseDate: "Jan 2, 2006" 
+      Details: 
+        selector: //div[@class="postcontent"]/p/text()
+        concat: "\n"
+      Performers: 
+        Name: 
+          selector: $performerTitle
+          postProcess:
+            - replace:
+              - regex: (.*) – (.*)
+                with: $1
+      Tags: 
+        Name: //a[@rel="category tag"]/text()
+      Studio:
+        Name:
+          selector: //title/text()
+          postProcess:
+            - replace: 
+              - regex: .* \| (.*)
+                with: $1
+      Image: //img[@class="thumb_img"]/@src
+
+# Last Updated January 27 2021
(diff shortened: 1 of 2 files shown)


scrapers/chastitybabes.yml
```diff
@@ -0,0 +1,47 @@
...
+              with: $2
+      Date: 
+        selector: //div[@class="post_info"]/text()[1]
+        postProcess:
+          - replace:
+            - regex: '.*Posted on (.*) in.*'
+              with: $1
+          - parseDate: "Jan 2, 2006" 
+      Details: 
+        selector: //div[@class="postcontent"]/p/text()
```
bnkai, line 27:
```yaml
        selector: //div[@class="postcontent"]/p
```
Should be used instead as the current one misses  the text from anchors in the paragraph

scrapers/chastitybabes.yml
```diff
@@ -0,0 +1,47 @@
...
+                with: $1
+      Tags: 
+        Name: //a[@rel="category tag"]/text()
+      Studio:
+        Name:
+          selector: //title/text()
+          postProcess:
+            - replace: 
+              - regex: .* \| (.*)
+                with: $1
```
bnkai, line 44: Is there a case where the studio isnt `Chastity Babes` ?
If not then a fixed should be used instead eg
```yaml
          fixed: Chastity Babes
```

nocrad349: I know the parent studio has more sites and the updates on chastitybabes aren't sequentially numbered, so I'm guessing there may be more sites using the same layout. I dont know which though.

scrapers/chastitybabes.yml
nocrad349: Interesting, I would expect that selector to produce html instead of text, but it seems the tags are removed afterall.

Your solution does have a disadvantage though, it removes all linebreaks.
I'll try and see if there is a solution that has both linebreaks and the anchor text later this week.

If I can't find that I think I prefer the odd missing anchor text over missing all linebreaks though. But you decide.

bnkai: Yeah losing the br is the trade-off. Imho full text should be preferred instead of the brs ( but thats a personal preference).
If nothing else works we can have both options with the one commented out.

scrapers/chastitybabes.yml
bnkai: Oh you mean more sites could be added to the scraper later on. Yeah that makes sense.

nocrad349: I checked 14 sites by the same parent studio (all I could find) and they all have different layouts that share barely anything. So I made the studio name fixed as suggested

scrapers/chastitybabes.yml
nocrad349: I can't think of a way to keep the line breaks, so I changed the file to your suggestion.

I have no experience or knowledge of go, but I think the linebreaks are intentionally removed by stash:
https://github.com/stashapp/stash/blob/56b47ff9bed64b2e0c61e0f37e1c20112175fe7e/pkg/scraper/xpath.go#L250
If that ever gets optional we should opt not to remove them for the Details of this scraper.

bnkai: > I have no experience or knowledge of go, but I think the linebreaks are intentionally removed by stash:
> https://github.com/stashapp/stash/blob/56b47ff9bed64b2e0c61e0f37e1c20112175fe7e/pkg/scraper/xpath.go#L250
> If that ever gets optional we should opt not to remove them for the Details of this scraper.
Yeah there is no easy way with the current implementation ( i've looked  at that at least a couple of times ) to exclude only for the description ( it does a good job of cleaning  tiitles,performers,tags ....)

PR #352, opened 2021-01-29
Clips4sale fix

bnkai:
A quick fix, needs testing with more scens/studios
I removed the `performers` since i couldnt find scenes with the performers in tags as stated in the original scraper.


jb19774:
tested 4+ studios, lgtm

peolic:
> I removed the `performers` since i couldnt find scenes with the performers in tags as stated in the original scraper.

Please keep the performers. I can confirm there are studios who put performer names as tags.
For example studios `81593` and `67653`.

bnkai:
Readded performers.
@peolic thanks for the samples , i tried to find some before your post but i wasn't lucky.

nocrad349:
Tested 5 studios, 20 clips, everything looks good.

PR #357, opened 2021-02-03
MindGeek API (Python)

Belleyy:
- You need to set the URL (You don't need to save, just be in the field.)

I am not sure about the UserAgent... (You don't need it to have the request work)
I guess the image can be wrong too... (Maybe if the scene is not full HD ?)

Tested on Python 3.9.1
Tested with Realitykings, Brazzers, DigitalPlayground

-------

If you want to save the JSON, write - logJSON in the .yml

```yaml
sceneByFragment:
    action: script
    script:
      - python
      - MindGeekAPI.py
      - logJSON
```

### DIFF
--- a/scrapers/MindGeekAPI.py
+++ b/scrapers/MindGeekAPI.py
@@ -36,31 +36,6 @@ r = requests.get(api_URL, headers=headers)
 try:
     api_json = r.json().get('result')
     api_json['url'] = fragment["url"]
-    try:
-        # Saving the JSON to a file (Write '- logJSON' below MindGeekAPI.py in MindGeekAPI.yml)
-        if sys.argv[1] == "logJSON":
-            with open(id+".json", 'w', encoding='utf-8') as f:
-                json.dump(api_json, f, ensure_ascii=False, indent=4)
-    except IndexError:
-        pass
-
-    # Time to scrape all data
-    scrape = {}
-    scrape['title'] = api_json.get('title')
-    date = datetime.strptime(api_json.get(
-        'dateReleased'), '%Y-%m-%dT%H:%M:%S%z')
-    scrape['date'] = str(date.date())
-    scrape['details'] = api_json.get('description')
-    scrape['studio'] = {}
-    scrape['studio']['name'] = api_json['collections'][0].get('name')
-    scrape['performers'] = [
-        {"name": x.get('name')} for x in api_json.get('actors')]
-    scrape['tags'] = [{"name": x.get('name')} for x in api_json.get('tags')]
-    # Image can be poster or poster_fallback
-    scrape['image'] = api_json['images']['poster'][0]['xx'].get('url')
-    if '/poster/' not in scrape['image']:
-        scrape['image'] = api_json['images']['poster'][1]['xx'].get('url')
-    print(json.dumps(scrape))
 except:
     print("An error has occurred", file=sys.stderr)
     print(f"Request status: `{r.status_code}`", file=sys.stderr)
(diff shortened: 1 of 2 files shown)


bnkai:
@Belleyy i think you should move the save after the json parsing is done.
That way you could also check if there  are results and only then add/save the URL (url could be wrong  (404) or not valid not only missing )

SpedNSFW:
Tested successfully with sites from:
 - Babes
 - Brazzers
 - Deviante
 - Digital Playground
 - Mofos
 - Reality Kings
 - Twistys

Tested with Python 3.8.1

scrapers/MindGeekAPI.py
```diff
@@ -0,0 +1,82 @@
...
+        f.write("Request:\n{}".format(r.text))
+    exit(1)
+try:
+    # Saving the JSON to a file (Write '- logJSON' below MindGeekAPI.py in MindGeekAPI.yml)
+    if sys.argv[1] == "logJSON":
+        try:
+            os.makedirs('MindGeekAPI_JSON')
+        except FileExistsError:
+            pass  # Dir already exist
+        with open("MindGeekAPI_JSON\\" + id + ".json", 'w', encoding='utf-8') as f:
```
bnkai, line 61: this only works for windows
```python
        filename = os.path.join("MindGeekAPI_JSON", id+".json")
        with open(filename, 'w', encoding='utf-8') as f:
```
the path join should be cross platform safe instead

SpedNSFW: Forward slashes work on Windows (or have for years now at least) but this would be the smarter option.

scrapers/MindGeekAPI.py
```diff
@@ -0,0 +1,208 @@
...
+    if not token:
+        try:
+            r = requests.get(url,timeout=(3, 5))
+        except requests.Timeout:
+            print("Request Timeout", file=sys.stderr)
+            exit(1)
+        id = re.search('(.+releaseId\":\")(.+?)(\".+)',
+                       r.text, re.IGNORECASE).group(2)
+        token = re.search('(.+jwt\":\")(.+?)(\".+)',
+                          r.text, re.IGNORECASE).group(2)
```
bnkai, line 42: this should be `tried`
on first run you wont find any matches for group(2)

bnkai, line 42: if a request is made for a site that has not the above regexes (eg not  in the supported list ) then no match is made and so no group will be available
`AttributeError: 'NoneType' object has no attribute 'group'`
those should be `tried` probably

scrapers/MindGeekAPI.py
```diff
@@ -0,0 +1,249 @@
...
+def get_info(url):
+    todaystr = datetime.today().strftime('%Y-%m-%d')
+    today = datetime.strptime(todaystr, '%Y-%m-%d')
+    token, id = check_config(url, today)
+    if not token:
+        print("No instance token found, sending request...", file=sys.stderr)
+        try:
+            r = requests.get(url, timeout=(3, 5))
+        except requests.Timeout:
+            print_exit("Request Timeout")
```
bnkai, line 43: here you are only excepting for timeout errors
you should another catchall `except:` or just one error print for all errors
You can try this by adding a bogus url eg `not_a_url` to a scene when editing and try scraping

SpedNSFW:
Just as a reminder, you will need to add this to the `SCRAPERS-LIST.md` with Python as a requirement.

bnkai:
@SpedNSFW as the script doesnt implement the scrape by url functionality adding that to the list would probably be a bit misleading. We were thinking of creating a new page/subpage for these kind of scrapers ( the ones that scrape db files also fall into this category). @Belleyy proposed something like this <https://github.com/Belleyy/CommunityScrapers/blob/master/script_list.md>

scrapers/MindGeekAPI.py
SpedNSFW: Catch-all excepts should not be used. Instead use the parent exception thrown by requests: `requests.exceptions.RequestException`

scrapers/MindGeekAPI.py
```diff
@@ -0,0 +1,286 @@
...
+        print("Check your MindGeekAPI.log for more details", file=sys.stderr)
+        with open("MindGeekAPI.log", 'w', encoding='utf-8') as f:
+            f.write("Headers used: {}\n".format(headers))
+            f.write("API URL: {}\n".format(url))
+            f.write("Response:\n{}".format(r.text))
+        sys.exit(1)
+    try:
+        api_json = r.json().get('result')
+    except:
+        print_exit("Error getting the JSON from request", file=sys.stderr)
```
bnkai, line 183: should be
```suggestion
        print_exit("Error getting the JSON from request")
```

scrapers/MindGeekAPI.py
```diff
@@ -0,0 +1,286 @@
...
+    scrape = {}
+    scrape['title'] = api_json.get('title')
+    date = datetime.strptime(api_json.get(
+        'dateReleased'), '%Y-%m-%dT%H:%M:%S%z')
+    scrape['date'] = str(date.date())
+    scrape['details'] = api_json.get('description')
+    if url:
+        scrape['url'] = url
+    scrape['studio'] = {}
+    scrape['studio']['name'] = api_json['collections'][0].get('name')
```
bnkai, line 199: we get an `IndexError: list index out of range ` exception here if collections  dont have entries
eg `https://www.brazzers.com/video/4030401`

Belleyy: Stupid no studio scene 😡 . Good catch 👍

PR #358, opened 2021-02-04
This is a python scraper to read torrent files for metadata.

Tweeticoats:
This is a python scraper to read torrent files for metadata.
If the title field contains the filename of a file downloaded or a .torrent file.
It will look through the configured path for torrent files and see if there are metadata tags for description, tags etc.

This uses https://github.com/7sDream/torrent_parser under the MIT licence (included as torrent_parser.py)

### DIFF
--- a/scrapers/torrent.py
+++ b/scrapers/torrent.py
@@ -1,0 +1,59 @@
+import os
+import sys
+import json
+import torrent_parser as tp
+'''  This script parses all torrent files in the specified directory for embedded metadata.
+     The title can either be a filename or the filename of the .torrent file
+     
+     This requires python3.
+     This uses the torrent_parser library to parse torrent files from: https://github.com/7sDream/torrent_parser
+     This library is under the MIT Licence.
+
+     '''
+
+path='/root/.stash/scrapers/torrents/'
+
+def query(title):
+#    print(f"Test",file=sys.stderr)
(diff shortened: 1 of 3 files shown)


bnkai:
I haven't tested yet is there though a specific reason you chose `title` over `filename` ? If i am not mistaken scene by fragment scrapers should have access to that.

bnkai:
It should be noted that this works for torrents from a specific tracker
Most torrents that  i have available only have a comment section  usually with a url  (or a please seed text ) and a created by /  creation date for metadata

dLCarbonX:
How does this work exactly? I am getting errors

Tweeticoats:
Yes, you give it a json map with a title key with ether the name of a .torrent file or a filename and it will look through torrent files in the directory for the torrent that file came from.
echo '{"title":"test.torrent"}'| python3 torrent.py query

If there are comments, tags or other metadata it can return results but I need to add more error handling and have it return valid json to standard out when something goes wrong.

dLCarbonX:
That still doesn't actually do anything lol
`echo '{"title":"/data/media/pr0n/file.mp4"}' | python3 torrent.py query
{"title": "/data/media/pr0n/file.mp4"}`

bnkai:
Merging this. @Tweeticoats Feel free to make a new PR if you need to make any changes.

PR #362, opened 2021-02-07
Teamskeet API ? (Script JSON)

Belleyy:
Copy paste my MindGeekAPI to Teamskeet.

You need the URL  (`https://www.teamskeet.com/movies/`) then choose "TeamskeetJSON"
Probably avoid to get blocked by Cloudflare.

Tested on Python 3.9.1
Tested on Teamskeet.com

~~Draft atm because the URL don't look like something permanent (**ts-elastic-d5cat0jl5o**)~~

-------

If you want to save the JSON, write - logJSON in the .yml

```yaml
sceneByFragment:
    action: script
    script:
      - python
      - TeamskeetJSON.py
      - logJSON # Save the JSON from the API to a file
```


bnkai:
The `doc` approach seems safer to avoid false positives.
It might be worth it to validate the url before quering eg making sure that it has the `https://www.teamskeet.com/movies/` +`id` form.
That way we avoid spamming the api for wrong requests and we are sure that the url saved if a result is returned is ok.
I think saving to a file the json with id.json would be safe enough ( as you only look for  one site now )

scrapers/TeamskeetJSON.py
```diff
@@ -0,0 +1,80 @@
...
+date = datetime.strptime(api_json.get(
+    'publishedDate'), '%Y-%m-%dT%H:%M:%S.%f%z')
+scrape['date'] = str(date.date())
+scrape['details'] = api_json.get('description')
+scrape['studio'] = {}
+scrape['studio']['name'] = api_json['site'].get('name')
+scrape['performers'] = [{"name": x.get('modelName')}
+                        for x in api_json.get('models')]
+scrape['tags'] = [{"name": x} for x in api_json.get('tags')]
+# Image can be poster or poster_fallback
```
bnkai, line 74: leftover comment from copy/paste ?

scrapers/TeamskeetJSON.py
```diff
@@ -0,0 +1,80 @@
...
+            filename = os.path.join(
+                "Teamskeet_JSON", str(api_json['id'])+".json")
+            with open(filename, 'w', encoding='utf-8') as f:
+                json.dump(api_json, f, ensure_ascii=False, indent=4)
+    except IndexError:
+        pass
+
+
+fragment = json.loads(sys.stdin.read())
+# Get the id and instance token
```
bnkai, line 30: leftover comment?

SpedNSFW:
Just as a reminder, you will need to add this to the `SCRAPERS-LIST.md` with Python v3.6 or later as a requirement.

bnkai:
> Just as a reminder, you will need to add this to the `SCRAPERS-LIST.md` with Python v3.6 or later as a requirement.

As mentioned in the MindGeek api scraper  its probably better to add to a separate section and not in the SCRAPERS-LIST.
If it ends up supporting scrape by URL also then we should propably remove the old `plain` scraper and add to the list.

bnkai:
@Belleyy 
Should we use this as the default scraper for Teamskeet? The one we have now is probably useless.
You would only need to rename your yml file to match the one we have now
TeamskeetJSON.yml -> Teamskeet.yml
and use scenebyurl instead of fragment (maybe change the name to Teamskeet also)
```yaml
name: "Teamskeet"
sceneByURL:
  - url:
      - teamskeet.com/movies/
    action: script
    script:
      - python
      - TeamskeetJSON.py
      # - logJSON # Save the JSON from the API to a file
# Last Updated February 07, 2021
```

Belleyy:
> 
> 
> @Belleyy
> Should we use this as the default scraper for Teamskeet? The one we have now is probably useless.

I am for it, because i can't manage to get any scene with the Xpath, and many people seem to have the same problem...
So it's probably easier to have a working method that work for everyone (just need python with requests...) 👍

SCRAPERS-LIST.md
```diff
@@ -661,7 +661,7 @@ sweetyx.com|SweetyX.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 taboopov.com|taboopov.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 teacherfucksteens.com|Nubiles.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|-|-
 teachmyass.com|Teencoreclub.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
-teamskeet.com|TeamSkeet.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
+teamskeet.com|Teamskeet.yml|:heavy_check_mark:|:x:|:x:|:x:|Python (with requests)|-
```
SpedNSFW, line 664:
```suggestion
teamskeet.com|Teamskeet.yml|:heavy_check_mark:|:x:|:x:|:x:|Python >= 3.6|-
```

Keeps it inline with the other Python scrapers.

SpedNSFW (changes requested):

PR #364, opened 2021-02-08
Update vixen scraper

SpedNSFW:
This commit replaces the old XPath scraper with a new script scraper. This eliminates the need for the CDP.
The script performs faster than the existing scraper (due to no CDP) and even includes scene tags.

Python v3.3 or later is required.

### DIFF
--- a/scrapers/vixenNetwork.py
+++ b/scrapers/vixenNetwork.py
@@ -35,12 +35,15 @@ def fetch_page_json(page_html):
     return None if len(matches) == 0 else json.loads(matches[0])
 
 
-def save_json(scraped_json, video_id):
+def save_json(scraped_json, video_id, save_location):
+    location = 'VixenNetwork_JSON'
+    if save_location != 'save':
+        location = save_location
     try:
-        os.makedirs('VixenNetwork_JSON')
+        os.makedirs(location)
     except FileExistsError:
         pass
-    filename = os.path.join('VixenNetwork_JSON', '%s.json' % video_id)
+    filename = os.path.join(location, '%s.json' % video_id)
     with open(filename, 'w', encoding='utf-8') as f:
         json.dump(scraped_json, f, ensure_ascii=False, indent=4)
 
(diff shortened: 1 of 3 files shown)


bnkai:
Can you make the saving json file optional eg an argument like the teamskeet or mg PR ?
While the json file is useful for a lot of reasons some users may simply not want it.
Another cosmetic issue only. Can you sort the url list in the yml file ?
```yml
      - blacked.com/
      - blackedraw.com/
      - deeper.com/
      - tushy.com/
      - tushyraw.com/
      - vixen.com/
  ```
It wasnt enforced till recently but it makes debugging/checking scrapers easier.

File: scrapers/vixenNetwork.py [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+from urllib.parse import urlparse
+
+
+def log(*s):
+    print(*s, file=sys.stderr)
+
+
+def get_from_url(url_to_parse):
+    p = urlparse(url_to_parse)
+    return p.netloc, p.netloc[4:-4], p.path[1:]
```
bnkai, line 15: This is not actually that safe
In the yaml file we match vixen.com/ so in theory a user could have there `vixen.com/this-whole-time`
Using something like `ltdextract` to get the domain name would be probably better.

SpedNSFW: Yup! You're right! There are four ways the URL could be inserted and two of them would end up broken. `tldextract` is something that users would have to install so I'm not doing that, especially when there's a simple answer available with RegEx:

```py
m = re.match(r'(?:https?:\/\/)?(?:www\.)?((\w+)\.com)\/([a-z-]+)', url_to_parse)
return  m.groups()
```

scrapers/vixenNetwork.py
```diff
@@ -0,0 +1,98 @@
...
+import requests
+import sys
+
+
+def log(*s):
+    print(*s, file=sys.stderr)
+
+
+def get_from_url(url_to_parse):
+    m = re.match(r'(?:https?://)?(?:www\.)?((\w+)\.com)/([a-z-]+)', url_to_parse)
```
bnkai, line 13: `([a-z-]+)`
thats very exclusive
scenes with numbers fail eg `part-1` , `https://www.tushy.com/eva-part-1`

SpedNSFW: Ahh... That's what you call an oversight.

PR #366, opened 2021-02-08
Update VNAGirls.yml

dLCarbonX:
add siri


bnkai:
Siri was already there.
I did a sort for the URL list so the PR comes in handy.

PR #370, opened 2021-02-17
Create wetandpissy.yml

JaseNZC:

PR #371, opened 2021-02-17
Update SCRAPERS-LIST.md

JaseNZC:


SpedNSFW (changes requested):
This needs to be a part of #370, not in a new PR.

PR #375, opened 2021-02-19
Brazzers fix, now CDP is required

bnkai:
Not fully tested yet
scene changes only
- uses CDP now to get performers 
- the title and date from men was messed up so i used a script selector instead
- added a fallback for the studio when no subsite is present (mainly brazzers)
- added a fallback for the image from a script selector ( mainly for men)
- added scene URL

Also very important
For `brazzers` users must make sure the video URL has the title after the id otherwise we get redirected to the main page
If the user only has till the id part  for the url he needs to add an extra char after it eg for `https://www.brazzers.com/video/123456/` -> `https://www.brazzers.com/video/123456/a` to get to the scene page and then scrape ( thats were the scraped URL is needed)


Andiu1:
the tags dont be scraped from brazzers. any idea? what is CDP? Thanks :)

PR #376, opened 2021-02-20
add evolvedfightslez.yml

JaseNZC:


bnkai:
This should work  as is with `evolvedfights.com/updates`
Only the studio needs to change, maybe from `twitter:site`

JaseNZC:
I was going to include evovedfights but their layout is not quite the same and you get a bad scrape for the performers.

bnkai:
If its just the performers its just a matter of choosing a more `restrictive` selector eg
```//div[contains(@class,"model_update_block_info")]/span[@class="tour_update_models"]/a```

**Comment by JaseNZC** [comment]
Cheers for the help works for both now

File: scrapers/evolvedfightslez.yml [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+          - parseDate: January 2 2006
+     Tags:
+      Name: //span[@class="tour_update_tags"][1]//a
+     Performers:
+      Name: Name: ///div[contains(@class,"model_update_block_info")]/span[@class="tour_update_models"]/a
+     Details: //span[@class="latest_update_description"]/text()
+     Image: //meta[@name="twitter:image"]/@content
+     Studio:
+        Name:
+          fixed: Evolvedfightslez
```
bnkai, line 28: You forgot to adjust the studio... Something like this should work
```yaml
          selector: //meta[@name="twitter:site"]/@content
          postProcess:
            - map:
                "@evolvedLez": Evolved Fights Lez
                "@EvolvedFights": Evolved Fights
```

JaseNZC: Yup works a treat

PR #380, opened 2021-02-22
add peeonher.yml

JaseNZC:


scrapers/peeonher.yml
```diff
@@ -0,0 +1,28 @@
...
+        selector: //h1[@class='page_title']
+     Date:
+        selector: //strong[text()="Published"]/following-sibling::text()
+        postProcess:
+          - replace:
+            - regex: (\d+)[dhnrst]+
+              with: $1
+          - parseDate: 2 January 2006
+     Performers:
+      Name: //div[@class="pull-left"]//a/text()
```
bnkai, line 20: this doesnt seem to work always eg `https://www.peeonher.com/updates/pissing-foursome/`
```yaml
      Performers:
        Name: //a[@class="sub_sub_model"]
```
should be enough

scrapers/peeonher.yml
```diff
@@ -0,0 +1,28 @@
...
+     Details: //div[@class='con_left']/p[1]
+     Tags:
+        Name: //a[@class="sub_sub_tag"]//a/text()
```
bnkai, line 23: although it somehow seems to work is it a typo or something (the //a[]//a part) ?
```yaml
      Tags:
        Name: //a[@class="sub_sub_tag"]
```
should be ok

scrapers/peeonher.yml
```diff
@@ -0,0 +1,28 @@
...
+     Date:
+        selector: //strong[text()="Published"]/following-sibling::text()
+        postProcess:
+          - replace:
+            - regex: (\d+)[dhnrst]+
+              with: $1
+          - parseDate: 2 January 2006
+     Performers:
+      Name: //div[@class="pull-left"]//a/text()
+     Details: //div[@class='con_left']/p[1]
```
bnkai, line 21: /p`[1]`
Was there a scene that needed that? I tried some scenes but could only find single paragraphed ones.

JaseNZC: Hmmm yeah don't know why I had that in there it serves no purpose at all.

scrapers/peeonher.yml
JaseNZC: No idea how that was even in there to be honest.

scrapers/peeonher.yml
JaseNZC: Yup good spotting.

scrapers/peeonher.yml
bnkai: I wasnt sure for the `[1]` part ( whether to use  `//div[@class='con_left']/p`  and concat also.)
The `/p` is needed though because otherwise we get the **Story ** part with every description
It should be `//div[@class='con_left']/p`

PR #381, opened 2021-02-22
Fix found_scene_id extraction from scene URL

Void-Killer:
Fixed regex to correctly match scene ID from scene URL.

Current regex fails to match scene ID if scene name starts from number
E.g: https://regex101.com/r/tEuW9c/1/

Fixed regex should correctly match scene ID no matter the scene name
E.g: https://regex101.com/r/Pt7KxT/1


Belleyy:
Correct is : `/(\d+)/*` because you can maybe have a url without `/` at the end. 
E.g: https://regex101.com/r/arwGOe/1

PR #385, opened 2021-02-23
Adultime API scraper (Python script)

Belleyy:
### ~~Change the path to your sqlite database (Line 23).~~ Get your database from the config.yml now

Try to find your scene, work best with original file from this network. 
- Setting a URL can be inaccurate because the ID is different by site. If you get the wrong scene, try to remove the url (you can add it back when the scraping is done)

### DIFF
--- a/scrapers/AdultimeAPI.py
+++ b/scrapers/AdultimeAPI.py
@@ -25,7 +25,7 @@ def debug(q):
 
 def check_db(DB_PATH, scene_id):
     try:
-        sqliteConnection = sqlite3.connect(DB_PATH)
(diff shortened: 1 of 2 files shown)


scrapers/AdultimeAPI.py
```diff
@@ -0,0 +1,467 @@
...
+}
+
+
+def debug(q):
+    print(q, file=sys.stderr)
+
+
+def check_db(DB_PATH, scene_id):
+    try:
+        sqliteConnection = sqlite3.connect(DB_PATH)
```
bnkai, line 28: Should we open the DB as read only just to be safer?
```python
        sqliteConnection = sqlite3.connect(DB_PATH+"?mode=ro")
```

PR #386, opened 2021-02-23
Added scraper for my-slavegirl.com

nocrad349:


scrapers/my-slavegirl.yml
```diff
@@ -0,0 +1,26 @@
+name: "My Slavegirl"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - 'my-slavegirl.com/en/collections/'
```
bnkai, line 5: `'` are not needed here
just for cosmetic reasons and to maintain the same format as the  rest of the scrapers i would remove them
```yaml
      - my-slavegirl.com/en/collections/
```

scrapers/my-slavegirl.yml
```diff
@@ -0,0 +1,26 @@
...
+      Title:
+        selector: //meta[@property="og:title"]/@content
+      Details: 
+        selector: //meta[@name="description"]/@content
+      Performers:
+        Name: //span[@class="text-nowrap"]/i[@title="Models"]/following-sibling::span/a/text()
+      Tags: 
+        Name: 
+          selector: //meta[@name="keywords"]/@content
+          split: ','
```
bnkai, line 20: The keywords have a space after the comma, even though stash cleans that out its better to do that in the split
eg 
```yaml
        Name:
          selector: //meta[@name="keywords"]/@content
          split: ", "
```
another alternative would be to use
```yaml
        Name: //div[@class="tags"]/a
```

PR #390, opened 2021-02-24
add pantyjobs.yml

JaseNZC:


scrapers/pantyjobs.yml
```diff
@@ -0,0 +1,25 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - pantyjobs.com/scenes/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: '//div[@class=''col-md-12 aligncenter'']/h1'
```
bnkai, line 11: you are using 2 `'` instead of `"`
the starting/ending  `'`  are not needed
```suggestion
        selector: //div[@class="col-md-12 aligncenter"]/h1
```

scrapers/pantyjobs.yml
```diff
@@ -0,0 +1,25 @@
...
+      Performers:
+        Name: '//h2[1]'
+      Details:
+        selector: >-
+          //div[@class="post-entry"]/p[not (starts-with(.,"Members get full
+          access"))]
+        concat: |+
+
+
```
bnkai, line 20: You are using a combo of the concat function and yaml's block chomping indicator
While it seems to work it is less readable  and prone to accidental removal of the blank line after the concat (and thus failure) compared to a simple 
```suggestion
        concat: "\n\n"
```

PR #391, opened 2021-02-24
Switch to a JSON scraper for SARJ-LLC (metartnetwork)

bnkai:
<del>This needs a custom version `https://github.com/stashapp/stash/pull/1150` to work properly. (binaries here <https://github.com/stashapp/stash/actions/runs/595500176>)
I will update the list with a version req when thats available
If anyone wants to try with a `normal` version they would need to replace `{url}` with `{title}` in the scraper , paste the URL in the TITLE field in stash and do a scrape with</del>
Stash version >= v0.5.0-17 is required


gitgiggety:
Hi,

I made a quite extensive script / python based scraper for these sites which also supports scraping models/performers (by URL) and can scrape scenes & galleries by the combination of title and date (obviously when it's an exact match) and I might try adding real search support as well (so search based on just the title, or title and possible studio / models/performers based on their search API/endpoint).

So I don't know if there's any interest in me sharing that script (when I cleaned it up a bit).

bnkai:
@gitgiggety feel free to make a PR whenever you think its ready(or  clean enough :-)).
We could either replace the json scraper then or use the json for byURL queries and yours for byFragment or whatever we feel is better by that time.

PR #397, opened 2021-02-28
Tweak VNAGirls

bnkai:
Fixes tags (it seems that now no space is used after the `,` )
Tweaks description so that all of it is scraped.

Not tested on all sites yet.


bnkai:
TODO: fix studio, image for some of the sites

bnkai:
Closing, will open a new PR instead

PR #398, opened 2021-03-01
add wearehairy.yml

JaseNZC:


scrapers/wearehairy.yml
```diff
@@ -0,0 +1,25 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //h1[@itemprop="name"]//h1
+      Date:
+        selector: //span[@class="added"]//@datetime
+        postProcess:
+          - parseDate: 2006 Jan 02
```
bnkai, line 15: You forgot to edit the parseDate
Its the same format that we use so we could skip the postProcess altogether
Just to  be safe though
```suggestion
          - parseDate: 2006-01-02
```

scrapers/wearehairy.yml
```diff
@@ -0,0 +1,25 @@
...
+      Details: //div[@class="desc"]//p
```
bnkai, line 16: This (if you set the log to debug and check it out) matched a lot of extra stuff with the first being the description
```suggestion
      Details: //div[@class="line"][contains(h3,"Description")]/p
```

scrapers/wearehairy.yml
```diff
@@ -0,0 +1,25 @@
...
+      Tags:
+        Name: //div[@class="tagline"]//a
+      Image: //video[@id="trailer"]/@poster
+      Performers:
+        Name: //div/div[@class="title" and 1]/a[1]
```
bnkai, line 21: This only works for single performer scenes ( fail here `http://www.wearehairy.com/models/Lesbian/Bossaia_Golloia_and_Lika_Luna_have_naughty_fun/`) and is probably more complex than its needed
```suggestion
        Name: //div[@class="meet"]//a
```

PR #399, opened 2021-03-01
add sapphix.yml

JaseNZC:
Include 4 sites under the same network, Sapphix network.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -274,6 +274,7 @@ firstanalquest.com|Firstanalquest.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 firstbgg.com|TeenMegaWorld.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 firstclasspov.com|Spizoo.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fistertwister.com|Fistertwister.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
+fistflush.com|sapphix.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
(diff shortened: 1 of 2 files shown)


JaseNZC:
Remove stuff up.

JaseNZC:
Let me know if you want to map the names, thought it might be better as is.

bnkai:
Most of the selectors seem like that they could break from a small theme change.
I used some that seem safer
```yaml
name: sapphicnetwork
sceneByURL:
  - action: scrapeXPath
    url:
      - fistflush.com
      - givemepink.com
      - sapphicerotica.com
      - sapphix.com
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    scene:
      Title:
        selector: //div[@class="container"]//h2
      Date:
        selector: //div[@class="container"]//span[contains(.,"Added")]
        postProcess:
          - replace:
              - regex: "Added|,"
                with: ""
          - parseDate: January 2 2006
      Details: //p[@class="mg-md"]
      Tags:
        Name: //div[contains(h4,"Tags:")]//a
      Performers:
        Name: //div[contains(h4,"Featured model")]//a
      Studio:
        Name:
          selector: //a[text()="Join Now"]/@href
          postProcess:
            - replace:
                - regex: ^.*//(?:in\.)?(.+?)/.+$
                  with: $1
      Image: //video[@id="video"]/@poster | //div[@class="col-sm-12"]//img/@src
# Last Updated March 01, 2021
```
Also i added an `img` to the image failover to limit some results
I tested with all sites for a couple of scenes each and they seem to work ok.
@JaseNZC  Can you also try and update ?

JaseNZC:
Tested around 8 - 10 scenes on each site without error.

PR #400, opened 2021-03-01
ModelCentro Api (Python Script)

Belleyy:
Scraper for mylifeinmiami & fallinlovia & vinaskyxxx. 
Could probably have more URL.

There is no way to know the performer 😕 

I think the name (SAPI) is bad, any idea ? 😃 

Xpath scraper could be make but the date won't be scraped.


bnkai:
Both sites seem to be designed/made by modelcentro? Maybe use that?

PR #403, opened 2021-03-02
create pjgirls.yml

JaseNZC:


scrapers/pjgirls.yml
```diff
@@ -0,0 +1,37 @@
+name: pjgirls
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - pjgirls.com
```
bnkai, line 5: Make sure we use the videos and not the pictures
```suggestion
      - pjgirls.com/en/video
```

scrapers/pjgirls.yml
```diff
@@ -0,0 +1,37 @@
...
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class='container']/h1
+        postProcess:
+          - replace:
+              - regex: "VIDEO:"
+                with: ""
+     Date:
+        selector: //div[@class='info']/h3[1]
```
bnkai, line 17: From a quick look it doesnt appear that there is a video without a date
Just to be sure though
```suggestion
        selector: //div[@class='info']/h3[not(starts-with(.,"LENGTH"))]
```

PR #404, opened 2021-03-02
add eurosexsites.yml

JaseNZC:
Works with 6 sites , did not really know what to call it , Just change name if you want.


JaseNZC:
Changed name.

PR #407, opened 2021-03-04
add lethalpass.yml

JaseNZC:
This site contains a lot of scenes as well as exclusive sites.
If you scrape an exclusive site scene it will show up in the scrape with the studio
If you scrape a scene which is a scene from a dvd the studio will be blank with the scrape.

### DIFF
--- a/scrapers/lethalpass.yml
+++ b/scrapers/lethalpass.yml
@@ -23,7 +23,7 @@ xPathScrapers:
         Name:
           selector: //div[@class="pdSRC"]//a[starts-with(@href,"/dvds/")]
      Tags:
-      Name: //a[@class='cat' and 1]
+      Name: //a[@class='cat']
      Image:
         selector: //a[@class="noplayer"]//@src
      Studio:
--- a/scrapers/lethalpass.yml
+++ b/scrapers/lethalpass.yml
@@ -1,0 +1,29 @@
+name: lethalpass
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - lethalpass.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class="title"]/h1
+     Date:
+        selector: //div[@class="views"]/span
+        postProcess:
+          - replace:
+              - regex: "Added "
+                with: ""
+          - parseDate: 2 January 2006
+     Performers:
+      Name: //a[@class="model"]//a
+     Details: //div[@class="description"]/p
+     Tags:
+      Name: //a[@class='cat' and 1]
+     Image:
+        selector: //a[@class="noplayer"]//@src
+     Studio:
+        Name:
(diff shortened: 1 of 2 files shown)


scrapers/lethalpass.yml
```diff
@@ -0,0 +1,29 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - lethalpass.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class="title"]/h1
```
bnkai, line 11: from a quick test
title doesnt work for episodes from dvds `https://www.lethalpass.com/5956970` 
HD is returned most of the times.

scrapers/lethalpass.yml
```diff
@@ -0,0 +1,29 @@
...
+        postProcess:
+          - replace:
+              - regex: "Added "
+                with: ""
+          - parseDate: 2 January 2006
+     Performers:
+      Name: //a[@class="model"]//a
+     Details: //div[@class="description"]/p
+     Tags:
+      Name: //a[@class='cat' and 1]
```
bnkai, line 23: what does the `and 1` accomplish? isnt it always true?

bnkai:
a `movies` part  can be added for scenes from dvds
```yaml
      Movies:
        Name:
          selector: //div[@class="pdSRC"]//a[starts-with(@href,"/dvds/")]
```

scrapers/lethalpass.yml
JaseNZC: Bugger, was so focused on the exclusive site name I did not notice the Title changing.

scrapers/lethalpass.yml
JaseNZC: Yup 1 can come out it is indeed always true.

JaseNZC:
Forgot about the ability to be able to add movies.

JaseNZC:
Should have addressed all issues.

JaseNZC:
Fixed up tags. DOH

scrapers/lethalpass.yml
```diff
@@ -0,0 +1,32 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - lethalpass.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class="pdSRC"]//a[starts-with(@href,"/sites/")]
```
bnkai, line 11: The title now is kinda `useless`....
```suggestion
        selector: //div[@class="title"]/h1/text()|//div[@class="pdSRC"]//a[starts-with(@href,"/dvds/")]
```
The suggested one adds a `/text()` to the one you used before so no `hd` is printed/matched in case of dvds and the movie title is used instead... 
Does the suggeted one work well for you (and is it enough or should we try to get the ep number also) ?

JaseNZC:
Works well for me , at this stage without the id's I think is good enough.

PR #414, opened 2021-03-06
replace scraper for SARJ LLC with a script

gitgiggety:
Replace the JSON based scraper for the SARJ LLC websites with a more
elabore script.

The scripts supports the same features. Including remapping site ids to
their respective names, supporting the same sites, mapping the same
fields etc.. But also supports some additional features, which are
scraping performers by URL, and scraping scenes and galleries by their
fragments based on name and date. This thus allows to fill in the title
and date for a scene, applying the scraper, and all other information
will automatically be scraped. When scraping information the URL is also
automatically updated. So for example when scraping a scene released by
Viv Thomas but using the metartnetwork.com URL the URL will
automatically be replaced by the URL on vivthomas.com.

Using this script based scraper would also allow to add additional
features. Like a more elaborate search by using the search API and using
additional fields from the fragment. But this is more complex as for one
the fragment contains publishes ids (like performer "1", not including
the name), so that information should then first be fetched using the
GraphQL API. And the search API of the site seems to be rather picky in
the search terms.

### DIFF
--- a/scrapers/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC.py
@@ -1,0 +1,147 @@
+import json
+import os
+import sys
+import re
+import requests
+from urllib.parse import urlparse, urlunparse, urlencode
+
+def scrape_url(url, type):
+    parsed = urlparse(url)
+
+    path = parsed.path.split('/')
+    baseUrl = f"{parsed.scheme}://{parsed.netloc}"
+    if type == 'scene':
+        index = path.index('movie')
+        scraped = scrape_movie(baseUrl, path[index+1], path[index+2])
+    elif type == 'gallery':
+        index = path.index('gallery')
+        scraped = scrape_gallery(baseUrl, path[index+1], path[index+2])
+    elif type == 'performer':
+        index = path.index('model')
+        scraped = scrape_model(baseUrl, path[index+1])
+    else:
+        return None
+
+    return scraped
+
+def search(fragment, type):
+    if type == 'scene' or type == 'gallery':
+      name = re.sub(r'\W', '_', fragment['title']).upper()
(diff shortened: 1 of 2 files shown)


scrapers/SARJ-LLC.py
```diff
@@ -0,0 +1,141 @@
...
+      scraper = globals()['scrape_' + ('movie' if type == 'scene' else type)]
+      res = scraper('https://metartnetwork.com', date, name)
+      if res != None:
+          return res
+
+def fetch(baseUrl, type, arguments):
+    url =f"{baseUrl}/api/{type}?{urlencode(arguments)}"
+    response = requests.get(url)
+    if response.status_code >= 400:
+        return None
```
bnkai, line 41: Can you add a timeout for the `requests` get ?  Also a try for  exceptions should also be present eg network error. timeout ,... You can have a look at https://github.com/stashapp/CommunityScrapers/blob/1c65e2b793123713751cafe582745799a304b072/scrapers/vixenNetwork.py#L25 The referer and agent from there are not needed but don't hurt either.

scrapers/SARJ-LLC.py
```diff
@@ -0,0 +1,141 @@
...
+    if data == None:
+        return None
+
+    studio = get_studio(data['siteUUID'])
+    return map_media(data, studio, baseUrl)
+
+def map_model(baseUrl, model):
+    return {
+        'Name': model['name'],
+        'Gender': model['gender'].capitalize(),
```
bnkai, line 88: This should be `upper()` not capitalize
For the perforer gender enums all allcaps https://github.com/stashapp/stash/blob/ecac7a80137d830900bcd5fbe120c5c8ee648036/graphql/schema/types/performer.graphql#L1

PR #415, opened 2021-03-07
add bellapass.yml

JaseNZC:
Covers off 14 sites under the bellapass umbrella.

I have mapped some of the sites because they clashed with the name's but if you feel those mapped don't need to be just remove them.

### DIFF
--- a/scrapers/bellapass.yml
+++ b/scrapers/bellapass.yml
@@ -1,4 +1,4 @@
-name: bellapass.yml
+name: bellapass
 sceneByURL:
   - action: scrapeXPath
     url:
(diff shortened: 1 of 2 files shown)


JaseNZC:
Adjusted name.

JaseNZC:
I think that is better. Dunno what I was even thinking with the other way.

scrapers/bellapass.yml
```diff
@@ -0,0 +1,49 @@
...
+        Name: //li[@class="update_models"]//a
+      Tags:
+        Name: //div[@class="featuring clear"][2]//a
+      Image:
+        selector: //base/@href|//img[@class="update_thumb thumbs stdimage"]/@src0_1x
+        concat: "|"
+        postProcess:
+          - replace:
+              - regex: "[|]"
+                with: ""
```
bnkai, line 35: i havent checked all sites yet, from a quick look though some of the sites dont have the `https:` in their base eg bryci, bellanextdoor. While the ui from stash doesnt mind ( you can see the images) you'll get an error if you try to save because the backend expects a valid URL with the schema included.
An extra regex takes care of that
```suggestion
        postProcess:
          - replace:
              - regex: "[|]"
                with: ""
              - regex: "^//"
                with: "https://"
```

PR #417, opened 2021-03-07
add superbemodels.yml

JaseNZC:


scrapers/superbemodels.yml
```diff
@@ -0,0 +1,29 @@
...
+  - action: scrapeXPath
+    url:
+      - superbemodels.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="item-name-model-video item-tags-change"]//h1
+      Date:
+         selector: //span[@class="duration-box duration-box-change"]//span[contains(.,"Release date:")]/em
```
bnkai, line 12: although it seems to work we can simpliify and restrict  it  a bit
```suggestion
        selector: //span[@class="duration-box duration-box-change"][contains(span,"Release date:")]/em
```

scrapers/superbemodels.yml
```diff
@@ -0,0 +1,29 @@
...
+         postProcess:
+           - parseDate: January 2, 2006
+      Details: //div[@class="item description-video"]/p
+      Performers:
+        Name: //a[@class="models-name-box "]//a
```
bnkai, line 17: `//a` of `//a` doesnt feel right. It seems to work though
The
```suggestion
        Name: //a[@class="models-name-box "]
```
is enough

PR #427, opened 2021-03-11
add hushpass.yml

JaseNZC:


scrapers/hushpass.yml
```diff
@@ -0,0 +1,30 @@
...
+      Details: //div[@class="update-info-block"]//h3[contains(.,"Description:")]/following-sibling::text()
+      Tags:
+        Name: //ul[@class="tags"]//a
+      Performers:
+        Name: //div[@class="update-info-block models-list-thumbs"]//a
+      Studio:
+        Name:
+          fixed: HushPass
+      Image:
+        selector: //img[@class="update_thumb thumbs stdimage"][1]/@src0_1x
```
bnkai, line 25: The `[1]` soesnt work like you would think. As a result multiple images are returned if you check the log, of those the first happens to be the image we want ( <https://stackoverflow.com/questions/3674569/how-to-select-specified-node-within-xpath-node-sets-by-index-with-selenium> )
I would suggest the below instead
```suggestion
        selector: //div[@class="player-thumb"]//img[@class="update_thumb thumbs stdimage"]/@src0_1x
```

PR #430, opened 2021-03-12
add milfed.yml

JaseNZC:
Pretty funky classes but I went through about 20 scenes going right back and all worked just fine.


bnkai:
milfed is part of mindgeek. I think we should adjust that scraper instead to work with the site ( some selectors from that scraper will need to be updated as they dont work with this site)

JaseNZC:
Ahhh did not realise it was part of them.

PR #431, opened 2021-03-12
add tags to mylf.yml

JaseNZC:
Introduce tags to mylf.yml

### DIFF
--- a/scrapers/Mylf.yml
+++ b/scrapers/Mylf.yml
@@ -31,4 +31,4 @@ xPathScrapers:
               - regex: \sX\s.*
                 with: ""
 
-# Last Updated September 13, 2020
+# Last Updated March 12, 2021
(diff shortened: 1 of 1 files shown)


JaseNZC:
I always forget to change updated at bottom, DOH

scrapers/Mylf.yml
```diff
@@ -13,6 +13,13 @@ xPathScrapers:
         selector: //div[contains(@class,"sceneDate")]/text()
         postProcess:
             - parseDate: 01/02/2006
+      Tags:
+        Name: 
+          selector: //div[@class="col-12 tags-container"]//a
+          postProcess:
+            - replace: 
+              - regex: "[^a-zA-Z0-9]+"
+                with: $1
```
bnkai, line 22: The `$1 ` here doesnt serve a purpose. You havent used a matching group in the regex so that is equivalent to
`with:  ""` or `with:`. I would use the `with: ""` . Also if you can add a comment next to it eg `comment the  whole postProcess if you want to preserve the original tags` that would be great. ( i for example like to keep whitespaces or `/` with my tags)

PR #432, opened 2021-03-12
Added Hookup Hotshot

HereIam80:
Never used Github before. Let me know if I did anything wrong.

PR #433, opened 2021-03-13
add SeeHimFuck.yml 

yousefgunz:


bnkai:
@yousefgunz can we close this and continue in #432 ? The layout of that site is exactly the same so i combined the two scrapers (used parts from both) and posted there

yousefgunz:
@bnkai Sure, I got no problem with that!

PR #438, opened 2021-03-14
Adding Desperate Amateurs

killjoy4115:


scrapers/DesperateAmateurs.yml
```diff
@@ -0,0 +1,35 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - desperateamateurs.com/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="title_bar"]/text()
```
bnkai, line 11: the title selector needs to  be restricted a bit 
it matches the `individual clips` and `updates` also if you check the log
```suggestion
        selector: (//div[@class="title_bar"])[1]
```
should be ok

scrapers/DesperateAmateurs.yml
```diff
@@ -0,0 +1,35 @@
...
+      Tags:
+        Name: //a[@class="model_category_link"][contains(@href, 'category.php')]/text()
+      Image:
+        selector: //img[@class="thumbs"]/@src
+        postProcess:
+          - replace:
+            - regex: (.+image:\s+")(.+jpg)(.+)
+              with: $2
+            - regex: ^
+              with: "https://desperateamateurs.com/"
```
bnkai, line 24: the image has the same issue with above it matches all the thumbnails / 16 images
using this selector instead
```yaml
        selector: //img[@class="thumbs" and contains(src,"001.jpg")]/@src
```
only matches the first thumbnail
Another simpler alternative is using
```yaml
      Image: //meta[@name="twitter:image"]/@content
```
or 
```yaml
      Image: meta[@property="og:image"]/@content
```
Note that the meta use a different image each and usually not the same with the video player ( but neither does the 1rst thumb)

scrapers/DesperateAmateurs.yml
```diff
@@ -0,0 +1,35 @@
...
+      Studio:
+        Name:
+          fixed: Desperate Amateurs
+      Date:
+        selector: //td[@class="date"]/text()
```
bnkai, line 29: same issue with the date
the selector matches the dates from the relevant clips
```suggestion
        selector: //table[@width="100%"]//td[@class="date"]/text()
```
Only matches the correct one

scrapers/DesperateAmateurs.yml
killjoy4115: This makes sense. Thanks so much for checking this over.

PR #443, opened 2021-03-20
fix title in Puretaboo

jeremymeyers:
was pulling incorrect scene title data


bnkai:
It wasnt an address redesign just a case of the selector not working with specific scenes

PR #444, opened 2021-03-20
Add JeffsModels

killjoy4115:


scrapers/JeffsModels.yml
```diff
@@ -0,0 +1,30 @@
...
+        selector: //a[@class="hi iconPlay"]/img[@alt="Main Sample"]/@src
+        postProcess:
+          - replace:
+             - regex: ^
+               with: "https://jeffsmodels.com"
+      Studio:
+        Name:
+          fixed: "Jeff's Models"
+      Date:
+        selector: //div[@class="updateMeta"]/span[@class="updateMetaDate"]/span[@class="updateDate"]/text()
```
bnkai, line 26: For the date there seems to be only one span with that class
Is there a reason not to use the simpler `//span[@class="updateDate"]/text()` ?

killjoy4115: This makes sense. Thanks for finding that.

Issue #446, opened 2021-03-23
Scraper for vixenNetwork is not working anymore


davmuell:
they changed the url to the videos

bnkai:
Thanks for the report and the fix
Closing as the fix was merged

PR #447, opened 2021-03-23
fixed regex in get_from_url

davmuell:


davmuell:
fixes issue #446

scrapers/vixenNetwork.py
```diff
@@ -14,7 +14,7 @@ def log(*s):
 
 
 def get_from_url(url_to_parse):
-    m = re.match(r'(?:https?://)?(?:www\.)?((\w+)\.com)/([a-z0-9-]+)', url_to_parse)
+    m = re.match(r'(?:https?://)?(?:www\.)?((\w+)\.com)/videos/([a-z0-9-]+)', url_to_parse)
```
bnkai, line 17: while that would work for scenes that have the new urls it wont work for old ones
the below should work with old and updated urls
```suggestion
    m = re.match(r'(?:https?://)?(?:www\.)?((\w+)\.com)/(?:videos/)?([a-z0-9-]+)', url_to_parse)
```

bnkai:
Also line 68 should be changed from
```python
    response, err = make_request('https://%s/%s' % (site, slug), 'https://%s' % site)
```
to
```python
    response = make_request('https://%s/videos/%s' % (site, slug), 'https://%s' % site)
```
Even though the old urls seem to redirect to the new we should use the updated ones to be safe

PR #453, opened 2021-03-27
Update performer image scraper

peolic:
As reported on Discord.
~~Fixed to work with Windows paths (by converting them to POSIX paths before making the image URL)~~
Removed the need to set up and use the `custom_served_folders` config option.
Instead the script reads the image data and converts it to a [base64-encoded data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs).

I don't use this regularly, but I did ~~2~~ multiple tests with this and it seems to work as intended.
@ me if this needs further work.

### DIFF
--- a/scrapers/performer-image-dir.py
+++ b/scrapers/performer-image-dir.py
@@ -1,8 +1,9 @@
+import base64
 import json
+import mimetypes
 import os
 import sys
 from pathlib import Path
-from urllib.parse import quote, urljoin
 
 '''
 This script is here to allow you to search for performer images based on a directory.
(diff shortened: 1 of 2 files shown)


peolic:
Actually, I can get rid of the need for custom served folders and URL-based access altogether...
(Also, the current version doesn't work if you have username & password set)

bnkai:
Yeah when the scraper was made there was no support for getting the images as base64 (before the set image from url PR)

peolic:
Oh, didn't know that.
Well, this is easier to set up and use, as you only need to set the path to the cloned repository.

bnkai:
Yes for a couple of days (probably) , till the 0.6 is merged that change will require a dev version (after 11 March).
No biggie though and it indeed makes the setup a lot easier.

peolic:
> that change will require a dev version

I'm not sure that is true, as per this change:
https://github.com/stashapp/stash/pull/1193/files#diff-b3cb4284d8eddcdaeb2f50fcbb997e83e26e71e2d4499537953c6805e2078d42

(Base64 was already handled before that change?)

But, I'll try it with a v0.5.0 right now.

PR #454, opened 2021-03-27
Update LittleCapriceDreams.yml

JaseNZC:
Was not working was pulling in tags for performers etc.

My interpretation for a fix :-) 

Everything does work now.


bnkai:
From a quick test
```yaml
      Performers:
        Name: //div[@class="vid_infos"]/div[contains(div, "Models")]//a
      Tags:
        Name: //div[@class="vid_infos"]/div[contains(div, "Tags")]//a
```
seems to be enough

JaseNZC:
You win this round :-)

PR #462, opened 2021-04-01
Update NewSensationsNetworkSites.yml

JaseNZC:
Update to fix a few things, all sites should now work.

Some of the sites have tags some don't.

### DIFF
--- a/scrapers/NewSensationsNetworkSites.yml
+++ b/scrapers/NewSensationsNetworkSites.yml
@@ -4,7 +4,7 @@ sceneByURL:
     url:
       - thetabutales.com/tour_tt/
       - jizzbomb.com/tour_jb/
-      - freshouttahighschool.com/tour_fohs/
+      - freshoutofhighschool.com/tour_fohs/
       - newsensations.com/tour_rs/
       - fourfingerclub.com/tour_ffc/
       - ashlynnbrooke.com/tour_ab/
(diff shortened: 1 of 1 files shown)


bnkai:
the `thelesbianexperience` site matches images from the related scenes using the original selector. If the main image is not there we could end up with an image from the related scenes instead
I tried this `Image: //video[@id]/@poster` instead (video that has an id instead of any video) and it seems to work ok with all sites
Can you also update the url list with a sorted one ?
```yaml
      - ashlynnbrooke.com/tour_ab/
      - fourfingerclub.com/tour_ffc/
      - freshoutofhighschool.com/tour_fohs/
      - jizzbomb.com/tour_jb/
      - newsensations.com/tour_rs/
      - parodypass.com/tour_pp/
      - shanedieselsbanginbabes.com/tour_sdbb/
      - thelesbianexperience.com/tour_tle/
      - thetabutales.com/tour_tt/
      - unlimitedmilfs.com/tour_um/
   ```

JaseNZC:
> the `thelesbianexperience` site matches images from the related scenes using the original selector. If the main image is not there we could end up with an image from the related scenes instead
> I tried this `Image: //video[@id]/@poster` instead (video that has an id instead of any video) and it seems to work ok with all sites
> Can you also update the url list with a sorted one ?
> 
> ```yaml
>       - ashlynnbrooke.com/tour_ab/
>       - fourfingerclub.com/tour_ffc/
>       - freshoutofhighschool.com/tour_fohs/
>       - jizzbomb.com/tour_jb/
>       - newsensations.com/tour_rs/
>       - parodypass.com/tour_pp/
>       - shanedieselsbanginbabes.com/tour_sdbb/
>       - thelesbianexperience.com/tour_tle/
>       - thetabutales.com/tour_tt/
>       - unlimitedmilfs.com/tour_um/
> ```

Yup good idea for the image I did not notice that.

PR #463, opened 2021-04-01
Update NewSensationsMain.yml

JaseNZC:
Update to NewSensationsMain.

There was a lot of changes to the layout.

This is my go at it, laugh if you will but it does work. :-)


bnkai:
> This is my go at it, laugh if you will but it does work. :-)

It actually works pretty well no need to laugh :P
Only a minor change for the images
`span[@id='trailer_thumb']//img[not(@alt)]/@src`
The not alt part makes sure that we dont match the png icon 
Without it we get two images as matched the second being a png from the player

JaseNZC:
That should be all good to go now.

JaseNZC:
Typo : Doh

PR #465, opened 2021-04-04
Fix GloryHoleSwallow Details concat multiple paragraphs

SmartCap66:
Was previously scraping only the first paragraph of scene description, now all available text will be concatenated, providing a full scene description.


bnkai:
Thanks for the  fix. I had to try a lot of scenes to find one with a long enough description.

PR #466, opened 2021-04-07
add pinupgirls.yml

JaseNZC:


JaseNZC:
Fixed up scrapers list and cleaned up sites in scraper.

scrapers/pinupdollars.yml
```diff
@@ -0,0 +1,40 @@
...
+      Tags:
+        Name: //ul[@class="tags"]//a
+      Image:
+        selector: //script[contains(text(),'poster')]/text()
+        postProcess:
+          - replace:
+              - regex: .+(?:poster=")([^"]+).+
+                with: $1
+              - regex: ^
+                with:
```
bnkai, line 36: leftover?

bnkai:
Can you rename the file to `PinupDollars.yml` ?
Some users suggested we stick to one format either camelcase or lower case for the names so we have a  consistent naming scheme and a working sort for the file listing.
I was thinking of using the CamelCase one for all new scrapers and when editing/updating  the older ones.

scrapers/pinupdollars.yml
```diff
@@ -0,0 +1,40 @@
...
+          - parseDate: January 2, 2006
+      Performers:
+        Name: //div[@class="update-info-block models-list-thumbs"]//a
+      Details:
+        selector: //div[@class="update-info-block"]/h3[contains(.,"Description:")]/following-sibling::text()
+        concat: "\n\n"
+      Tags:
+        Name: //ul[@class="tags"]//a
+      Image:
+        selector: //script[contains(text(),'poster')]/text()
```
bnkai, line 30: The script selector doesn't work for all sites
I tried the below and it seems to work ok
```yaml
      Image:
        selector: //script[contains(text(),'poster')]/text()|//div[@class="player-thumb"]/img/@src0_1x
        postProcess:
          - replace:
              - regex: .+(?:poster=")([^"]+).+
                with: $1
              - regex: ^
                with:
```
I havent tried yet
```yaml
        Image: //div[@class="player-thumb"]/img/@src0_1x
```
to see if it is enough for all sites

JaseNZC:
Yup seems to do the trick

bnkai:
Can you also update the list with the new filename?

JaseNZC:
> Can you also update the list with the new filename?

Yup forgot about ye old list :-)

bnkai:
@JaseNZC  is it ok for merging or should it wait and test with only the non script image selector.
Also the last updated date is  a bit off ?

JaseNZC:
Thought I had made the change dunno what happened there, I did test it out the other day and it does indeed work fine on all sites.

PR #468, opened 2021-04-07
[validator] add support for `driver.headers`

peolic:
Adds support for https://github.com/stashapp/stash/pull/1273


bnkai:
ty @peolic 
I made a commit with your validator change in the relevant scraper PR so that it passes the check

PR #475, opened 2021-04-10
add BlowBangGirls.yml

JaseNZC:
This is part of a 7 site group under the CDM Networks umbrella.

Only 3 of the other sites have a similar layout but the Performer, tags and image are slightly different and I think it would be easier to have each one as a different scraper to avoid breakage.

### DIFF
--- a/scrapers/BlowBangGirls.yml
+++ b/scrapers/BlowBangGirls.yml
@@ -1,0 +1,30 @@
+name: BlowBangGirls
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - blowbanggirls.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class="videoDetails clear"]/h3
+     Date:
+        selector: //div[@class="videoInfo clear"]//span[contains(.,"Date Added:")]/following-sibling::text()
+        postProcess:
+          - parseDate: January 02, 2006
+     Performers:
+      Name: //span[@class="tour_update_models"]/a
+     Details: //div[@class="videoDetails clear"]/p
+     Tags:
+       Name: //span[@class="tour_update_tags"]/a
+     Image:
+        selector: //script[contains(text(),'/v3/content/')]/text()
+        postProcess:
+          - replace:
+              - regex: .+(?:poster=")([^"]*).+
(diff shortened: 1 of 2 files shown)


scrapers/BlowBangGirls.yml
```diff
@@ -0,0 +1,30 @@
...
+  sceneScraper:
+    scene:
+     Title:
+        selector: //div[@class="videoDetails clear"]/h3
+     Date:
+        selector: //div[@class="videoInfo clear"]//span[contains(.,"Date Added:")]/following-sibling::text()
+        postProcess:
+          - parseDate: January 02, 2006
+     Performers:
+      Name: //span[@class="tour_update_models"]/a
```
bnkai, line 17: was that copied from somewhere else?
 i needed to change that to
```yaml
        Name: //li[@class="update_models"]/a
```

scrapers/BlowBangGirls.yml
```diff
@@ -0,0 +1,30 @@
...
+     Details: //div[@class="videoDetails clear"]/p
+     Tags:
+       Name: //span[@class="tour_update_tags"]/a
```
bnkai, line 20: same here
```yaml
        Name: //ul/li[contains(text(),"Tags:")]/..//a
```

scrapers/BlowBangGirls.yml
```diff
@@ -0,0 +1,30 @@
...
+     Image:
+        selector: //script[contains(text(),'/v3/content/')]/text()
+        postProcess:
+          - replace:
+              - regex: .+(?:poster=")([^"]*).+
+                with: https://www.blowbanggirls.com/$1
```
bnkai, line 26: In some scenes `https://www.blowbanggirls.com/v3/trailers/Vee.html` , `https://www.blowbanggirls.com/v3/trailers/Blowbang-Bukkake-Completely-Covered-in-Cum.html` , `https://www.blowbanggirls.com/v3/trailers/I-Want-Candee.html` the script was missing
I used the below
```yaml
      Image:
        selector: //script[contains(text(),'/v3/content/')]/text()|(//div[@class="player_thumbs"]/a)[1]/img/@src0_1x
        postProcess:
          - replace:
              - regex: .+(?:poster=")([^"]*).+
                with: $1
              - regex: ^
                with: https://www.blowbanggirls.com/
```
to get a fallback image

PR #480, opened 2021-04-13
add VegasDreamsLLC.yml

JaseNZC:
This is 6 sites under the VegasDreamsLLC branch.

It also replaces ScrewMeToo.yml which is part of the six and is also a lot more simple.


bnkai:
I've tried 3-4 scenes for every site and all seems ok.
The only exception seems a single scene from here <https://trikepatrol.com/content/bi-asian-swinger-looking-for-thrills/>
The div from the description selector there doesnt use any `p`.
I couldnt find anymore failing scenes from that site so idk if its worth it to change something for this (we would need to recheck).

JaseNZC:
> I've tried 3-4 scenes for every site and all seems ok.
> The only exception seems a single scene from here https://trikepatrol.com/content/bi-asian-swinger-looking-for-thrills/
> The div from the description selector there doesnt use any `p`.
> I couldnt find anymore failing scenes from that site so idk if its worth it to change something for this (we would need to recheck).

Personally I would not worry about it, It would be a very few select scenes that it would effect, if people wanted to they could just copy and paste description in, Unless somebody made a request I am happy with how it is if you are.

PR #501, opened 2021-04-25
add MadeInCanada.yml

JaseNZC:


bnkai:
An altenate selector for the date
```yaml
        selector: //div[@class="released-date"]/span[@class="grey"]
```
The site seems to have very few scenes (compared to the 3500 it claims ) in the tour (free access)

JaseNZC:
Yeah a lot of site's offer limited access to scenes even though they have many pages before sending you to the pay wall. Bastards

PR #504, opened 2021-04-26
add HollyRandall.yml

JaseNZC:
Not much  in the way of descriptions, some here and there.


bnkai:
Just out of curiocity was there a date that didnt have an `Added:` ? I would have gone for `(^Added:\s+)` instead of `(^\S\s+)` just to make sure i don't get a false match.

JaseNZC:
> Just out of curiocity was there a date that didnt have an `Added:` ? I would have gone for `(^Added:\s+)` instead of `(^\S+\s+)` just to make sure i don't get a false match.

I did not see any that were missing it, but make's perfect sense to avoid a false match. Has been changed.

PR #505, opened 2021-04-26
[validator] support `distinct` post-processing

peolic:
Adds support for `distinct` post-processing option, to filter only unique values.
Ready for when https://github.com/stashapp/stash/pull/1338 is merged.


bnkai:
Marking as blocked till we sort this out.

peolic:
> Distinct is now applied by default during the pp phase.

PR #506, opened 2021-04-26
add GenuineSin.yml

JaseNZC:


bnkai:
They use the same template as femdomempire/dickdrainers with a little bit of changes. It's better to have a separate scraper like this though

The sister site `mixedx.com` works with the scraper as is if we use the below selector for studio
```yaml
      Studio:
        Name: //div[@class="logo"]/a/img/@alt
```

I would also use 
```yaml
      Tags:
        Name: //ul[li[contains(text(),"Tags:")]]//a
```
as i try to avoid the following siblings if i can


Can you add `      - mixedx.com` to the `url:` and update the SCRAPERS-LIST?

JaseNZC:
Done :-)

PR #521, opened 2021-05-05
add VegasDreamworks.yml

JaseNZC:


scrapers/VegasDreamworks.yml
```diff
@@ -0,0 +1,28 @@
+name: VegasDreamworks
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - toticos.com
+      - paradisegfs.com
```
bnkai, line 6: can you fix the order?

scrapers/VegasDreamworks.yml
```diff
@@ -0,0 +1,28 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //h1[@class="post-title"]
+      Performers:
+        Name: //div[@class="single-model-head"]/a | //div[@class="update_bb_info"]/a
```
bnkai, line 14: the parent is the same in both sites a simple
```suggestion
        Name: //div[@class="update_bb_info"]//a
```
should be ok

bnkai:
The details when the scenes have a lot of paragraphs(Read More...) seem a little bit off (only the appearance) but its not worth the effort to do something more imo.

JaseNZC:
Yeah the details arn't the best but as you said not really worth doing any more to it.

PR #522, opened 2021-05-06
add HandDomination.yml

JaseNZC:


scrapers/HandDomination.yml
```diff
@@ -0,0 +1,34 @@
...
+      Title: //h3[@class="videoTitle"]
+      Details:
+        selector: //li[@class="videoDescription"]/p
+      Tags:
+       Name: //li[@class="videoCategories"]/p/a
+      Date:
+        selector: //li[@class="update_posted"]/p
+        postProcess:
+          - replace:
+              - regex: "date:"
```
bnkai, line 19: I would replace  the spaces after the `date:` just to be safe and not rely on stash's trimming.
```suggestion
              - regex: 'date:\s+'
```

PR #525, opened 2021-05-07
add Hentaied.yml

JaseNZC:
Man what a crazy site with both content and layout.

Did the best I could and I think I have covered off everything , Have done a lot of testing and get results on all scrapes.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -385,6 +385,7 @@ hdsex18.com|Wtfpass.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 hegre.com|Hegre.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:heavy_check_mark:|-|-
 helloladyboy.com|YegasDreamsLLC.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 hentai2read.com|hentai2read.yml|:x:|:heavy_check_mark:|:x:|:x:|-|Hentai
+hentaied.com|Hentaied.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 hergape.com|AnalizedNetwork.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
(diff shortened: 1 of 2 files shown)


bnkai:
The layout was too crazy!!!
The tags selector you used contains stuff that shouldnt  be included IMHO. A plain `//ul[@class="post-categories"]//a` should be enough, again IMHO.
Now for the tricky part....
Scenes like `https://hentaied.com/a-drow-elf/` , `https://hentaied.com/cumflation/`, `https://hentaied.com/mr-alien-always-knocks-twice/` dont work that well with your details
I adjusted the tags and used the below
```yaml
name: Hentaied
sceneByURL:
  - action: scrapeXPath
    url:
      - hentaied.com
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $desc: "*[local-name()='p' or local-name()='h3' or local-name()='ol' or local-name()='ul']"
    scene:
      Title:
        selector: //div[@class="post-content"]/h1
      Date:
        selector: //div/h2
        postProcess:
          - replace:
              - regex: 'PUBLISHED:\s*'
                with: ""
          - parseDate: 2006-01-02
      Performers:
        Name: //div[@class="tagsmodels"]/a
      Tags:
        Name: //ul[@class="post-categories"]//a
      Details:
        selector: //div[@class="cont" or @class="column"]/$desc
        concat: "\n\n"
      Image:
        selector: //meta[@property="og:image"]/@content
      Studio:
        Name:
          fixed: Hentaied
# Last Updated May 9, 2021
```
which seemed to include all the info from my limited testing
If needed we can add to the common `desc`  more `names` ( p,h3,ol and ul where what i noticed from the scenes i tested)
The form of the `ul` and `ol` ( video including and facts) is broken but thats the best i could do without messing the text order.
If you can test the above with your samples too that would be great.

JaseNZC:
Yup one of the most crazy sites I have come across.

About as good as its going to get I think. Can be revisited if need be in the future.

PR #526, opened 2021-05-07
add NastyStepFamily.yml

JaseNZC:


bnkai:
Thats a part of vivid. Using the vivid scraper as is works fine. You only need to add the `url` and get the `Studio` part from something like the `canonical` url or the `alt` of the logo image.

PR #527, opened 2021-05-07
add BoppingBabes.yml

JaseNZC:
Could not find a way or figure out if it was possible to add a date.


bnkai:
> Could not find a way or figure out if it was possible to add a date.

from a quick look
```yaml
      Date:
        selector: //span[@class="day"]//text()
        concat: "/"
        postProcess:
          - replace:
              - regex: '^(\w{3})/(\d\d)$'
              # 21 is the current year. Needs to be updated next year
                with: $1/21/$2
          - parseDate: Jan/06/02
```
thats the best i could come up with.
On 2022 for example the above needs to be adjusted with `22` instead of `21`

JaseNZC:
Yup that works.

PR #530, opened 2021-05-07
add Cospuri.yml

JaseNZC:
No title that I am able to find.


bnkai:
Something like this
```yaml
      Title: //div[@class="sample-model"]  
```
could work for the title. They even have the ` & ` if more than one performers are present

JaseNZC:
Yup sounds good.

PR #531, opened 2021-05-08
Update Mylf.yml (Tags)

Belleyy:
They added tags to the scenes.

_Issue:_
There is a comma as tags (tried using replace regex, it will give you a blank tag)

-------------
Using regex replace for the image to get a HD image (1280x720) instead of mid quality (765x431)


bnkai:
For the tags its either ignore the comma or do a concat split
```yaml
        Name:
          selector: //div[contains(@class,"tags-container")]/a/text()
          postProcess:
          concat: " "
          split: " , "
```
Both are fine by me

PR #532, opened 2021-05-08
Switch to a python scraper for IAFD

bnkai:
Requires extra modules `cloudscrape` and `html`
Sometimes the image isnt loaded on time in the page (thats my best guess) and needs a rescrape to get it.
(images from iafd arent that good anyway so it shouldnt be a problem)


scrapers/IAFD.py
```diff
@@ -0,0 +1,609 @@
...
+        p.birthdate = re.sub(r'(\S+\s+\d+,\s+\d+).*', r'\1', p.birthdate)
+        try:
+            p.birthdate = datetime.datetime.strptime(p.birthdate, iafd_date).strftime(stash_date)
+        except:
+            pass
+
+    performer_deathdate = tree.xpath('(//p[@class="bioheading"][text()="Date of Death"]/following-sibling::p)[1]//text()')
+    p.death_date = p.set_value(performer_deathdate)
+    if p.death_date:
+        p.death_date = re.sub(r'(\S+\s+\d+,\s+\d+).*', r'\1', p.deathdate)
```
7dJx1qP, line 459: p.deathdate should be p.death_date

PR #540, opened 2021-05-13
add Gasm.yml

JaseNZC:
This has Scene's from a lot of major sites.
I have put the studio from where the scene has come from if you prefer the studio to be GASM just change it :-)

### DIFF
--- a/scrapers/GASM.yml
+++ b/scrapers/GASM.yml
@@ -1,0 +1,22 @@
+name: GASM
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - gasm.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="gqText"]/span[@class="gqTitle" and 2]
+      Performers:
+        Name: //div[@class="gqOwner"]/a
+      Tags:
+        Name: //div[@class="content tag"]/a
+      Details:
+        selector: //span[@class="gqDescription"]
+      Image: //meta[@property="og:image"]/@content
+      Studio:
+        Name:
+          selector: //span[@class="gqOwner"]//a
+# Last Updated May 13, 2021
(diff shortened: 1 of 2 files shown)


scrapers/GASM.yml
```diff
@@ -0,0 +1,22 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - gasm.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="gqText"]/span[@class="gqTitle" and 2]
```
bnkai, line 11: What is the `and 2` supposed to do? Afaik it will always be true?

scrapers/GASM.yml
```diff
@@ -0,0 +1,22 @@
...
+      Performers:
+        Name: //div[@class="gqOwner"]/a
+      Tags:
+        Name: //div[@class="content tag"]/a
+      Details:
+        selector: //span[@class="gqDescription"]
```
bnkai, line 17: In scenes where the scene is part of a movie we get a Part of .... at the start of the description, eg `https://www.gasm.com/post/details/20936` `https://www.gasm.com/post/details/21234`
Using
```yaml
      Details:
        selector: //span[@class="gqDescription"]/text()
        concat: "\n\n"
        postProcess:
          - replace:
              - regex: '^Part\sof\s*\n+'
                with:
      Movies:
        Name: //span[@class="gqDescription"]/a[@class="gqTitle"]
```
looks better imo.

scrapers/GASM.yml
JaseNZC: Nothing it does nothing and I don't even know why its there, I was working on something else so might have transposed it maybe lol.

scrapers/GASM.yml
JaseNZC: Good point on the movie have added it in :-)

PR #543, opened 2021-05-14
fix encoding, set lang

astringa1234:


bnkai:
Can you add the `lang` part as an extra argument to the script?
By default it should be `en` unless the arg is set to `fr` for example
The user should then be able to just uncomment/edit the .yaml file to his liking

astringa1234:
As the website is in french, and the English version is just a (bad) translation, doesn't it make sense to keep the default to french?
I'll add the argument to the script

bnkai:
> As the website is in french, and the English version is just a (bad) translation, doesn't it make sense to keep the default to french?
> I'll add the argument to the script

The original scraper was already in english as i would guess most of the users would expect.
If an english version exists it should be prefered imo, the same goes with the JAV stuff for example. Given that the english translations are not that good should we include the japanese by default?

astringa1234:
How can I pass an argument?

Tried adding this, but seems it does not work

```
    script:
              - python3
              - JacquieEtMichelTV.py 
              - lang=fr
```

bnkai:
I woud do something like this
```yaml
name: JacquieEtMicaelTV
sceneByFragment:
    action: script
    script:
      - python
      - JacquieEtMichelTV.py
sceneByURL:
  - url:
     - jacquieetmicheltv.net
    action: script
    script:
      - python
      - JacquieEtMichelTV.py
      #- fr # uncomment if you want to use the french description
# Last Updated May 20, 2021
```
and
```python
import base64
import datetime
import json
import string
import sys
from urllib.parse import urlparse
# extra modules below need to be installed
import cloudscraper
from lxml import html

lang = 'en'

def log(*s):
    print(*s, file=sys.stderr)
    ret_null = {}
    print(json.dumps(ret_null))
    sys.exit(1)


def strip_end(text, suffix):
    if suffix and text.endswith(suffix):
        return text[:-len(suffix)]
    return text

if len(sys.argv) > 1:
        if sys.argv[1] == 'fr':
            lang = 'fr'


frag = json.loads(sys.stdin.read())
if not frag['url']:
    log('No URL entered.')

url = frag["url"]
scraper = cloudscraper.create_scraper()
try:
    cookies = {'lang': lang}
    scraped = scraper.get(url, cookies=cookies)
except:
    log("scrape error")

if scraped.status_code >= 400:
    log('HTTP Error: %s' % scraped.status_code)

tree = html.fromstring(scraped.text)

title = tree.xpath("//h1/text()")[0]
date = tree.xpath("//span[@class='publication']/text()")[0]
details = tree.xpath("//div[@class='video-description']/p")[0]
tags = tree.xpath("//span[@class='categories']/a/strong/text()")
imgurl = tree.xpath("//video[@id='video-player']/@poster")[0]
img = scraper.get(imgurl).content
b64img = base64.b64encode(img)
datauri = "data:image/jpeg;base64,"

if lang == 'fr':
    date = datetime.datetime.strptime(date, "%d/%m/%Y").strftime("%Y-%m-%d")
else:
    # en
    date = datetime.datetime.strptime(date, "%m/%d/%Y").strftime("%Y-%m-%d")

ret = {
    'title': title,
    'tags': [{
        'name': strip_end(x, ', ')
    } for x in tags],
    'date': date,
    'details': details.text_content(),
    'image': datauri + b64img.decode('utf-8'),
    'studio': {
        'name': 'Jacquie Et Michel TV'
    },
}

print(json.dumps(ret))
#Last Updated May 20, 2021
```

That way any user that wants the `fr` text can just uncomment a line
Note: If you want the fr texts make sure to use a url that doesnt have the `/en/` part.

astringa1234:
Ok so it seems that I need to set the parameter under sceneByFragment, and not sceneByURL

I'm using the button "scrape with.." next to "save" and "delete". Is this the correct way?

bnkai:
I updated the above yml file, you can uncomment in both sections.

astringa1234:
Should be good now

PR #545, opened 2021-05-15
Sarj llc improvements

gitgiggety:
Some improvements on the API based SARJ-LLC scraper. These are:
 * Fixing "random" 404 errors when scraping a model
 * Converting more information from a model to a performer
 * Supporting searching models by name (kind of a hack, but seems to work fine)

Note that the combination of searching a performer by name / fragment doesn't properly work in combination with Tags. This as the tags aren't properly handled by stash which means they aren't converted to (possible) existing tags and thus are listed as new tags, but adding them by clicking the tag / + results in an error for the tag already existing. This is fixed in stashapp/stash#1387

### DIFF
--- a/scrapers/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC.py
@@ -147,16 +147,34 @@ def scrape_gallery(baseUrl, date, name):
     return map_media(data, studio, baseUrl)
 
 def map_model(baseUrl, model):
+    tags = list(map(lambda t: {'Name'}, model['tags']))
+
+    def add_tag(key, format):
+        nonlocal tags
+        if key in model and model[key] != "":
+            tags.append({
+                'Name': format.format(model[key])
+            })
+
+    add_tag('hair', '{} hair')
+    add_tag('pubicHair', '{} pussy')
+    add_tag('eyes', '{} eyes')
+    add_tag('breasts', '{} breasts')
+
     return {
         'Name': model['name'],
         'Gender': model['gender'].upper(),
         'URL': f"{baseUrl}{model['path']}",
         'Ethnicity': model['ethnicity'],
         'Country': model['country']['name'],
-        'EyeColor': model['eyes'],
+        'hair_color': model['hair'].capitalize(),
+        'eye_color': model['eyes'].capitalize(),
         'Height': str(model['height']),
+        'Weight': str(model['weight']),
         'Measurements': model['size'],
-        'Image': f"{baseUrl}{model['headshotImagePath']}"
+        'Details': model['biography'],
+        'Image': f"https://cdn.metartnetwork.com/{model['siteUUID']}{model['headshotImagePath']}",
+        'Tags': tags
     }
 
 studios = {
--- a/scrapers/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC.py
@@ -24,7 +24,7 @@ def scrape_url(url, type):
 
     return scraped
 
-def search(fragment, type):
+def query(fragment, type):
     if type == 'scene' or type == 'gallery':
       name = re.sub(r'\W', '_', fragment['title']).upper()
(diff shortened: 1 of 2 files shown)


scrapers/SARJ-LLC.py
```diff
@@ -89,16 +147,35 @@ def scrape_gallery(baseUrl, date, name):
...
         'Name': model['name'],
         'Gender': model['gender'].upper(),
         'URL': f"{baseUrl}{model['path']}",
         'Ethnicity': model['ethnicity'],
         'Country': model['country']['name'],
-        'EyeColor': model['eyes'],
+        'HairColor': model['hair'].capitalize(),
+        'hair_color': model['hair'].capitalize(),
+        'EyeColor': model['eyes'].capitalize(),
+        'eye_color': model['eyes'].capitalize(),
```
gitgiggety, line 173: `HairColor` is documented in the scraper configuration, but doesn't work. Found some other scraper which used `hair_color` and does work. Same applies to `EyeColor` / `eye_color` as well.

scrapers/SARJ-LLC.py
```diff
@@ -50,7 +94,21 @@ def fetch(baseUrl, type, arguments):
...
+          lambda p:
+            re.sub(
+              '[_-]',
+              ' ',
+              re.sub('\w\S*', lambda m: m.group(0).lower().capitalize(), p),
+            ),
+          name.split('-')
+        )
+      )
+    )
```
gitgiggety, line 110: This logic is converted from the JavaScript of the website. In short it capitalizes every part of the name (splitted on the `-`) and joins the parts with a space. So URL `/model/foo-bar` would become `/api/model?name=Foo%20Bar`. Without this (and adding `order` and `direction`) sometimes calls would result in a model not found error. With these 2 changes it seems to work fine every time.

scrapers/SARJ-LLC.py
bnkai: yeah you can have a look at
https://github.com/stashapp/stash/blob/develop/pkg/models/model_scraped_item.go
we need the `json` value to be the same

scrapers/SARJ-LLC.py
bnkai: i noticed the search was acting weird good catch

scrapers/SARJ-LLC.py
bnkai, line 173: As mentioned above the `hair_color` and `eye_color` are what we expect in the  json output.
EyeColor and HaiColor are not needed/used.
It seems a 
```python
        'Details': model['biography'],
```
can also be added to the performer data.

gitgiggety: Was a bit unsure about this, and I myself don't feel the need fot it, but added it now.

gitgiggety: IMO it's a bit strange that these are documented as being in CamelCase like all the others, but being defined in snake_case. But seems like all are actually defined in lowercase to I guess the JSON unmarshaller is case insensitive then.

It just felt a bit odd to use hair_color etc while all others are capitalized. So was also unsure if the desired behavior / name actually shouldn't have been in CamelCase here as well. But seeing the code it seems like more are actually in snake_case.

bnkai: Yeah when used in stash, they have the json definition available so the json marshalller/unmarsaller can use that. I don't know if there is something equivalent in python. Imo for the python part they should all be snake_case but since i have seen a lot of users use the CamelCase i didnt bother to mention that. In some cases though like here we can''t avoid snake_case definitions.

bnkai: Yeah i dont think i'll actually use sars for a performer scraper but its there if anyone wants to. The same goes with some of the fields i guess. The user can decide for him/herself what he wants to add when the scraper dialog appears.

PR #546, opened 2021-05-16
add Pegasproductions.yml

JaseNZC:
Their site layout sucks big time.

### DIFF
--- a/scrapers/PegasProductions.yml
+++ b/scrapers/PegasProductions.yml
@@ -7,7 +7,7 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-# To access other scenes go to Scene and select model which will show all their scenes
+      # To access other scenes go to Scene and select model which will show all their scenes
       Title:
         selector: //span[@itemprop="name"]
       Date:
--- a/scrapers/PegasProductions.yml
+++ b/scrapers/PegasProductions.yml
@@ -15,15 +15,15 @@ xPathScrapers:
         postProcess:
           - parseDate: 02/01/2006
       Performers:
-        Name: //div[@class="span10"]/p[1]/a
+        Name: //div[@class="span10"]/p[contains(b,"Starring")]/a
       Tags:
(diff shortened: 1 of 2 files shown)


JaseNZC:
Adjusted where it was supposed to be in the list.

scrapers/PegasProductions.yml
```diff
@@ -0,0 +1,46 @@
...
+    scene:
+# To access other scenes go to Scene and select model which will show all their scenes
+      Title:
+        selector: //span[@itemprop="name"]
+      Date:
+        selector: //div[@id="date-duree"]/div[1]/p[1]
+        postProcess:
+          - parseDate: 02/01/2006
+      Performers:
+        Name: //div[@class="span10"]/p[1]/a
```
bnkai, line 18: I think we can use 
```yaml
        Name: //div[@class="span10"]/p[contains(b,"Starring")]/a
```
instead

scrapers/PegasProductions.yml
```diff
@@ -0,0 +1,46 @@
...
+      Tags:
+        Name:
+          selector: //div[@class="span10"]/p[2]
```
bnkai, line 21:
```
          selector: //div[@class="span10"]/p[contains(b,"Tags")]
```

scrapers/PegasProductions.yml
```diff
@@ -0,0 +1,46 @@
...
+          postProcess:
+            - replace:
+                - regex: "Tags:"
+                  with:
+          split: ","
```
bnkai, line 26: either use 
```yaml
          split: ", "
```
(note the extra space after the comma) or do a replace regex for `,\s` with `,` instead
The tags after the first get a leading space since `split` is the last action in the pp chain and the thus they wont be `cleaned` after the split.

JaseNZC: Should be all good to go now :-) , Thanks for the comments lets me learn all the time.

PR #552, opened 2021-05-17
Update DDFNetwork.yml

JaseNZC:
`Date: //meta[@itemprop="uploadDate"]/@content`

This is no longer any good as half the sites have changed this to a different date layout.

The fix gets it from a different location now which is the same across all sites.

### DIFF
--- a/scrapers/DDFNetwork.yml
+++ b/scrapers/DDFNetwork.yml
@@ -2,8 +2,8 @@ name: DDFNetwork
 sceneByURL:
   - action: scrapeXPath
     url:
-      - ddfbusty.com/videos/
       - 1by-day.com/videos/
+      - ddfbusty.com/videos/
       - eurogirlsongirls.com/videos/
       - euroteenerotica.com/videos/
       - handsonhardcore.com/videos/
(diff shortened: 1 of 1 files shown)


scrapers/DDFNetwork.yml
```diff
@@ -15,12 +15,15 @@ xPathScrapers:
   sceneScraper:
     scene:
       Title: //meta[@itemprop="name"]/@content
-      Date: //meta[@itemprop="uploadDate"]/@content
+      Date:
+        selector: //div[@class='title-border d-inline-flex pr-3 flex-sm-fill justify-content-sm-center pr-sm-0']/p[@class='m-0 text-center']
```
peolic, line 19: if I may suggest this instead
```suggestion
        selector: //div[@id="video-specs"]//p[text()]
```

peolic: oh, wait, it fails on some sites

bnkai:
Normally i would avoid this kind of selectors, but it works for all sites and i wouldn't want to search for a `cleaner` selector that works for all sites :-)

PR #556, opened 2021-05-19
add TACAmateurs.yml

JaseNZC:
Covers a lot of amateurs.

### DIFF
--- a/scrapers/TACAmateurs.yml
+++ b/scrapers/TACAmateurs.yml
@@ -15,7 +15,9 @@ xPathScrapers:
           - parseDate: 02-January-2006
      Performers:
       Name: //div[@id="facepic_wrapper"]/h2
-     Details: //div[@id="update_wrapper"]/p[1]
+     Details: 
+       selector: //div[@id="update_wrapper"]/p[not(@class) and not(@align)]
+       concat: "\n\n"
      Tags:
       Name: //p[@class="tags"]/a
      Image:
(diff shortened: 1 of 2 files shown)


scrapers/TACAmateurs.yml
```diff
@@ -0,0 +1,26 @@
...
+    scene:
+     Title:
+        selector: //div[@id="update_wrapper"]/h1
+     Date:
+        selector: //div[@class="date_wrapper"]
+        postProcess:
+          - parseDate: 02-January-2006
+     Performers:
+      Name: //div[@id="facepic_wrapper"]/h2
+     Details: //div[@id="update_wrapper"]/p[1]
```
bnkai, line 18: depending on the performer there are scenes with more paragraphs
eg `https://www.tacamateurs.com/sites/Hot-Milf/my-new-bustier-pt2/24490/`
I did a limited testing with 
```yaml
     Details: 
       selector: //div[@id="update_wrapper"]/p[not(@class) and not(@align)]
       concat: "\n\n"
```
and it seemed ok. It might need a little more testing to make sure no extra p are added.

JaseNZC: Yeah that works ok, Gonna get one past ya one day that does not need editing hahahahahahaha

bnkai: You already had a few that didnt need any editing :-)
Usualy if the site is structured ok there is nothing to notice. I blame the site devs for not keeping it simple and tidy :P

JaseNZC: Yeah there is normally at least one  thing that is  a pain in the arse to try and scrape.

PR #558, opened 2021-05-20
add MVG.yml

JaseNZC:
The issue with this one is that some of the site's they have screwed the date up. On the scene's page it show the correct date but when you go into the actual scene they have the date as 1969-12-31.

If you want just cut out the date altogether.

### DIFF
--- a/scrapers/MVG.yml
+++ b/scrapers/MVG.yml
@@ -1,0 +1,42 @@
+name: MVG
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - italianshotclub.com
+      - lesbiantribe.com
+      - pornlandvideos.com
+      - sologirlsmania.com
+      - vangoren.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class='centerwrap container']/h1
+      Date:
+        selector: //div[@class="videodetails datetime"]/div[@class="date"]
+        postProcess:
+          - replace:
+              - regex: ^(\d{1,2} [a-zA-Z]* \d{4}).+
+                with: $1
+          - parseDate: 2 January 2006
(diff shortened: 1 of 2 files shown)


scrapers/MVG.yml
```diff
@@ -0,0 +1,42 @@
...
+    scene:
+      Title:
+        selector: //div[@class='centerwrap container']/h1
+      Date:
+        selector: //div[@class="videodetails datetime"]/div[@class="date"]
+        postProcess:
+          - replace:
+              - regex: ^(\d{1,2} [a-zA-Z]* \d{4}).+
+                with: $1
+          - parseDate: 2 January 2006
```
bnkai, line 22:
```yaml
      Date: # the date from the scene page is not correct most of the time, we use the one from the search  instead
        selector: //head/title
        postProcess:
          - replace:
              - regex: (Lesbiantribe|Pornlandvideos|Sologirlsmania)\.com$
                with: "- $1"
              - regex: ^(.*)\s*-\s*([^-]+)$
                with: 'https://$2.com/search.php?x=0&y=0&query="$1"'
              - regex: '\s'
                with: "+"
          - subScraper:
              selector: //div[@class="modeldetails clear"][1]//div[@class="video_date col-xl-5 col-lg-5 col-md-5 col-sm-5 col-xs-5 col-5 text-right"]
          - parseDate: 2 January 2006
 ```
Tested the above for a few scenes and it looked ok
The only problem would be if they dont use `unique` titles or the title is too short -> more than 1 search results
It needs a little more testing to make sure that doesnt happen, i think the probabilty of that is pretty low though and certainly a lot lower than having a 1969 date :-)

JaseNZC: Did some testing and looks good.
What an awesome way to get the date, would not have thought of or been able to do that. 
Respect.

PR #559, opened 2021-05-20
add Femanic.yml

JaseNZC:


scrapers/Femanic.yml
```diff
@@ -0,0 +1,27 @@
...
+      Title:
+        selector: //div[@class="dis_main_heading"]
+      Performers:
+        Name: //div[@class="main-modals"]/p//a
+      Tags:
+        Name: //div[@class="content tag"]/a
+      Details:
+        selector: //div[@class="disc-text"]/p
+      Image:
+        selector: //video[@style="max-width:100%;max-height:100%;"]/@poster
```
bnkai, line 19: Some scenes eg `https://femanic.com/video/Femanic-Cathy-Elvira-Kristyna` dont have the video part.
I tried the bewlow for  a couple 
```yaml
        selector: //video[@style="max-width:100%;max-height:100%;"]/@poster|//img[@class="video-container-img"]/@src
```
and it seems to work

bnkai:
I couldn't find a scene that has tags ?

scrapers/Femanic.yml
JaseNZC: Yup all sorted, Yeah there are no tags I just forgot to take it out. doh.

PR #562, opened 2021-05-21
add SissyPov.yml

JaseNZC:
add Sissy.yml

### DIFF
--- a/scrapers/SissyPov.yml
+++ b/scrapers/SissyPov.yml
@@ -12,14 +12,11 @@ xPathScrapers:
       Performers:
         Name: //li[@class="update_models"]/a
       Date:
-         selector: //div[@class="videoInfo clear"]/p[1]
+         selector: //p[contains(span,"Date Added:")]/text()
          postProcess:
-           - replace:
-               - regex: 'Date Added:\s*'
-                 with: ""
            - parseDate: January 2, 2006
       Tags:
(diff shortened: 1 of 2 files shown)


scrapers/SissyPov.yml
```diff
@@ -0,0 +1,34 @@
...
+        selector: //div[@class="videoDetails clear"]/h3
+      Performers:
+        Name: //li[@class="update_models"]/a
+      Date:
+         selector: //div[@class="videoInfo clear"]/p[1]
+         postProcess:
+           - replace:
+               - regex: 'Date Added:\s*'
+                 with: ""
+           - parseDate: January 2, 2006
```
bnkai, line 20: an altenative without relying on positioning
```yaml
      Date:
         selector: //p[contains(span,"Date Added:")]/text()
         postProcess:
           - parseDate: January 2, 2006
```

scrapers/SissyPov.yml
```diff
@@ -0,0 +1,34 @@
...
+      Tags:
+        Name: //div[@class="featuring clear"]//li/a
```
bnkai, line 22: the above selector includes performers
```yaml
        Name: //ul[contains(li,"Tags:")]//a
```
contains only the tags
Or you could filter yours so that only the `a` with `/categories/` in the `@href` are matched

JaseNZC:
Sorted, Did not notice they have been included in the tags.

PR #563, opened 2021-05-21
add PissyNetwork.yml

JaseNZC:


scrapers/PissyNetwork.yml
```diff
@@ -0,0 +1,27 @@
...
+      Title:
+        selector: //h2[@class="video__headline"]
+      Performers:
+        Name: //strong[contains(.,"Models:")]/following-sibling::a
+      Date:
+         selector: //strong[contains(.,"Published:")]/following-sibling::text()
+         postProcess:
+           - parseDate:  02 Jan 2006
+      Tags:
+        Name: //strong[contains(.,"Tags:")]/following-sibling::a
```
bnkai, line 19: the selector above includes the performers (the `a` for performers and tags are siblings )
```yaml
        Name: //strong[contains(.,"Tags:")]/following-sibling::a[contains(@href,"/movies/")]
```
selects only the tags

the Performers work ok because they are listed after the tags so the following doesnt have any tags present
we could use 
```yaml
      Performers:
        Name: //strong[contains(.,"Models:")]/following-sibling::a[contains(@href,"/models/")]
```
for the performers as an extra caution i guess

JaseNZC: Yup that all works , I gotta remember to take better notice of the tags that they are not including other info.

PR #565, opened 2021-05-21
add LoneStarAngel.yml

JaseNZC:


scrapers/LoneStarAngel.yml
```diff
@@ -0,0 +1,28 @@
...
+      - lonestarangel.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //span[@class="update_title"]
+      Performers:
+        Name:
+          fixed: Angel
```
bnkai, line 14: I would add the URL as a extra ?
```yaml
        Name:
          fixed: Angel
        URL:
          fixed: https://www.lonestarangel.com/tour/models/lonestarangel.html
```

JaseNZC: :-)

PR #566, opened 2021-05-21
add FFstockings.yml

JaseNZC:


scrapers/FFstockings.yml
```diff
@@ -0,0 +1,24 @@
...
+    scene:
+      Title:
+        selector: //div[@class="trailerContent"]/h2
+      Performers:
+        Name: //li/a/span
+      Tags:
+        Name:
+          selector: //div[@class="tags"]/a
+      Details:
+        selector: //div[@class="trailerContent"]/p
```
bnkai, line 18: it seems that in scenes with free trailers (eg `http://ffstockings.com/tour2/trailers/Cum-On-My-Face.html`) the trailer info gets in the description
keeping only the plain  text part of the p seems to work (with very limited testing)
```yaml
        selector: //div[@class="trailerContent"]/p/text()
```

bnkai:
it seems that 
```yaml
      - stockingvideos.com
```
can be added to the urls
with
```yaml
      Studio:
        Name: //div[@class="logo"]/a/@title
```
as the studio name

JaseNZC:
Did not see the other site, Works ok with 7 tested scenes.

PR #567, opened 2021-05-21
add LingerieTales.yml

JaseNZC:


scrapers/LingerieTales.yml
```diff
@@ -0,0 +1,29 @@
...
+        selector: //h1[@class="title"]
+      Date:
+        selector: //time
+        postProcess:
+          - parseDate: January 2, 2006
+      Performers:
+        Name: //ul[@class='filters_buttons']/li[2]/a
+      Tags:
+        Name:
+          selector: //ul[@class="filters_buttons"]/li/a
```
bnkai, line 20: the tags include the performers
i think we should filter the href from the a 
```yaml
      Tags:
        Name: //ul[@class="filters_buttons"]/li/a[contains(@href,"/videotag/")]
```
and maybe use the same for performers?
```yaml
      Performers:
        Name: //ul[@class="filters_buttons"]/li/a[contains(@href,"/model-item/")]
```

scrapers/LingerieTales.yml
```diff
@@ -0,0 +1,29 @@
...
+      Details:
+        selector: //div[@class="the_content_wrapper"]/p
```
bnkai, line 22: found a scene where the main desc is in a `pre` -> `https://www.lingerietales.com/video-item/leah-rising-early/`
```yaml
        selector: //div[@class="the_content_wrapper"]/p|//pre[@id="code"]
```
takes care of that

JaseNZC: Right behind ya hahahahaha, Bloody tags have screwed me over in this batch of studios DOH

PR #570, opened 2021-05-22
add SwingerBlog.yml

JaseNZC:


scrapers/SwingerBlog.yml
```diff
@@ -0,0 +1,27 @@
...
+      Date:
+        selector: //span[@class="update_date"]
+        postProcess:
+          - parseDate: 01/02/2006
+      Performers:
+        Name: //span[@class="tour_update_models"]/a
+      Tags:
+        Name: //span[@class="tour_update_tags"]/a
+      Details:
+        selector: //div[@class="video-info mt-3"]/p
```
bnkai, line 21: this selector doesnt seem to work ?
```yaml
        selector: //span[@class="latest_update_description"]
```
is the `safest` i could think of as there are descriptions in strong, in plain spans, ....
In some cases the above includes the tags section. Feel free to tweak it if needed.

JaseNZC:
I have no idea why I had that in there its nothing to do with that scraper, To many open tabs maybe.

PR #573, opened 2021-05-22
data18: support day-less dates

### LINKED ISSUE #572: data18 scraper data parseDate issue
https://github.com/stashapp/CommunityScrapers/blob/5ad18be5ee1ed4060a2a325eec7b428a5d3c6f57/scrapers/data18.yml#L22

On data18 there are cases where the release date is formatted like January 2 2006 but other formatted like January 2006.  Is there a way to account for 2 different parseDate formats?

peolic:
Fix #572

Example content [**NSFW**]: `http://www.data18.com/content/394108`

### DIFF
--- a/scrapers/data18.yml
+++ b/scrapers/data18.yml
@@ -12,7 +12,7 @@ xPathScrapers:
     scene:
       Title: //h1/text()
       Date:
-        selector: //span[contains(text(),'Release date:')]
+        selector: //span[contains(text(),"Release date:") and not(contains(text(),"No date release yet"))]
         postProcess:
           - replace:
               - regex: "^Release date:|,"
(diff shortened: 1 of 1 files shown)


bnkai:
I would only use data18 if i cant find the scene anywhere else so defaulting the day part is ok by me.
Will anyone have an issue with it, maybe add a comment to the regex explaining 1 is used when no day is available?

While testing the above i found this [NSFW] `http://www.data18.com/content/1195602`, a scene with no date field
I dont know if the text will be the same for other scenes without dates. 
I tried this for that  case
```yaml
        selector: //span[contains(text(),'Release date:')][not contains(text(),"no date release")]
```

PR #574, opened 2021-05-23
update Jhdv.yml

JaseNZC:
Added new sites and cleaned things up.


JaseNZC:
Let me know if you want me to go back the the naming convention as it was.

PR #579, opened 2021-05-30
Add data18.empirestores.co xPath scene, movie scraper

bnkai:
Another request from #73 

- not fully tested
- it seems that performers are not listed in the scene page


CuddleBear92:
oh! that was a fast pull, just tested editing the AdultEmpireCash file and just added a new line to movies and scenes and it works. atleast on the movie side fully. 
Did not test this Pull though!
Its own file shouldn't be needed.

bnkai:
The scenes had a few differences eg tags, no performers, movie name
The movies could be the same i didnt check them as i didnt want to test the scene changes for all sites.
If you can test the scenes part we can use your PR instead.

CuddleBear92:
struggled to find a scene that had a movie i could easily match.
But it seems to be working fine. 
chrome_2021-05-30_23-31-57

PR #581, opened 2021-05-30
Add XConfessions xPath scene, movie scraper

bnkai:
Another from #73


quirkyjoe:
Thank you very much for putting this together!

I've given it a spin, and it seems to work quite well, though I think I've found a small bug.
I think the performers selector is such that it only grabs the first performer.
And example of this behavior can be seen at this page:
https://xconfessions.com/film/take-me-home

PR #583, opened 2021-05-31
Update AdultimeAPI.py

Belleyy:
Using Stash to get the database

### DIFF
--- a/scrapers/AdultimeAPI.py
+++ b/scrapers/AdultimeAPI.py
@@ -12,12 +12,25 @@ import requests
 USERFOLDER_PATH = str(pathlib.Path(__file__).parent.parent.absolute())
 DIR_JSON = os.path.join(USERFOLDER_PATH, "scraperJSON","Adultime")
 
+SERVER_IP = "http://localhost:9999"
+SERVER_URL = SERVER_IP + "/graphql"
+# STASH API (Settings > Configuration > Authentication)
+APIKEYS = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1aWQiOiJmYXN0NjUiLCJpYXQiOjE2MTc4ODU4NDQsInN1YiI6IkFQSUtleSJ9.wAcjHE_ncn0jTE-vK5kcBEWKSOEDKlJf5JNR7aUE04E"
+
 HEADERS = {
     "User-Agent": 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0',
     "Origin": "https://members.adulttime.com",
     "Referer": "https://members.adulttime.com/"
 }
 
+STASH_HEADERS = {
+    "Accept-Encoding": "gzip, deflate, br",
+    "Content-Type": "application/json",
+    "Accept": "application/json",
+    "Connection": "keep-alive",
+    "DNT": "1",
+    "ApiKey": APIKEYS
+}
 
 def debug(q):
     print(q, file=sys.stderr)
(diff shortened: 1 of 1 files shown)


scrapers/AdultimeAPI.py
```diff
@@ -370,12 +382,38 @@ def create_index(api_json):
...
+        sys.exit("HTTP Error 401, Unauthorised.")
+    else:
+        raise ConnectionError("GraphQL query failed:{} - {}. Query: {}. Variables: {}".format(
+            response.status_code, response.content, query, variables))
+
+def getdbPath():
+    query = "query Configuration {  configuration {...ConfigData}}fragment ConfigData on ConfigResult {  general {...ConfigGeneralData}}fragment ConfigGeneralData on ConfigGeneralResult { databasePath }"
+    result = callGraphQL(query)
+    return result["configuration"]["general"]["databasePath"]
+
```
bnkai, line 407: if the user doesnt configure properly the SERVER_URL requests throws an exception

```python
def callGraphQL(query, variables=None):
    json = {'query': query}
    if variables is not None:
        json['variables'] = variables
    try:
        response = requests.post(SERVER_URL, json=json, headers=STASH_HEADERS)
        if response.status_code == 200:
            result = response.json()
            if result.get("error"):
                for error in result["error"]["errors"]:
                    raise Exception("GraphQL error: {}".format(error))
            if result.get("data"):
                return result.get("data")
        elif response.status_code == 401:
            sys.exit("Unauthorised! Make sure the API key was added to the scraper.")
        else:
            raise ConnectionError("GraphQL query failed:{} - {}. Query: {}. Variables: {}".format(
                response.status_code, response.content, query, variables))
    except Exception as err:
        debug(err)
        exit(1)

def getdbPath():
    query = "query Configuration {  configuration {...ConfigData}}fragment ConfigData on ConfigResult {  general {...ConfigGeneralData}}fragment ConfigGeneralData on ConfigGeneralResult { databasePath }"
    try:
        result = callGraphQL(query)
    except Exception as err:
        debug(err)
        exit(1)
    return result["configuration"]["general"]["databasePath"]
```
adding a try seems safer

scrapers/AdultimeAPI.py
```diff
@@ -370,12 +382,38 @@ def create_index(api_json):
...
 
 fragment = json.loads(sys.stdin.read())
 
 # Get your database
 DB_PATH = None
 config_path = os.path.join(USERFOLDER_PATH, "config.yml")
```
bnkai, line 413: shouldnt this be removed since we get the path from getdbPath below?

scrapers/AdultimeAPI.py
```diff
@@ -370,12 +382,38 @@ def create_index(api_json):
...
+def getdbPath():
+    query = "query Configuration {  configuration {...ConfigData}}fragment ConfigData on ConfigResult {  general {...ConfigGeneralData}}fragment ConfigGeneralData on ConfigGeneralResult { databasePath }"
+    result = callGraphQL(query)
+    return result["configuration"]["general"]["databasePath"]
+
 
 fragment = json.loads(sys.stdin.read())
 
 # Get your database
 DB_PATH = None
```
bnkai, line 412: DB_PATH if fetched from getdbPatgh below, shouldnt this be removed? (not needed)

scrapers/AdultimeAPI.py
Belleyy: Don't think it's useful `try` for getdbPath() because the error will be catch by callGraphQL, if a error happend in callGraphQL the script will exit.

scrapers/AdultimeAPI.py
Belleyy: I guess it don't harm anyone with this, if the user is too lazy to put the ApiKeys, this method could still work

PR #584, opened 2021-06-01
add ClaudiaMarie.yml

JaseNZC:


JaseNZC:
I have left the .com in the Studio name, If you don't want it that way just remove it.

bnkai:
> I have left the .com in the Studio name, If you don't want it that way just remove it.

I see no .com in the fixed? Merging as is.

PR #588, opened 2021-06-05
Update javdb.yml

Belleyy:
Replace the JSON by a Xpath one (But the JSON still worked).

Use the mirror site javdb9 because javdb have a better protection (cloudflare 😡).

#123

### DIFF
--- a/scrapers/javdb.yml
+++ b/scrapers/javdb.yml
@@ -1,7 +1,7 @@
 name: JavDB
 sceneByFragment:
-  action: scrapeJson
-  queryURL: https://javdb.com/videos/search_autocomplete.json?q={filename}
+  action: scrapeXPath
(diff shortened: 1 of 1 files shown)


bnkai:
> Replace the JSON by a Xpath one (But the JSON still worked).
The json didnt seem to work last time i tried (with a docker cdp instance though). Will recheck later on.
Do any of the javdb.com urls scrape ok ? ( should the javdb urls be removed so that only the javdb9 remain?)

Belleyy:
> The json didnt seem to work last time i tried (with a docker cdp instance though). Will recheck later on.

Don't work on javdb.com, they changed the search system.

> Do any of the javdb.com urls scrape ok ? ( should the javdb urls be removed so that only the javdb9 remain?)

I don't think, you get cloudflare everytime. I will comment javdb url because it could maybe work later.

PR #589, opened 2021-06-06
Velvetveronica

nocrad349:
Xpath scraper for velvetveronica.com

Date scraping is a bit iffy. For most videos the year and month of upload can be retrieved from image filename, but not the day. This PR therefore uses 1 for the day.
There are a limited number of old videos on the site that do have full date information in the image filename, but I don't see how to retrieve both forms in an xpath scraper.

### DIFF
--- a/scrapers/velvetveronica.yml
+++ b/scrapers/velvetveronica.yml
@@ -32,6 +32,6 @@ xPathScrapers:
       Studio:
         Name: 
           selector: //meta[@property="og:site_name"]/@content
-      Image: //meta[@property="og:image"]/@content
+      Image: //meta[@name="twitter:image"]/@content
 
 # Last Updated June 06, 2021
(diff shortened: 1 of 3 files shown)


scrapers/velvetveronica.yml
```diff
@@ -0,0 +1,37 @@
...
+      Performers: 
+        Name:
+          fixed: Veronica Dyer
+      Tags:
+        Name:
+          selector: //span[@class="tagged_as"]/a/text()
+      Studio:
+        Name: 
+          selector: //meta[@property="og:site_name"]/@content
+      Image: //meta[@property="og:image"]/@content
```
JaseNZC, line 35:
```suggestion
      Image: //meta[@name="twitter:image"]/@content
```
This returns a cleaner result for the image, og:image returns 6 or 7 different results.

PR #596, opened 2021-06-10
add MomXXX.yml

JaseNZC:


bnkai:
@JaseNZC the layout is the same as `Mypervmom` , you only need to add the url there 

as a sidenote (and not needed now)  your date regex should match/replace using the digit  (\d(st|nd|rd|th)) otherwise August gets trimmed

JaseNZC:
> @JaseNZC the layout is the same as `Mypervmom` , you only need to add the url there
> 
> as a sidenote (and not needed now) your date regex should match/replace using the digit (\d(st|nd|rd|th)) otherwise August gets trimmed

Did not realise it was the same. 
Good point on the Regex I did not think about August.

PR #598, opened 2021-06-13
add POVR.yml

JaseNZC:
Request from #73 (end of thread)

Covers all 18 studios on the site.

I have not removed WankzVR or MIlfVR as they are sites on their own.

Date is a bit Janky but does work, Change it if you want a better way to pull it :-)

### DIFF
--- a/scrapers/POVR.yml
+++ b/scrapers/POVR.yml
@@ -13,7 +13,7 @@ xPathScrapers:
         selector: //p[contains(@class, 'player__date')]
         postProcess:
           - replace:
-              - regex: .*(\d+\s[a-zA-Z]+,\s\d+)$
+              - regex: .*\s(\d+\s[a-zA-Z]+,\s\d+)$
                 with: $1
           - parseDate: 2 January, 2006
       Performers:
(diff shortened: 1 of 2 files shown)


scrapers/POVR.yml
```diff
@@ -0,0 +1,30 @@
...
+    scene:
+      Title:
+        selector: //span[@class="video__quality video__quality--vr mr-3"]/following-sibling::text()
+      Date:
+        selector: //p[contains(@class, 'player__date')]
+        postProcess:
+          - replace:
+              - regex: ^.{1,11}(.*)$
+                with: $1
+          - parseDate: 02 January, 2006
```
bnkai, line 18: for the date imo it is better to use a regex bassed on content than positioning
the date in the site does not include 0 padding btw
```yaml
      Date:
        selector: //p[contains(@class, 'player__date')]
        postProcess:
          - replace:
              - regex: .*(\d+\s[a-zA-Z]+,\s\d+)$
                with: $1
          - parseDate: 2 January, 2006
```

JaseNZC: Done, Told ya I had a Janky date there lol.

bnkai: @JaseNZC i added a last second  \s to the regex before the (
(your merge was faster than my edit :-) )

JaseNZC: Bam, Lightning fast.

PR #599, opened 2021-06-15
LetsDoeIt - Remove extra text in title

Muldec:
An extra test is appended after the title of every video. This removes it.

### DIFF
--- a/scrapers/LetsDoeIt.yml
+++ b/scrapers/LetsDoeIt.yml
@@ -22,6 +22,8 @@ xPathScrapers:
           - replace:
               - regex: (.+)(?:\s\|.+)
                 with: $1
+              - regex: (Exclusive\W+(\w+\W+)?Porn\W+Video)
+                with: $1
       Date:
         selector: //meta[@itemprop="uploadDate"]/@content
         postProcess:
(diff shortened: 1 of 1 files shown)


scrapers/LetsDoeIt.yml
```diff
@@ -20,7 +20,7 @@ xPathScrapers:
         selector: //title/text()
         postProcess:
           - replace:
-              - regex: (.+)(?:\s\|.+)
+              - regex: (.+)(?:Exclusive Porn Video)(?:\s\|.+)
                 with: $1
```
bnkai, line 24: In cases where the `Exclusive Porn Video` is not present or is a bit different eg dirtycosplay the above breaks the match
I think adding a second regex removing the extra part is safer
```yaml
        postProcess:
          - replace:
              - regex: (.+)(?:\s\|.+)
                with: $1
              - regex: (Exclusive\s(\S+\s)?Porn\sVideo)
                with: ""
```

Muldec: You're right !

Wouldn't `(Exclusive\W+Porn\W+Video)` be better ?

bnkai: If you mean `(Exclusive\W+(\w+\W+)?Porn\W+Video)` that also works
without the extra match group like you have it, it fails in the dirtycosplay case

Muldec: Right !
File updated and Squashed.

PR #600, opened 2021-06-15
Add IWantClips to the scrapers.

Muldec:
Create a new scraper for the IWantClips store

### DIFF
--- a/scrapers/IWantClips.yml
+++ b/scrapers/IWantClips.yml
@@ -19,6 +19,11 @@ xPathScrapers:
       Tags:
         Name:
           selector: //div[@class="col-xs-12 hashtags fix"]/span/em | //div[@class="col-xs-12 category fix"]/span
+          concat: ","
+          postProcess:
+            - replace:
+                - regex: ',\s+'
+                  with: ","
           split: ","
       Image:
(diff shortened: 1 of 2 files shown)


Muldec:
Nice, I've merged both into a new commit.

scrapers/IWantClips.yml
```diff
@@ -0,0 +1,31 @@
...
+        postProcess:
+          - replace:
+            - regex: Published\s(.+)
+              with: $1
+          - parseDate: Jan 2, 2006
+      Details: //div[@class="col-xs-12 description fix"]/span
+      Tags:
+        Name:
+          selector: //div[@class="col-xs-12 hashtags fix"]/span/em | //div[@class="col-xs-12 category fix"]/span
+          split: ","
```
bnkai, line 22: For the split to work properly you first need to concat the results. Otherwise you'll only get the split from the first matches, in this case the `em` tags only
```suggestion
          selector: //div[@class="col-xs-12 hashtags fix"]/span/em | //div[@class="col-xs-12 category fix"]/span
          concat: ","
          postProcess:
            - replace:
                - regex: ',\s+'
                  with: ","
          split: ","
```

PR #605, opened 2021-06-22
PornDB (URL): Remove CDP

peolic:
Update: Thanks to the PornDB team CDP is no longer needed.

~~It _sometimes_ panics on the first scrape with `Could not find node with given id (-32000)`,
and sometimes does not panic, just fails to scrape anything.
But in any case, it seems that consistently the first scrape does not work.
The panic issue might be a race condition in the way clicks are coded in, @bnkai, this might be more up your alley?
`https://github.com/chromedp/chromedp/issues/762#issuecomment-788836126`~~

~~\!\[image](https://user-images.githubusercontent.com/66393006/122960418-0d7b5200-d38c-11eb-894e-0cdd38da9235.png)~~

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -606,7 +606,7 @@ men.com|Brazzers.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|Gay
 menpov.com|AMAMultimedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|Gay
 messyxxx.com|ChickPass.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 metadataapi.net (JSON API)|ThePornDB.yml|:heavy_check_mark:|:x:|:x:|:heavy_check_mark:|-|-
-metadataapi.net (URL)|PornDB.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
+metadataapi.net (URL)|PornDB.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 metart.com|SARJ-LLC.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|python|-
 metartnetwork.com|SARJ-LLC.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|python|-
 metartx.com|SARJ-LLC.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|python|-
--- a/scrapers/PornDB.yml
+++ b/scrapers/PornDB.yml
@@ -47,10 +47,13 @@ xPathScrapers:
         #      with: "1000x1000"
 driver:
   useCDP: true
-  sleep: 5
+  clicks:
+    - # Click through Age Verification
+      xpath: //form[@action="https://metadataapi.net/age-verification"]//button[@type="submit"]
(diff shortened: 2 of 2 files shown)


DirtyRacer1337:
This is would be not required soon, just need to wait our server update

bnkai:
From a quick look it seems that if a session cookie is available the check is bypassed. That would require though the users to look for the session cookie in the browser, while the above `click` method works ok (at least with a remote cdp instance). I would reduce the sleep to `3` though.
When using the cdp with a remote docker instance the click is used only the first time you scrape since the cookies are kept while when using the chrome binary directly i suspect it would be used every time.


For the panics i couldn't replicate the issue as it worked fine every single time i tried it (again remote cdp instance) but from what you pointed out i got to <https://github.com/chromedp/chromedp/issues/551> 
and the below diff
```diff
diff --git a/pkg/scraper/url.go b/pkg/scraper/url.go
index baa35b07..73204311 100644
--- a/pkg/scraper/url.go
+++ b/pkg/scraper/url.go
@@ -15,7 +15,6 @@ import (
        "time"

        "github.com/chromedp/cdproto/cdp"
-       "github.com/chromedp/cdproto/dom"
        "github.com/chromedp/cdproto/network"
        "github.com/chromedp/chromedp"
        jsoniter "github.com/json-iterator/go"
@@ -175,14 +174,7 @@ func urlFromCDP(url string, driverOptions scraperDriverOptions, globalConfig Glo
                chromedp.Navigate(url),
                chromedp.Sleep(sleepDuration),
                setCDPClicks(driverOptions),
-               chromedp.ActionFunc(func(ctx context.Context) error {
-                       node, err := dom.GetDocument().Do(ctx)
-                       if err != nil {
-                               return err
-                       }
-                       res, err = dom.GetOuterHTML().WithNodeID(node.NodeID).Do(ctx)
-                       return err
-               }),
+               chromedp.OuterHTML("body", &res, chromedp.ByQuery),
                printCDPCookies(driverOptions, "Cookies set"),
        )

```
@peolic  can you try the above and see if it fixes your `panics` ?

peolic:
> This is would be not required soon, just need to wait our server update
> 
> Update: server got update, this PR not required anymore

@DirtyRacer1337 
Thank you very much for this, it works without CDP now.
PR updated.

> can you try the above and see if it fixes your panics ?

@bnkai 
I like the idea behind it, however I think (didn't dive in) that if this is added you should get the `outerHTML` of `html` rather than `body`, since some scrapers use `meta` tags that are in the `head` section.

Regarding testing,
I only saw this now, so I don't have a test case anymore 🙂
The only other scraper that uses clicks is `Hustler.yml` but:
1. I'm unable to find the element that needs to be clicked (neither manually nor while actually scraping).
2. I think what was... special in this case was the fact that the click was submitting a form and causing a redirect from `/age-verificaiton` to `/` (home).

peolic:
@bnkai 
Mark this down as confirmed.
Tried it with the Bang scraper - `"body"` = got no title, no description and no image, `"html"` - everything works.
> you should get the `outerHTML` of `html` rather than `body`, since some scrapers use `meta` tags that are in the `head` section.

bnkai:
@peolic my main concern was the `panics` so i didnt look that much into it. I'll test all the CDP scrapers over the weekend with the diff and the `html` part and make a PR if nothing is broken.

PR #606, opened 2021-06-23
create BamVisions.yml

JaseNZC:
As per requested from [https://github.com/stashapp/CommunityScrapers/issues/73#issuecomment-865998699](url)

Not much meta to be able to grab but there is some.


bnkai:
Some scenes have a description `https://tour.bamvisions.com/trailers/Zoey-Takes-A-Pounding.html`,`https://tour.bamvisions.com/trailers/Doing-Something-I-Like.html`
```yaml
      Details: //p[@class="description"]
```
is ok i think

JaseNZC:
> Some scenes have a description `https://tour.bamvisions.com/trailers/Zoey-Takes-A-Pounding.html`,`https://tour.bamvisions.com/trailers/Doing-Something-I-Like.html`
> 
> ```yaml
>       Details: //p[@class="description"]
> ```
> 
> is ok i think

Would ya look at that did not notice that.

PR #611, opened 2021-06-29
Add new scraper for staxus.com

rudi123github:
This is a new scaper for staxus.com. It does parse data for scenes and performers.
The majority of the code lines are necessary to convert the ethnicity to a country name, I found this list in a script and it works very well, even though I think we should have a function for this (like parsedate). I felt the country feature is nice to have since the crew is quite international.


bnkai:
The map was used originally in the IAFD xPath scraper and then was converted to python
The source was from `https://en.wikipedia.org/wiki/List_of_adjectival_and_demonymic_forms_for_countries_and_nations`
This scraper and the IAFD were the only places where it was used afaik so it doesnt make much sense to add as a separate function imo (for now at least)

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+  sceneScraper:
+    common:
+      $info: //div[@class="text"]
+    scene:
+      Title:
+        selector: //title/text()
+        postProcess:
+          - replace:
+              - regex: 'Staxus Scene - '
+                with: ""
```
bnkai, line 24: Using
```suggestion
      Title: //div[@class="video-descr__title"]//h2
```
as a title saves us the pp

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+      Date:
+        selector: //div[@class='video-details'][1]
```
bnkai, line 26:
```suggestion
        selector: //div[@class='video-details'][strong='Added:']
```
using the above is safer imo

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+      Image:
+        selector:  //script[@type="application/ld+json"][contains(.,'"thumbnailUrl":')][1]
+        postProcess:
+          - replace:
+              - regex: '^.*thumbnailUrl": "'
+                with: ""
+              - regex: '.jpg.*$'
+                with: ".jpg"
+              - regex: 'contentthumbs'
+                with: "content/contentthumbs"
```
bnkai, line 49: I think there is a single script with the selector used so `[1]` is not needed
We can also use a different regex to remove a replace
```yaml
      Image:
        selector: //script[@type="application/ld+json"][contains(.,'"thumbnailUrl":')]
        postProcess:
          - replace:
              - regex: '^.*thumbnailUrl":\s"([^"]+)".*$'
                with: "$1"
              - regex: "contentthumbs"
                with: "content/contentthumbs"
```

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+      Gender:
+        fixed: Male
+      Height:
+        selector: //p[@class="flex-item"]/b[text()="Height:"]/following-sibling::span[1]/text()
+        postProcess:
+          - replace:
+              - regex: 'cm.*/.*'
+                with: ""
+              - regex: '^.*/'
+                with: ""
```
bnkai, line 66: For the height we want to keep only the digits ( we save in metric units in the db)
```yaml
        postProcess:
          - replace:
              - regex: 'cm.*'
                with: ""
              - regex: '^.*/(\s)?'
                with: ""
```
the above seems to work better since in most cases i found the cm were after the `/`

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+      Measurements:
+        selector: //p[@class="flex-item"]/b[text()="Cock size:"]/following-sibling::span[1]/text()
+        postProcess:
+          - replace:
+              - regex: 'cm.*/.*'
+                with: ""
+              - regex: '^.*/'
+                with: ""
```
bnkai, line 74: I am not sure about how to store the size, i would probably keep here only the digits like in the case above
```yaml
        postProcess:
          - replace:
              - regex: 'cm.*'
                with: ""
              - regex: '^.*/(\s)?'
                with: ""
```

scrapers/Staxus.yml
```diff
@@ -0,0 +1,382 @@
...
+              "Åland Island": "Åland Islands"
+      Image: 
+        selector: //div[@class="item active"]/img/@src
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https:"
+      Details:
+        selector: //div[@class="video-descr__section col-md-12"]//p[2]/text()
+
```
bnkai, line 381:
```yaml
      Birthdate:
        selector: //p[@class="flex-item"]/b[text()="Date of birth:"]/following-sibling::span[1]/text()
        postProcess:
          - replace:
              - regex: '(\d+)\S+\s+([a-zA-Z]+)\D+(\d+)'
                with: "$1 $2 $3"
          - parseDate: 2 January 2006
 ```
Birthdate can also be added
The above seems to work in most cases, not when the date is partial though (found cases with missing the day part)
Should we ignore it in that case or use `1` as the day?

PR #612, opened 2021-06-29
add Baberotica.yml

JaseNZC:
Covers off 4 sites. :-)


scrapers/Baberotica.yml
```diff
@@ -0,0 +1,36 @@
...
+      - baberoticavr.com
+      - teenthais.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //h1[@itemprop="name"]
+      Performers:
+        Name: //p/span/a/span
```
bnkai, line 16: Although it works i would suggest something like
```suggestion
        Name: //p/span[@itemprop="actor"]/a/span
```
as a little more specific/safe

scrapers/Baberotica.yml
```diff
@@ -0,0 +1,36 @@
...
+      Tags:
+        Name: //a[@itemprop="genre"]
+      Details:
+        selector: //div[@itemprop="description"]
+      Image:
+        selector: //meta[@itemprop="thumbnailUrl"]/@content
```
bnkai, line 22: The above selector gets a lot of images that the 1st happens to be what we need
```yaml
//div[contains(@class,"pure-u-1")or @class="container"]/meta[@itemprop="thumbnailUrl"]/@content
```
seems to work and returns a single image, it needs some testing though.
If from the suggested above we replace `thumbnailUrl` with 'datePublished' we can also get the published date of the scene (with a regex replace to clean the `T.+$` part)

JaseNZC: Should be good to go now, Did a fair bit of testing on the images and works well :-)

PR #613, opened 2021-06-29
fixes scraper break if model doesn't have bio

jeremymeyers:


bnkai:
Instead of adding those two lines i would suggest editing the `return`  to something like the below, that should catch cases where something else is missing 
```python
    return {
        'Name': model.get("name"),
        'Gender': model.get("gender").upper(),
        'URL': f"{baseUrl}{model.get('path')}",
        'Ethnicity': model.get("ethnicity"),
        'Country': model.get("country", {}).get("name"),
        'Height': str(model.get("height")),
        'Weight': str(model.get("weight")),
        'Measurements': model.get("size"),
        'Details': model.get("biography"),
        'hair_color': model.get("hair").capitalize(),
        'eye_color': model.get("eyes").capitalize(),
        'Image':
        f"https://cdn.metartnetwork.com/{model.get('siteUUID')}{model.get('headshotImagePath')}",
        'Tags': tags      
    }
```
The get() method returns `None` by default if a dict value is not present so the if  check is a bit redundant.

Belleyy:
+1 for bnkai but i think errors could still appears. 
Str()/upper()/capitalize() wont work great with Nonetype (not tested on phone atm)

bnkai:
From the phone also, so not tested but something like
```python
(model.get("hair") or "").capitalize()
```
should work for cases where a function call is needed, in this case the `capitalize`

jeremymeyers:
@bnkai so the suggestion is to add "or """ to each of the return elements?

gitgiggety:
> @bnkai so the suggestion is to add "or """ to each of the return elements?

And to use `get(...)` instead of `[...]` as well. Seems to me like Python does threath both differently, which I didn't know when writing this script (nor did I know not every model would have a bio).

But shouldn't this be applied to _every_ mapping then? As I guess it could then just as well by that a gallery or movie/scene misses some element which could lead to an error.

bnkai:
> But shouldn't this be applied to _every_ mapping then? As I guess it could then just as well by that a gallery or movie/scene misses some element which could lead to an error.

I think so too, for the purpose of this PR the performer section is enough though

PR #616, opened 2021-07-01
add WatchYouJerk.yml

JaseNZC:


bnkai:
The image part could be fetched from the `og:image` meta but the above works fine also

PR #618, opened 2021-07-01
add PlayWithRae.yml

JaseNZC:
Most of the scenes don't have a description but a few do. This does pick them up
i.e [https://playwithrae.com/trailers/Shower-with-Jayln-Part-1-Video.html](url)

### DIFF
--- a/scrapers/PlayWithRae.yml
+++ b/scrapers/PlayWithRae.yml
@@ -1,0 +1,31 @@
+name: PlayWithRae
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - playwithrae.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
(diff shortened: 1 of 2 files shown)


scrapers/PlayWithRae.yml
```diff
@@ -0,0 +1,31 @@
...
+      Date:
+        selector: //div[@class='videoInfo clear']/p[contains(span,"Date Added:")]/text()
+        postProcess:
+          - parseDate: January 2, 2006
+      Performers:
+        Name: //li[@class="update_models"]//a
+      Details:
+        selector: //div[@class="videoDetails clear"]/h3/following-sibling::text()
+      Tags:
+        Name: //li[@class="label" and contains(text(), "Tags:")]/following-sibling::text()
```
bnkai, line 21: Instead of `::text()`,`::li/a` could be used here
It works ok though so no biggie

PR #620, opened 2021-07-02
add Waybig.yml

JaseNZC:
As requested on discord.

Covers scenes and performers.

### DIFF
--- a/scrapers/Waybig.yml
+++ b/scrapers/Waybig.yml
@@ -14,7 +14,7 @@ xPathScrapers:
     scene:
       Title: //meta[@name="twitter:title"]/@content
       Date:
-        selector: //i[@class="far fa-calendar"]/following-sibling::text()
+        selector: //div/i[@class="far fa-calendar"]/../span
         postProcess:
           - replace:
               - regex: ^(\d{4}-\d{2}-\d{2}).+
@@ -23,15 +23,19 @@ xPathScrapers:
       Details:
         selector: //div[@class="expand-inner"]
       Tags:
-        Name: //div[contains(text(),"Tags:")]/following-sibling::a
+        Name: //div[@class="content-links -tags"]//a
       Performers:
-        Name:
-          selector: //div[@class="models-box"]/a
+        Name: //div[@class="models-box"]/a
+        URL: //div[@class="models-box"]/a/@href
       Image:
         selector: //video/@poster
       Studio:
         Name:
-          fixed: Waybig
+          selector: //a[@class="hoverLink"]
+          postProcess:
+            - replace:
+                - regex: '^Visit\s+'
+                  with: ""
   performerScraper:
     performer:
       Name: //div[@class="title-col -normal col"]/h2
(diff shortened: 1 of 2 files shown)


scrapers/Waybig.yml
```diff
@@ -0,0 +1,67 @@
...
+  - action: scrapeXPath
+    url:
+      - waybig.com/pornstars/
+    scraper: performerScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //meta[@name="twitter:title"]/@content
+      Date:
+        selector: //i[@class="far fa-calendar"]/following-sibling::text()
```
bnkai, line 17:
```yaml
        selector: //div/i[@class="far fa-calendar"]/../span
```
seems safer

Also a note if you go to the last page by date you will notice some scenes (very few) have weird dates
Using the selector we can get either `0000-00-00` or something like `0000-02-18` 
Regex replacing `^(00.+)` with a `""` still has issues with parseDate so i think we should just ignore those dates

scrapers/Waybig.yml
```diff
@@ -0,0 +1,67 @@
...
+        postProcess:
+          - replace:
+              - regex: ^(\d{4}-\d{2}-\d{2}).+
+                with: $1
+          - parseDate: 2006-01-02
+      Details:
+        selector: //div[@class="expand-inner"]
+      Tags:
+        Name: //div[contains(text(),"Tags:")]/following-sibling::a
```
bnkai, line 26:
```yaml
        Name: //div[@class="content-links -tags"]//a
```

scrapers/Waybig.yml
```diff
@@ -0,0 +1,67 @@
...
+      Performers:
+        Name:
+          selector: //div[@class="models-box"]/a
```
bnkai, line 29:
```yaml
        Name: //div[@class="models-box"]/a
        URL: //div[@class="models-box"]/a/@href
```
since you already made a performer scraper adding the performer url here is very useful

scrapers/Waybig.yml
```diff
@@ -0,0 +1,67 @@
...
+      Image:
+        selector: //video/@poster
+      Studio:
+        Name:
+          fixed: Waybig
```
bnkai, line 34: the site seems to have scenes from various studios
using
```yaml
        Name:
          selector: //a[@class="hoverLink"]
          postProcess:
            - replace:
                - regex: '^Visit\s+'
                  with: ""
```
seems more accurate

scrapers/Waybig.yml
```diff
@@ -0,0 +1,67 @@
...
+  performerScraper:
+    performer:
+      Name: //div[@class="title-col -normal col"]/h2
+      Gender:
+        fixed: Male
+      Measurements:
+        selector: //span[contains(text(),"Cock size:")]/following-sibling::text()
```
bnkai, line 41:
```yaml
        selector: //span[contains(text(),"Cock size:")]/following-sibling::text()
        postProcess:
          - replace:
              - regex: ^
                with: "0''"
          - feetToCm: true
```
a small hack to get the size in cm (accuracy is not that good though since the decimal point will be ignored)

JaseNZC:
That should have taken care of everything 👍

PR #623, opened 2021-07-03
add Only3xGirls.yml

JaseNZC:
I was going to update Only3xGirls but it appears I never submitted it in the  first place. So erm yeah dunno what happened there.

Anyway this adds that site and another.

### DIFF
--- a/scrapers/Only3xGirls.yml
+++ b/scrapers/Only3xGirls.yml
@@ -1,0 +1,32 @@
+name: Only3xGirls
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - only3xgirls.com
+      - purebj.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="sv_update_title"]/h1
+      Date:
+        selector: //ul[@class="dotsbetween"]//span[contains(.,"since")]/following-sibling::text()
+        postProcess:
+          - parseDate: 01/02/2006
+      Performers:
+        Name: //ul[@class="dotsbetween"]//a
+      Details:
+        selector: //div[@class='set-desc']
+        concat: "\n\n"
+      Tags:
+        Name: //div[@class="content tag"]//a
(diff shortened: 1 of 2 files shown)


scrapers/Only3xGirls.yml
```diff
@@ -0,0 +1,32 @@
...
+        postProcess:
+          - parseDate: 01/02/2006
+      Performers:
+        Name: //ul[@class="dotsbetween"]//a
+      Details:
+        selector: //div[@class='set-desc']
+        concat: "\n\n"
+      Tags:
+        Name: //div[@class="content tag"]//a
+      Image: //img[@class="stdimage thumbs"]/@src0_2x | //img[@class="thumbs"]/@src
```
bnkai, line 24: this need to be more specific we get a lot of images
```yaml
Image: //span[@class="update_thumb"]/img[@class="stdimage thumbs"]/@src0_2x | //span[@class="update_thumb"]/img[@class="thumbs"]/@src
```
should work ok, needs more testing though

scrapers/Only3xGirls.yml
```diff
@@ -0,0 +1,32 @@
...
+        selector: //div[@class="sv_update_title"]/h1
+      Date:
+        selector: //ul[@class="dotsbetween"]//span[contains(.,"since")]/following-sibling::text()
+        postProcess:
+          - parseDate: 01/02/2006
+      Performers:
+        Name: //ul[@class="dotsbetween"]//a
+      Details:
+        selector: //div[@class='set-desc']
+        concat: "\n\n"
```
bnkai, line 21: the concat doesnt really do anything since there is a single div

if we wanted to  maintain the format a bit
```yaml
      Details:
        selector: //div[@class='set-desc']/*[local-name()="em" or local-name()="p"]
        concat: "\n\n"
```
looks ok from a few scenes but we would need to check if  only `p` and `em` elements are enough or something else is also used
```

JaseNZC: Yup that tests out well :-)

Issue #586, opened 2021-06-03
Scrape from Shoko Server with its api


MrX292:
a script scraper for Shoko Server through the api should make it a lot easier to scrape hentai
Documentation: https://shokoserver.docs.apiary.io/#

PR #628, opened 2021-07-04
Shoko api Script Scraper

MrX292:
#586 can be closed
not sure if i should add it to Supported Sites For Scraping since it scrapes your own shoko server?


bnkai:
> #586 can be closed
> not sure if i should add it to Supported Sites For Scraping since it scrapes your own shoko server?

You can close the issue if you want, since you are the one that created it. Especially if this covers what you mentioned there.

Scrapers like this, the torrent scraper, the xbvrdb one, ... should have a separate section as they do not work with urls. There is no need to do anything else for this PR, when we make that section/page we will update as needed.

Issue #634, opened 2021-07-08
House of Gord scraper returns 403


quirkyjoe:
Attempting to use the House of Gord scraper returns a 403 error in the error popup.

Example URL: https://www.houseofgord.com/d/another-naked-pussy-walker

Result:
```
Error
http error 403:Forbidden
```

JaseNZC:
> Attempting to use the House of Gord scraper returns a 403 error in the error popup.
> 
> Example URL: https://www.houseofgord.com/d/another-naked-pussy-walker
> 
> Result:
> 
> ```
> Error
> http error 403:Forbidden
> ```

I have created a Pull Request for it here if you wanted to try it. #635

quirkyjoe:
After replacing my current yml file and restarting Stash, trying to scrape yields the following:

![image](https://user-images.githubusercontent.com/85048415/125006735-5a2e8080-e04e-11eb-901b-8be0b7326315.png)

```
Error

exec: "google-chrome": executable file not found in $PATH
```

I don't _think_ this is on my end, because I have successfully used other scrapers on the two machines I tested the fix on. (?)

Thank you very much for your time! If I can provide any further info, let me know.

JaseNZC:
> After replacing my current yml file and restarting Stash, trying to scrape yields the following:
> 
> ![image](https://user-images.githubusercontent.com/85048415/125006735-5a2e8080-e04e-11eb-901b-8be0b7326315.png)
> 
> ```
> Error
> 
> exec: "google-chrome": executable file not found in $PATH
> ```
> 
> I don't _think_ this is on my end, because I have successfully used other scrapers on the two machines I tested the fix on. (?)
> 
> Thank you very much for your time! If I can provide any further info, let me know.

You need to have CDP enabled for the updated yml to work :-)

Under Settings->Configuration you will see the setting to populate.

The one here is my setup as I use docker so yours will be different depending on what operating system you are using for stash.
Ie say for windows would be something like.
`c:\Program Files\Google\Chrome\Application\chrome.exe`

<img width="1148" alt="Screen Shot 2021-07-09 at 3 38 27 PM" src="https://user-images.githubusercontent.com/64925341/125019868-d87d3800-e0cb-11eb-9d88-fc2bff513544.png">

Ruthalas:
I'm also running Stash in a docker, would you mind pointing me at the chrome docker you're using?

JaseNZC:
[...]

PR #635, opened 2021-07-08
Update houseofgord.yml

JaseNZC:
Fix 403 error as per [https://github.com/stashapp/CommunityScrapers/issues/634#issue-939947498](url)


bnkai:
Using CDP seems enough/works ok for me. (Lets hope it will last a bit )

PR #636, opened 2021-07-09
Diabolic

JaseNZC:
Its a pity there is not a section for Performers and Tags for the movie side of thing as this would be a perfect one of being able to add all that info for the movie in stash.

### DIFF
--- a/scrapers/Diabolic.yml
+++ b/scrapers/Diabolic.yml
@@ -1,0 +1,49 @@
+name: Diabolic
+sceneByURL:
+  - action: scrapeXPath
+    scraper: sceneScraper
+    url:
+      - diabolic.com/en/video/
+movieByURL:
+  - action: scrapeXPath
+    scraper: movieScraper
+    url:
+      - diabolic.com/en/movie/
+xPathScrapers:
+  sceneScraper:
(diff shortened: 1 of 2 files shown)


scrapers/Diabolic.yml
```diff
@@ -0,0 +1,47 @@
...
+        selector: //li[@class="updatedDate"]/text()
+        postProcess:
+          - replace:
+              - regex: ":"
+                with: ""
+          - parseDate: 2006-01-02
+      Performers:
+        Name: //div[@class="sceneCol sceneColActors"]/a
+      Tags:
+        Name: //div[@class="sceneCol sceneColCategories"]/a
```
bnkai, line 27: It seems that to get all tags we need to use CDP as tags that are displayed in the actual page are updated using a script.
(the main tags from the above are usually below 10)
```yaml
      Tags:
        Name: //div[contains(@class,"ScenePlayerHeaderDesktop-CategoriesContainer")]//a/@title
```
using the above and CDP seems to get all the visible tags

JaseNZC: Did not even notice that, should be done now :-)

PR #638, opened 2021-07-09
Xpath Scraper for FreakMobMedia

Belleyy:


scrapers/FreakMobMedia.yml
```diff
@@ -0,0 +1,31 @@
...
+      Date:
+        selector: //div[@class='views-single']/span/text()
+        postProcess:
+          - parseDate: January 2, 2006
+      Image:
+        selector: //meta[@property="og:image"]/@content
+        postProcess:
+          - replace:
+            - regex: .+(?:poster=")([^"]+).+
+              with: https://www.loveherfeet.com$1
```
bnkai, line 22: leftover from copy/paste ?

scrapers/FreakMobMedia.yml
```diff
@@ -0,0 +1,31 @@
...
+      Studio:
+        Name:
+          fixed: FreakMobMedia
+      Tags:
+        Name: //span[@class="meta-info" and text()="Category"]/following-sibling::a/text()|//span[@class="meta-info" and text()="Tag"]/following-sibling::a/text()
```
bnkai, line 27: can be simplified to
```yaml
        Name: //span[@class="meta-info"][text()="Category" or text()="Tag"]/following-sibling::a/text()
```

scrapers/FreakMobMedia.yml
```diff
@@ -0,0 +1,31 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - freakmobmedia.com/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="bodywhite"]/div[@class="title"]//h3/text()
+      Details: //div[@class='description']/text()
```
bnkai, line 11: there are cases where 
- there is more than one text eg `http://www.freakmobmedia.com/sandra-luberc-gets-anal-creampie-from-bbc/`
- a `p` is inside the div instead of plain text `http://www.freakmobmedia.com/julianna-vega-and-queen-rogue-pov/`

In both cases i think a `concat` is needed and maybe a union of the text() and p/text() cases ?

Belleyy: Found that there is a `//meta[@name="description"]/@content`. Find it everywhere. Easier to deal with.
Added a regex to remove stupid things

scrapers/FreakMobMedia.yml
```diff
@@ -0,0 +1,33 @@
...
+    scene:
+      Title: //div[@class="bodywhite"]/div[@class="title"]//h3/text()
+      Details: 
+          selector: //meta[@name="description"]/@content
+          postProcess:
+            - replace:
+                - regex: "Join FreakMob Media Now For Full Access!"
+                  with: ""
+                - regex: "JOIN NOW or BUY JUST THIS CLIP"
+                  with: ""
```
bnkai, line 18: i would suggest making the match case insensitive
and using a single regex for the cleaning up
eg
```yaml
        postProcess:
          - replace:
              - regex: "(?i)(Join FreakMob Media Now For Full Access!)|(JOIN NOW or BUY JUST THIS CLIP)"
                with: ""
```

PR #639, opened 2021-07-11
add DevilsTgirls.yml

JaseNZC:
Requested on Discord. I also added in gallery and movie as well, It did have performers but not really enough info worth getting.

### DIFF
--- a/scrapers/DevilsTgirls.yml
+++ b/scrapers/DevilsTgirls.yml
@@ -56,12 +56,6 @@ xPathScrapers:
       Studio:
         Name:
           fixed: DevilsTgirls
-      URL:
-        selector: //span[@class="fromDvdValue"]/a/@href
-        postProcess:
-          - replace:
-              - regex: ^
-                with: "https://www.devilstgirls.com/"
   movieScraper:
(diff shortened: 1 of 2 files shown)


JaseNZC:
Cleaned up my incompetence.

scrapers/DevilsTgirls.yml
```diff
@@ -0,0 +1,80 @@
...
+        Name: //span[@class="actorsValue"]/a
+      Studio:
+        Name:
+          fixed: DevilsTgirls
+      URL:
+        selector: //span[@class="fromDvdValue"]/a/@href
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https://www.devilstgirls.com/"
```
bnkai, line 64: thats the movie url 
i would either remove the `URL` (the user will have to fill it to scrape anyway) or construct it from the `//link[@rel="next"]/@href` (again probably not worth the trouble)

JaseNZC: Yeah agreed it's better just to remove it as it has to be filled in anyway.

PR #641, opened 2021-07-12
add HeavyOnHotties.yml

JaseNZC:


scrapers/HeavyOnHotties.yml
```diff
@@ -0,0 +1,23 @@
...
+      Title:
+        selector: //div[@class="vid-title clearfix text-center-mobile"]/h2
+      Date:
+        selector: //span[contains(@class, "released title")]/strong/text()
+        postProcess:
+          - parseDate: Jan 2, 2006
+      Details: //div[@class="col-md-9 col-sm-8 clearfix"]/p
+      Performers:
+        Name: //span[@class="feature title"]//a
+      Image: //video[@id="video"]/@poster
```
bnkai, line 19:
```suggestion
      Image:
        selector: //video[@id="video"]/@poster
        postProcess:
          - replace:
              - regex: ^//
                with: "https://"
```
The schema is missing from the image URL (at least for the scenes i tested). You'll notice the error if you try to `save` the scene.

JaseNZC: Have fixed that, Hmmmm I did not have an issue with the scenes when testing and saving so dunno what was happening there. 
All good now though :-)

PR #643, opened 2021-07-16
Added scraper for vod section on SubbyHubby/ClubDom

nocrad349:


scrapers/SubbyHubby.yml
```diff
@@ -33,4 +38,35 @@ xPathScrapers:
             - map:
                 //www.clubdom.com/tour/: Club Dom
                 //www.subbyhubby.com/tour/: Subby Hubby
-# Last Updated May 21, 2021
+  vodScraper:
+    scene:
+      Title:
+        selector: //span[@class="title_bar_hilite"]
+      Date:
+        selector: //div[@class="cell update_date"]
```
bnkai, line 46: The above selector seems to match all the dates from the related videos in subbyhubby, eg `https://www.subbyhubby.com/vod/scenes/SavingOurMarriagePart2_vids.html`
Something like 
```yaml
        selector: //div[@class="gallery_info"]/div[1]//div[@class="cell update_date"]
```
gets only a single date

nocrad349: Fixed in https://github.com/stashapp/CommunityScrapers/pull/643/commits/cbb2be80344e0b8a24fb70aefc23febd7972fad3

PR #644, opened 2021-07-18
create AllAnalAllTheTime.yml

JaseNZC:
Requested as it did not work under VNA girls. Just made its own scraper.


bnkai:
yeah VNA need to be split, i left it as a draft for too long i think

PR #649, opened 2021-07-19
Duplicating parsing from LegalPorno into Pornworld

ryosaeba75:
Both sites are similar. This add performerByURL and sceneByScrapper Pornworld scrapper.

### DIFF
--- a/scrapers/pornworld.py
+++ b/scrapers/pornworld.py
@@ -1,0 +1,51 @@
+import json
+import sys
+import requests
+from pathlib import Path
+
+def debug(t):
+  sys.stderr.write(t + "\n")
+
+def query_url(query):
+  res = requests.get(f"https://www.pornworld.com/api/autocomplete/search?q={query}")
+  data = res.json()
+  results = data['terms']
+  if len(results) > 1:
+    debug("Multiple results. Taking first.")
+  return results[0]
+
+def detect_delimiter(title):
+  delimiters = [" ", "_", "-", "."]
+  for d in delimiters:
+    if d in title:
+      return d
+
+  debug(f"Could not determine delimiter of `{title}`")
+
+def find_scene_id(title):
+  # Remove file extension
+  title = Path(title).stem
+  title = title.replace("'", "")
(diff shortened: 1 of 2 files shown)


ryosaeba75:
Does this need anything on my side or is it just waiting for approval?

bnkai:
@ryosaeba75 I haven't had the chance to test/check the scraper yet. Xpath/JSON ones are easier/faster to check so script scrapers usually end up at the bottom of the backlog. I will definitely check it out before the weekend.

Belleyy:
Is there any reason to use a python script instead of a scrapeJson for sceneByFragment ?

Belleyy:
Is there any reason why using a python script instead of a JSON scraper ?
Tested a JSON scraper and it work.

ryosaeba75:
@bnkai no worries, I was just checking to be sure that wasn't something missing on my side.
@Belleyy not sure, I just duplicated the one from legalporno as the website are similar. They probably sould be merged but I wasn't sur how to do that, especially for the `sceneByFragment` part.

bnkai:
It seems that the only thing the script does is clear the filename and get the id to search for.
We can now do that using `queryURLReplace` so i would suggest closing this PR in favor of #687

Belleyy:
> Is there anyway to have a common `sceneByFragment` for legalporno and pornworld? I know that it's possible for `sceneByUrl`but not sur how for `sceneByFragment`. 

No you can't have multiple `sceneByFragment` in a yml. 

> That could be the only reason to favor python over the classic yml/xpath scrapper.

How so ? Your python right now only deal with Pornworld too, and to launch the script you still need a yml file so you are limited with `sceneByFragment`, only way to have a common is having a script that search on Pornworld then on Legalporno.

ryosaeba75:
Yes your edit is what I was thinking about, having two yml referencing a single python, with probably a parameter to make the distinction between both srapper. But as you noted, right now either my scrapper nor your's deal with both so for now there is no use for python. I'm closing this PR in favor of your's and if/when I (or someone) else merge both scrappers (via a common python file) it will be another PR.

PR #650, opened 2021-07-19
Dorcel: Add Movie Scrapper

ryosaeba75:


scrapers/DorcelClub.yml
```diff
@@ -21,8 +27,39 @@ xPathScrapers:
...
+
+  movieScraper:
+    movie:
+      Name: //img[contains(@class, "cover")]/@alt
+      Duration:
+        selector: //span[@class='duration']/text()
+        # postProcess:
+        #   - replace:
+        #     - regex: ([0-9]h)? ([0-9]m)? ([0-9])?
+        #     with: $1:$2:$3
```
bnkai, line 47: a little complex but needed
```yaml
        postProcess:
           - replace:
             - regex: ([0-9]+)h\s*([0-9]+)?
               with: "$1:$2:00"
             - regex: ([0-9]+)m\s*([0-9]+)?
               with: "0:$1:$2"
             - regex: '^:'
               with: "0:"
             - regex: ':$'
               with: ":00"
             - regex: '::'
               with: ':00:'
```

scrapers/DorcelClub.yml
```diff
@@ -21,8 +27,39 @@ xPathScrapers:
...
+          - replace:
+            - regex: Director\s:\s(.*)
+              with: $1
+      FrontImage:
+        selector: //img[contains(@class, "cover")]/@data-src
+        postProcess:
+          - replace:
+            - regex: ([^\s]*)\s1x
+              with: $1
+      Synopsis: //span[@class="full"]/p/text()
```
bnkai, line 63:
```suggestion
      Synopsis: //span[@class="full"]/p|/div[@class="content-text"]/p
```
In cases where the synopsis is short there (no more to expand) there is no span

scrapers/DorcelClub.yml
ryosaeba75: Update in e8790ec

scrapers/DorcelClub.yml
ryosaeba75: Updated in e8790ec

PR #651, opened 2021-07-19
Add Transerotica support

ryosaeba75:


bnkai:
- We can avoid most indexes if we use a more specific selector
- Studio is missing
- Since CDP is not used we can just skip the driver section

The below seems ok
```yaml
name: Transerotica
sceneByURL:
  - action: scrapeXPath
    url:
      - transerotica.com/trailers
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $update: //div[@class="centerwrap clear"]/div[@class="updateDetails"]
    scene:
      Title: //h1[@class='title_bar']
      Image: //div[@id="player"]/video/@poster
      Date:
        selector: $update//span[@class="upddate"][1]
        postProcess:
          - parseDate: "01/02/2006"
      Details:
        selector: $update/p[starts-with(text(),"Description:")]
        postProcess:
          - replace:
              - regex: '^Description:\s*'
                with:
      Performers:
        Name:
          selector: $update//span[@class="tour_update_models"]/a
      Tags:
        Name:
          selector: //meta[@name="keywords"]/@content
          split: ","
      Studio:
        Name:
          fixed: Trans Erotica
# Last Updated July 25, 2021
```

PR #655, opened 2021-07-21
add FilthFlix.com

JaseNZC:


scrapers/FilthFlix.yml
```diff
@@ -0,0 +1,27 @@
...
+      Title:
+        selector: //h2[@class="video__headline"]
+      Date:
+        selector: //strong[contains(text(),"Published:")]/following-sibling::text()
+        postProcess:
+          - parseDate: 02 Jan 2006
+      Performers:
+        Name: //strong[contains(text(),"Models:")]/following-sibling::a
+      Tags:
+        Name: //strong[contains(text(),"Tags:")]/following-sibling::a
```
bnkai, line 19: The `models` are under the tags so the following sibling for the Tags includes the Performers as tags...
I would use the following instead
```yaml
      Performers:
        Name:  //p[@class="video__text"]//a[contains(@href,"/models/")]
      Tags:
        Name:  //p[@class="video__text"]//a[contains(@href,"/movies/")]
```

JaseNZC: Yup that is a much better way.

PR #657, opened 2021-07-21
add GymRotic

JaseNZC:
Covers off 4 sites :-)

### DIFF
--- a/scrapers/GymRotic.yml
+++ b/scrapers/GymRotic.yml
@@ -1,0 +1,41 @@
+name: DollRotic
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - dollrotic.com/trailers/
+      - femlatex.com/trailers/
+      - gymnastic.xxx/trailers/
+      - gymrotic.com/trailers/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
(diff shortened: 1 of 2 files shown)


scrapers/GymRotic.yml
```diff
@@ -0,0 +1,41 @@
...
+                with: ""
+      Studio:
+        Name:
+          selector: //base/@href
+          postProcess:
+            - map:
+                https://www.dollrotic.com/: DollRotic
+                https://www.gymnastic.xxx/: GymnasticXXX
+                https://www.femlatex.com/: FemLatex
+                https://www.gymrotic.com/: GymRotic
```
bnkai, line 40: depending on how you access the site you can end up with a `https://www.xxxx.com` or a `https://xxxxx.com` url
i would just use the author meta and do a simple regex replace eg
```yaml
      Studio:
        Name:
          selector: //meta[@name="author"]/@content
          postProcess:
            - replace:
                - regex: '\.com$'
                  with: ""
```
you might want to add an extra replace/map for the gymnastic site if you like the GymnasticXXX vs the  Gymnastic.xxx you will get from the proposed change

JaseNZC: Nar I am not worried about it. :-)

PR #659, opened 2021-07-21
add Corbinfisher.com

JaseNZC:


scrapers/CorbinFisher.yml
```diff
@@ -0,0 +1,28 @@
...
+        selector: //h1
+      Date:
+        selector: //div[@class="added"]
+        postProcess:
+          - replace:
+              - regex: (^\S+\s+)(.+\d{4})(.+)
+                with: $2
+          - parseDate: January 2, 2006
+      Performers:
+        Name: //div[@class="modelFeaturing"]/a | //div[@class="modelFeaturing"]//a
```
bnkai, line 20: `//div[@class="modelFeaturing"]//a` includes the `//div[@class="modelFeaturing"]/a` result.  Was there a reason for the union and not just using the second (//a) selector?
I would also add the URL field as for most performers only the name part is used
```yaml
        Name: //div[@class="modelFeaturing"]//a
        URL: //div[@class="modelFeaturing"]//a/@href
```

scrapers/CorbinFisher.yml
```diff
@@ -0,0 +1,28 @@
...
+      Details:
+        selector: //div[@class="description"]/p
```
bnkai, line 22: you forgot the concat for cases like `https://www.corbinfisher.com/tour/trailers/acm-william-fills-up-quinn.html`
```yaml
        selector: //div[@class="description"]/p
        concat: "\n\n"
```

JaseNZC: Yeah I dunno why I had that union there as its not needed.
All fixed now. :-)

PR #660, opened 2021-07-22
add NakedSword

JaseNZC:
As per requested on Discord.
Covers Scenes and Movies.

Look and no following-sibling::text() , hahahahahaha


bnkai:
> Look and no following-sibling::text() , hahahahahaha
Yeah if all sites used a separate div for each meta it would be a lot easier :P

PR #664, opened 2021-07-24
Update Jav Python

Belleyy:
Instead of having a single mirror, it will run a check of all site to get the one without Protection.

Probably need some more test just be sure.

### DIFF
--- a/scrapers/JavLibrary_python.py
+++ b/scrapers/JavLibrary_python.py
@@ -24,7 +24,8 @@ DEBUG_MODE = True
 STASH_SUPPORTED = False
 # Tags you don't want to see appear in Scraper window
 IGNORE_TAGS = ["Features Actress", "Digital Mosaic", "Hi-Def", "Risky Mosaic",
-               "Beautiful Girl", "Blu-ray", "Featured Actress", "VR Exclusive", "MOODYZ SALE 4"]
+               "Beautiful Girl", "Blu-ray", "Featured Actress", "VR Exclusive", "MOODYZ SALE 4",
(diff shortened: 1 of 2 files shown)


scrapers/JavLibrary_python.py
```diff
@@ -174,35 +175,43 @@ def debug(q):
...
+        if response.url == "https://www.javlib.com/maintenance.html":
+            debug("[{}] Maintenance".format(site))
+        if "Why do I have to complete a CAPTCHA?" in response.text or "Checking your browser before accessing" in response.text:
+            debug("[{}] Protected by Cloudflare".format(site))
+        elif response.status_code != 200:
+            debug("[{}] Other issue ({})".format(site, response.status_code))
+        else:
+            debug("[{}] Using this site for scraping ({})".format(site, response.status_code))
+            debug("[DEBUG] ======================================")
+            return site, response
```
bnkai, line 193: I would move the `response.status_code != 200` check to be the first one, as i am not sure if there will be a `response.text` or a `response.url` if the `status_code` is that of an error.

Belleyy: I don't think they can crash anyways:
`response.url` is None by default, since it's a compare it won't crash (`if None == "test"` work)
`response.text` should be same thing.

If i remember correctly the cloudflare page or maintenance is not a status_code 200 so if i put the `response.status_code != 200` first, this condition will always get trigger before other.

bnkai: yeh makes sense now

PR #665, opened 2021-07-27
Update ThePornDB to use the parent not the site name

Tweeticoats:
Entries from theporndb use the parent not the site id.
Scenes have a site specific performer id and this is usually linked to a performer id.
These can be different when a performer uses different names over time or some sites use stupid alias's and only use the first name for example.
You should return the canonical name of that performer.


bnkai:
Shouldnt  the same apply for the scene scraper ? Line 77 ?
```yaml
        Name: data.performers.#.name
```

PR #666, opened 2021-07-29
Rework AdultimeAPI scraper

Belleyy:
- Dropped the local support
- Probably better result using fragment
- Look better 😄


peolic:
Please also add it to `SCRAPERS-LIST.md` as it is missing from the list, as pointed out on Discord.

PR #668, opened 2021-07-29
Add ATKGirlfriends scene/performer xPath scraper

bnkai:
A scraper from the discord channel
Adding as a draft atm as it needs a bit of `cleaning`


Belleyy:
Cleaning is good word, don't think you need CDP, cookie, headers...

Belleyy:
Look like only `start_session_galleria` cookie is necessary with the headers.
ValueRandom: 26 don't work.

bnkai:
closing
switching to #812 instead

PR #670, opened 2021-07-30
Update Jav python (R18 API support)

Belleyy:
Switched to the API instead of using Xpath.


scrapers/JavLibrary_python.py
```diff
@@ -269,7 +269,14 @@ def r18_search(html, xpath):
         r18_result["series_name"] = r18_search_serie
     if r18_search_url:
         r18_search_url = r18_search_url[0]
-        r18_main_html = sendRequest(r18_search_url, R18_HEADERS)
+        r18_id = re.match(r".+id=(.+)\/.+", r18_search_url)
```
bnkai, line 272: should be
```python
        r18_id = re.match(r".+id=(.+)/.*", scene_url)
```

as `/` doesnt need to be escaped
and urls may be clean already so nothing after the `id=xxxx/` part

scrapers/JavLibrary_python.py
```diff
@@ -404,11 +411,16 @@ def th_imagetoBase64(imageurl, typevar):
     scene_domain = re.sub(r"www\.|\.com", "", urlparse(scene_url).netloc)
     # Url from Javlib 
     if scene_domain in SITE_JAVLIB:
-        debug("[DEBUG] Using search with URL: {}".format(scene_url))
+        debug("[DEBUG] Using URL: {}".format(scene_url))
         jav_main_html = sendRequest(scene_url, JAV_HEADERS)
     elif "r18.com" in scene_url:
-        debug("[DEBUG] Using search with URL: {}".format(scene_url))
-        r18_main_html = sendRequest(scene_url, R18_HEADERS)
+        r18_id = re.match(r".+id=(.+)\/.+", scene_url)
```
bnkai, line 417: same as above
```python
        r18_id = re.match(r".+id=(.+)/.*", scene_url)
```

bnkai:
also line 556 should now be
```python
    scrape['url'] = r18_result['url']
```
?
the `[0]` gets only the `h` from the url?

Belleyy:
> the `[0]` gets only the `h` from the url?

I was kind of lazy to edit last part, forget to transform the URL to a list... anyway fixing all (list to str) so it's simple.

PR #679, opened 2021-08-05
Update PaperStreetMedia.yml

rod8687:
added momswap


Belleyy:
You need to update SCRAPERS-LIST.md too.

PR #683, opened 2021-08-08
add SmokingMina.yml

JaseNZC:
Requested via Discord


bnkai:
The oldest video `http://tour.smokingmina.com/view/2855/feeling-you` has a wrong year (-0001) and the 9-10 scenes after that have `1970` as a year, i don't think that its worth it to regex remove those dates
So merging as is

JaseNZC:
> The oldest video `http://tour.smokingmina.com/view/2855/feeling-you` has a wrong year (-0001) and the 9-10 scenes after that have `1970` as a year, i don't think that its worth it to regex remove those dates though.
> So merging as is

Yeah I did notice that and forgot to mention it, Since it did only cover a few scene's I just left it as is.

PR #687, opened 2021-08-10
Update pornworld.yml (Xpath)

Belleyy:
Not sure for title if we keep the `featuring [MALE ACTOR]`

Edit: Just saw that there is already a PR about Pornworld (#649) so idk what to do 🤷

### DIFF
--- a/scrapers/pornworld.yml
+++ b/scrapers/pornworld.yml
@@ -27,7 +27,12 @@ xPathScrapers:
             - regex: featuring.+
               with: ""
       Date: //i[contains(@class,"calendar3")]/text()
-      Details: //meta[@name="description"]/@content
+      Details:
+        selector: //meta[@name="description"]/@content
+        postProcess:
+          - replace:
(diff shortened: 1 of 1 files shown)


scrapers/pornworld.yml
```diff
@@ -1,24 +1,40 @@
...
-      Date: //span[@title="Release date"]/a/text()
-      Details: //div[@class="scene-description__row"]/dt[contains(.,"Description")]/../dd/text()
+      Title: 
+        selector: //h1
+        postProcess:
+          - replace:
+            - regex: featuring.+
+              with: ""
+      Date: //i[contains(@class,"calendar3")]/text()
+      Details: //meta[@name="description"]/@content
```
bnkai, line 30: In some scenes there is a `weird` char when you scrape eg `https://pornworld.com/watch/207450`
A replace is needed to clean that up
```yaml
      Details:
        selector: //meta[@name="description"]/@content
        postProcess:
          - replace:
              - regex: '\x{0092}' # unicode char 0092 (looks like ')
                with: "'"
```

PR #695, opened 2021-08-23
Update Mylf.yml for title scraping

sharklazerwallo:
Look for an `h2` tag instead of `div` for scraping the title.


Belleyy:
Need to Update `# Last Updated` line

PR #697, opened 2021-08-26
Update ZeroTolerance.yml

Belleyy:
No longer have same xpath :(

For image some url don't have the correct thumbnail so my selector ignore it (http://www.thirdmovies.com/videos/yurizan-beltran-car-wash-slut)

### DIFF
--- a/scrapers/ZeroTolerance.yml
+++ b/scrapers/ZeroTolerance.yml
@@ -2,12 +2,15 @@ name: "ZeroTolerance"
 sceneByURL:
   - action: scrapeXPath
     url:
+      - ztod.com/videos/
+    scraper: sceneScraperZtod
+  - action: scrapeXPath
+    url:
       - thirdmovies.com/better/videos/
       - thirdmovies.com/videos/
-      - ztod.com/videos/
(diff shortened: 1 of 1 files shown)


bnkai:
It seems to me that the `better` url scenes are still ok with the old scraper?
Shouldn\t it be ?
```yaml
sceneByURL:
  - action: scrapeXPath
    url:
      - thirdmovies.com/better/videos/
      - ztod.com/videos/
    scraper: sceneScraperZtod
  - action: scrapeXPath
    url:
      - thirdmovies.com/videos/
    scraper: sceneScraperThird
```

Belleyy:
> 
> 
> It seems to me that the `better` url scenes are still ok with the old scraper?
> Shouldn\t it be ?
> 
> ```yaml
> sceneByURL:
>   - action: scrapeXPath
>     url:
>       - thirdmovies.com/better/videos/
>       - ztod.com/videos/
>     scraper: sceneScraperZtod
>   - action: scrapeXPath
>     url:
>       - thirdmovies.com/videos/
>     scraper: sceneScraperThird
> ```

good point forget about this url...

PR #698, opened 2021-08-30
Adding selfiesuck site

Jayran76:


scrapers/selfiesuck.yml
```diff
@@ -0,0 +1,32 @@
...
+          - parseDate: January 2, 2006
+      Tags:
+        Name: /html/body/div[1]/div/div[7]/ul/li/a
+      Performers:
+        Name:
+          selector: /html/body/div[1]/div/div[6]/ul/li[2]/a
+      Studio:
+        Name: 
+          fixed: Selfie Suck
+# Last Updated August 30, 2021
```
JaseNZC, line 32: - To help simplify things and get rid of all the [ ] which if a small change in the site is made can break the scraper I would suggest the following.

- Your image would not load either so the change also fixes that.

- Can you also please update the SCRAPERS-LIST.md file 

```name: "SelfieSuck"
sceneByURL:
  - action: scrapeXPath
    url:
      - selfiesuck.com
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    scene:
      Title:
        selector: //div[@class="videoDetails clear"]/h3
      Details: //div[@class="videoDetails clear"]/p
      Image: 
        selector: //img[@width="1160"]/@src0_1x
        postProcess:
          - replace:
              - regex: ^
                with: "https://www.selfiesuck.com"
      Date:
        selector: //span[contains(text(),"Date Added:")]/following-sibling::text()
        postProcess:
          - parseDate: January 2, 2006
      Tags:
        Name: //li[contains(text(),"Tags:")]/following-sibling::text()
      Performers:
        Name:
          selector: //li[@class="update_models"]/a
      Studio:
        Name: 
          fixed: Selfie Suck
# Last Updated August 30, 2021```

PR #699, opened 2021-08-31
Update Babepedia.yml (Death date)

Belleyy:
- Adding death date

### DIFF
--- a/scrapers/Babepedia.yml
+++ b/scrapers/Babepedia.yml
@@ -60,12 +60,10 @@ xPathScrapers:
         selector: $label[text()='Died:']]/text()
         postProcess:
           - replace:
-              - regex: st\s|nd\s|rd\s|th\s
+              - regex: (\d+)(st|nd|rd|th)
+                with: "$1"
+              - regex: (\sof\s)|(\(age.+)
                 with: " "
-              - regex: \sof\s
-                with: " "
-              - regex: \(age.+
-                with: ""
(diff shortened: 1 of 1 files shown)


scrapers/Babepedia.yml
```diff
@@ -56,6 +56,17 @@ xPathScrapers:
               - regex: (\w+:)(\s)(\w+)(\s)(\d+)(\w+)(\s)(\w+)(\s)(\w+)(\s)(\d+)
                 with: $5 $10 $12
           - parseDate: 2 January 2006
+      DeathDate:
+        selector: $label[text()='Died:']]/text()
+        postProcess:
+          - replace:
+              - regex: st\s|nd\s|rd\s|th\s
+                with: " "
```
bnkai, line 64: The above replaces the `st` from August eg `https://www.babepedia.com/babe/Victoria_Paris`
I would use the below instead
```yaml
              - regex: (\d+)(st|nd|rd|th)
                with: "$1"
```

scrapers/Babepedia.yml
```diff
@@ -56,6 +56,17 @@ xPathScrapers:
...
+              - regex: \sof\s
+                with: " "
+              - regex: \(age.+
+                with: ""
```
bnkai, line 68: since the replace target is the same  we could use
```yaml
             - regex: (\sof\s)|(\(age.+)
                with: " "
```
instead

PR #703, opened 2021-09-09
Sarj llc improvements

gitgiggety:
These changes contain one actual fix, which is handling of tags on models (line 150 / 158: `{'Name'}` must be `{'Name': t}`). Other changes are related to Python code style, unused imports etc.. Note that to my knowledge no performers actually have tags set on the website/API, so it's not really possible to test this, except testing that everything still works.

### DIFF
--- a/scrapers/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC.py
@@ -1,50 +1,52 @@
 import json
-import os
 import sys
 import re
 import requests
-from urllib.parse import urlparse, urlunparse, urlencode
+from urllib.parse import urlparse, urlencode
+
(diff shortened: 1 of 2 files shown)


bnkai:
Can you also update the last updated date in the yaml file?
We use the date from the yaml file to check for changes ( the relevant plugin at least)

gitgiggety:
> Can you also update the last updated date in the yaml file?
> We use the date from the yaml file to check for changes ( the relevant plugin at least)

Sorry. Obviously didn't open the YAML file so I didn't notice that there was a `Last Updated` in there as well

PR #705, opened 2021-09-11
Create YNGR

JaseNZC:
This was a request.
There was no provided performer but I noticed the name of the performer was mentioned as the first 2 names in the tittle. I pulled the name from there.
Have not seen where there are 2 performers that would break this so for the time being it work.


bnkai:
> This was a request.
> There was no provided performer but I noticed the name of the performer was mentioned as the first 2 names in the tittle. I pulled the name from there.
> Have not seen where there are 2 performers that would break this so for the time being it work.

Yeah for more consistent results i would recommend scraping from the parent bang site as it has performers and a search page.
Everything works though so i guess we can just the site to increase our coverage.

PR #706, opened 2021-09-13
add slayed to vixen scraper

SpedNSFW:
Slayed is Vixen Network's latest site. The existing scraper still works for existing sites in addition to the new one.


Belleyy:
Forget to update the `# Last Updated August 02, 2021`

PR #709, opened 2021-09-14
Update JavLibrary.yml (SearchByName)

Belleyy:
### You need the latest dev version or wait for the 0.10 release.

- Update the mirror list too

Btw should we keep the CDP here ? it's not really useful against cloudflare...


bnkai:
>Btw should we keep the CDP here ? it's not really useful against cloudflare...
I mostly use the python scraper now. If the CDP in this one doesnt work at all feel free to remove it.

bnkai:
@Belleyy i am getting errors 403 and 5xx when  trying to use the scraper. Either the new scene by name functionality or directly from the URL using either the javlibrary site or the mirrors. Does it work for you?

bnkai:
Merging as CF errors i got could be temporary

PR #710, opened 2021-09-14
Update AdultimeAPI (SearchByName)

Belleyy:
### You need the latest dev version or wait for the 0.10 release.

### DIFF
--- a/scrapers/AdultimeAPI.py
+++ b/scrapers/AdultimeAPI.py
@@ -548,6 +548,7 @@ if SEARCH_TITLE:
     SEARCH_TITLE = SEARCH_TITLE.replace("."," ")
     debug("[API] Searching for: {}".format(SEARCH_TITLE))
     api_search = api_search_req("query", SEARCH_TITLE, api_url)
+    final_json = None
     if api_search:
         result_search = []
         for scene in api_search:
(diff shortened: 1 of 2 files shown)


scrapers/AdultimeAPI.py
```diff
@@ -534,6 +542,27 @@ def scraping_json(api_json, url):
 api_url = "https://tsmkfa364q-dsn.algolia.net/1/indexes/*/queries?x-algolia-application-id={}&x-algolia-api-key={}".format(application_id, api_key)
 api_search = None
 api_json = None
+
+
+if SEARCH_TITLE:
+    debug("[API] Searching")
```
bnkai, line 548: I had to modify that a bit as the search doesn't like `.` in my case
```python
    SEARCH_TITLE = SEARCH_TITLE.replace("."," ")
    debug("[API] Searching for: {}".format(SEARCH_TITLE))
```
Do you think it might be better to use the regex part  you used for the `SCENE_TITLE` ?

Belleyy: I don't really like the fact that a regex change the query, so i will take your `replace`.
I think the script should only replace thing that broke the query but nothing more.

PR #722, opened 2021-09-20
SARJ-LLC: add sceneByName support

gitgiggety:


scrapers/SARJ-LLC.py
```diff
@@ -84,12 +84,24 @@ def search(type, name):
...
+                'url': f"https://www.metartnetwork.com{item['path']}",
+            }
+    elif type == 'scene':
+        def map_result(result):
+            item = result['item']
+            return {
+                'title': item['name'],
+                'url': f"https://www.metartnetwork.com{item['path']}",
+                'date': item['publishedAt'][0:item['publishedAt'].find('T')],
+                'performers': list(map(lambda m: {'name': m['name']}, item['models']))
```
bnkai, line 101: can we add the image ?
something like ?
```python
            studio = get_studio(item['siteUUID'])
            if studio:
                image = f"https://www.{studio[1]}{item['thumbnailCoverPath']}"
            return {
                'title': item['name'],
                'url': f"https://www.metartnetwork.com{item['path']}",
                'date': item['publishedAt'][0:item['publishedAt'].find('T')],
                'performers': list(map(lambda m: {'name': m['name']}, item['models'])),
                'image': image
```

gitgiggety: > can we add the image ?

Apparently, we can :)  Didn't know it would be shown in the search results I guess.

> something like ?

Copy / pasted 👍

PR #723, opened 2021-09-22
Add sceneByName support for vipissy scraper

gitgiggety:


bnkai:
I think the issue is that the xpath scraping code was supposed to be for scraping and very simple searching.
Eg 4 titles 4 urls for a search that has 4 results.
Adding the `performers` part that is an array and doesnt have a single value complicates things
In this specific case we would need to use a different selector eg `$scenerow//span//strong/span` so that we have a single title,url,image,date and a single combined performers value for every result. This should work BUT the fact that performers is an array  somehow breaks performer listing.(the same issue might be present in the JSON case also). I am not sure how easy it is to fix this. Script scrapers dont have this issue as they can group the results.

gitgiggety:
Updated to comment out the `Performers` part and added a comment why the code is commented out.

I would have hoped that Stash would magically group them (based on the `$scenerow` part perhaps, but it's just a variable / placeholder for the selector?). But I should have known magic doesn't exist.

PR #725, opened 2021-09-23
Added TwmVRnet.yml

philpw99:
The old twmvrnet entry from TeenMegaWorld.yml doesn't work.
It includes scene scraper and movie scraper.

### DIFF
--- a/scrapers/TmwVRnet.yml
+++ b/scrapers/TmwVRnet.yml
@@ -1,0 +1,60 @@
+name: "TMWVRNET"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - tmwvrnet.com
+    scraper: sceneScraper
+movieByURL:
+  - action: scrapeXPath
+    url:
+      - tmwvrnet.com
+    scraper: movieScraper
+xPathScrapers:
+  sceneScraper:
+    common:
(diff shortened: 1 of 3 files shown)


scrapers/TmwVRnet.yml
```diff
@@ -0,0 +1,60 @@
...
+  sceneScraper:
+    common:
+      $info: &infoSel //div[@class="content__main"]
+      $about: &aboutSel //div[@class="about-video"]
+    scene:
+      Title: &titleSel $about//h1[@class="title"]/text()
+      Date: &dateAttr
+        selector: $info//img[@alt="Release date"]/following-sibling::div/text()
+        postProcess:
+          - parseDate: January 02, 2006
```
bnkai, line 22: It should be
```yaml
          - parseDate: January 2, 2006
```
as the site doesnt use zero padding for the day

bnkai:
Can you also edit the `TeenMegaWorld.yml` file and remove the relevant section( sceneURL action and VRscraper) from there as it doesnt work anymore?

scrapers/TmwVRnet.yml
philpw99: Yup, you are absolutely right. Can't believe I missed this one!

PR #728, opened 2021-09-24
Update VRBangers.yml

philpw99:
Their website just changed. I updated the scraping.
Only VRBangers.yml was changed. The SCRAPERS-LIST.md should be the same.


bnkai:
There is a single `//h1 ` in the page so no need to be more specific.
Merging

PR #729, opened 2021-09-24
Update BaDoink.yml

philpw99:
Fix the image retrieval and add movie scraping.


bnkai:
Unless something changed while reviewing the old image selector works fine while the new on doesn't.
I'll merge the movie part, feel free to make a new PR if the image part changes again

PR #731, opened 2021-09-25
Add SLRoriginals.yml

philpw99:
Includes scene and movie scrapers.

### DIFF
--- a/scrapers/SLRoriginals.yml
+++ b/scrapers/SLRoriginals.yml
@@ -1,0 +1,60 @@
+name: "VRBangers"
+sceneByURL:
+  - action: scrapeXPath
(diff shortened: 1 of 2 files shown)


scrapers/SLRoriginals.yml
```diff
@@ -0,0 +1,60 @@
+name: "VRBangers"
```
bnkai, line 1: Leftover from copy paste? Use the correct name

philpw99: Yeah, sorry. It is a leftover.

philpw99:
The name has changed. Stupid me. 😄

PR #732, opened 2021-09-26
Update VirtualRealPorn.yml and add other virtual real sites

philpw99:
* 5 websites instead of 1.
* Get more tags.
* Can retrieve multiple performers. 
* Added movie scraper.


philpw99:
Thank you @bnkai ! Now this scraper is much better.

PR #734, opened 2021-09-28
Update WankzVR.yml

philpw99:
After more testing, it seems cdns-i.wankzvr.com is a website that can effectively block image scraping. I tried to use drivers: headers and CDP yet unsuccessful. So I have to quote them out, otherwise when saving the scraping result you will have 403 error and the data will not be saved.


philpw99:
Have to say sorry that I didn't listen to bnkai and realize this should be a big problem for this scraper.

PR #738, opened 2021-10-03
Update Mindgeek API (Local + Marker support)

Belleyy:
- Copy official marker. (You will need to have the main tag)
It could lead to incorrect marker because sometimes the time is off. (File with different duration that the official, official marker are just incorrect 🤷‍♂️ ). I have a file that is 20min off than the official 🤔 

- Use local json if you have them and the scene url fail


bnkai:
Merging with minimal testing though as not many scenes seem to have timetags..

PR #745, opened 2021-10-13
Add new SLRoriginals.com and its affilicated studios.

philpw99:
The previous branch had conflicts and could not merge, so I deleted it and created a new branch.
Other than SLRoriginals.com, I added other affiliated studios:
* vrfirsttimer.com
* allvr.porn
* lustreality.com
* hardonvr.com
Also added the ability to create a movie from the scene. No poster and studio though, due to the limit of creating new movies this way. I wish there will be an easy way to put URL in the created movie.

### DIFF
--- a/scrapers/SLRoriginals.yml
+++ b/scrapers/SLRoriginals.yml
@@ -2,11 +2,11 @@ name: "SLROriginals"
 sceneByURL:
   - action: scrapeXPath
     url: &urlAttr
+      - allvr.porn
+      - hardonvr.com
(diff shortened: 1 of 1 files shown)


bnkai:
> lustreality.com already has a scraper. Right now I am just listing both scrapers for that website. Is this OK?

I will have a quick look and respond later
In general though only one scraper should exist for a xxxxByURL action per site, so one of them should have the url removed from the yaml file and the scrapers list

PR #748, opened 2021-10-15
Add automatic scrape retrying to IAFD scraper

7dJx1qP:
The IAFD scrape often fails, returning an HTTP 403 error, and requires multiple attempts until it succeeds. This PR automates the retry process and makes the scrape request again until it succeeds with a max of 10 attempts and a 2 second delay between each one.

In my experience it always works after it retries a few times. It may be the case for some people that no amount of retrying will work, so I'm not sure if this retry behavior would be harmful to them.

### DIFF
--- a/scrapers/IAFD.py
+++ b/scrapers/IAFD.py
@@ -516,7 +516,7 @@ def performer_from_tree(tree):
     if performer_image_url:
         try:
             debug_print("downloading image from %s" % performer_image_url[0] )
-            p.image = scrape_image(performer_image_url[0])
+            p.images = [scrape_image(performer_image_url[0])]
         except Exception as e:
             debug_print("error downloading image %s" %e)
 


Failead0r:
I need it. Thanks for your time.

bnkai:
Looks good
@7dJx1qP instead of sleeping for 2 seconds could we use something `random.randint(1, 4)` ?
( not sure if it helps but if it `confuses` the cf anti bot detection even a little bit  it would be worth considering)

7dJx1qP:
Updated. Not sure if it helps either but it doesn't hurt to do it.

bnkai:
I would also add a 
```python
random.seed()
```
probably before line 608 `tree = scrape(url)` to make sure we get a different starting number everytime the script is run

7dJx1qP:
While testing I noticed that stash is failing to display the performer image that has been scraped. I did some debugging and it looks like the performer image scrape is working and returning a valid b64 encoded string, so I think the error is somewhere in stash and not the scraper. And I think the regression is new to v0.10.0, because I tested v0.9.0 and it's working there. I'll open up an issue in the stash repo.

PR #752, opened 2021-10-27
add shelovesblack

JaseNZC:


bnkai:
Final version above tested and looks ok

@JaseNZC  can you delete the new yaml file and update the LoveHerFeet one and the scrapers list?

JaseNZC:
> Final version above tested and looks ok
> 
> @JaseNZC can you delete the new yaml file and update the LoveHerFeet one and the scrapers list?

Hopefully that sorted it :-)

PR #755, opened 2021-10-31
create ThirdRockEnt

JaseNZC:
Remove AnalizedNetwork and replaced with ThirdRockEnt which is the group these all fall under.
Re-did scraper and added missing sites

I hate multi site scrapers only one site has to change and it screws it all up.

I think I got everything :-)


JaseNZC:
Updated to return single studio result and map studio names.

JaseNZC:
Forgot to fix name. Doh

PR #761, opened 2021-11-04
Traxxx Scrapers

stg-annon:
resolves request for `Traxxx` in #73

Adds the following scrape types for [Traxxx](https://traxxx.me)
- sceneByURL
- sceneByName
- sceneByFragment
- sceneByQueryFragment
- performerByName
- performerByFragment
- performerByURL

This PR also includes a general purpose logger module for use in any other scraper which facilitates logging to the stash console


scrapers/traxxx_interface.py
```diff
@@ -0,0 +1,551 @@
...
+      fragment["ethnicity"] = p.ethnicity
+
+    country = p.birthCountry.alpha2
+    if country:
+      fragment["country"] = country
+
+    fragment["images"] = []
+
+    if p.image:
+      fragment["images"].append( p.image )
```
bnkai, line 343: From a quick test the first image we might need is the avatar one as the some performers dont have profiles
```py
    if p.avatar:
      fragment["images"].append( p.avatar )
    if p.image:
      fragment["images"].append( p.image )
```

scrapers/Traxxx.py
```diff
@@ -0,0 +1,83 @@
...
+    # log.info(json.dumps(data))
+    print(json.dumps(data))
+  else:
+    print("")
+
+# Return a list of scenes from a search
+def scene_name(fragment):
+  title = None
+  if not title:
+    title = fragment.get("title")
```
bnkai, line 44: isnt the above the same as 
```py
def scene_name(fragment):
  title = fragment.get("title")
```
?

scrapers/traxxx_interface.py
```diff
@@ -0,0 +1,551 @@
+import re, sys, copy, json, requests
+
+import py_common.log as log
+from box import Box
```
bnkai, line 4:
```py

from box import Box # pip install python-box
```
add a comment for the module that needs to be installed

scrapers/traxxx_interface.py
stg-annon: Thanks, nice catch, I changed up the wording of the fragments at one point seems I didn't change the avatar image name to image like I did elsewhere so `p.image` never existed resolved it in the fragment

scrapers/Traxxx.py
stg-annon: yeah think I just wanted to be able to re organize the blocks in future if needed without changing any logic, I updated it as it is most likely unnecessary

scrapers/Traxxx.yml
```diff
@@ -0,0 +1,49 @@
+name: "Traxxx"
+sceneByURL:
+  - url:
+     - traxxx.me/actor/
```
bnkai, line 4: it should be `scene` here... not actor
and you havent added  the `scene_url` case in the python file....

bnkai:
A lot of actor details are also missing when you scrape for an actor, even though traxx returns that data

stg-annon:
> A lot of actor details are also missing when you scrape for an actor, even though traxx returns that data Can you also update the scrapers list file to add traxx as a site since you seem to support by URL queries?

Yeah seems after I changed around the parsing I missed the performers

scrapers/Traxxx.yml
stg-annon: nice catch ty

scrapers/Traxxx.py
```diff
@@ -0,0 +1,100 @@
...
+  if not title:
+    title = fragment.get("name")
+  if not title:
+    return
+  return traxxx.search_scenes(title)
+
+# Return a list of scenes from a search
+def scene_by_name(fragment):
+  scenes = search_traxxx_for_scene(fragment)
+  return [traxxx.parse_to_stash_scene_search(s) for s in scenes]
```
bnkai, line 53: A minor nitpick here
If no scenes are found an error is printed
We could leave it like that or do something
```python
  if not scenes:
    return [{}]
  return [traxxx.parse_to_stash_scene_search(s) for s in scenes]
```
or even
```python
  if not scenes:
    return [{"title": "No Scene found"}]
  return [traxxx.parse_to_stash_scene_search(s) for s in scenes]
```
I am not sure what is would be better ....

scrapers/Traxxx.py
```diff
@@ -0,0 +1,100 @@
...
+# return first result from scene_name
+def scene_fragment(fragment):
+  scenes = search_traxxx_for_scene(fragment)
+  if scenes:
+    return traxxx.parse_to_stash_scene(scenes[0])
+
+# Return a list of possible performer matches
+def performer_lookup(fragment):
+  performers = traxxx.search_performers(fragment["name"])
+  return [traxxx.parse_to_stash_performer_search(p) for p in performers]
```
bnkai, line 76: same here ?
```python
  if not performers:
    return [{}]
  return [traxxx.parse_to_stash_performer_search(p) for p in performers]
```

scrapers/Traxxx.py
stg-annon: Ideally it would return whatever stash is expecting as a null result which I believe is one of the following:
```python
print(json.dumps(None))
print(json.dumps([]))
print(json.dumps({}))
```

PR #765, opened 2021-11-09
Add StraponDreamer.yml

Muldec:


scrapers/StraponDreamer.yml
```diff
@@ -0,0 +1,21 @@
+name: StraponDreamer
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - strapondreamer.com/videos
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@id="videoWrapper"]/div[4]/div[2]
```
bnkai, line 10: We generally avoid using indexes if possible, for this case isnt a simple `//h1` selector enough?
```suggestion
      Title: //h1
```

Muldec: You're right, much simpler.

PR #767, opened 2021-11-10
Added gallery support to TheScoreGroup.yml

n0bodyyouknow:
Tested with galleries under all child sites except pornmegaload.com - could not locate gallery on that site, but assuming since all other sites in this scraper have the same layout, it should be compatible there too.


bnkai:
Looks ok. Couldnt find a gallery as mentioned above in pmg (not sure if it has in the free section) but i think it should work...
Merging

PR #770, opened 2021-11-13
Add Kinkbomb.com

NozomiClambie:


scrapers/KinkBomb.yml
```diff
@@ -0,0 +1,48 @@
...
+        postProcess:
+          - replace:
+              - regex: //
+                with: https://
+
+      Details:
+        selector: //div[@class="description-inner"]/p
+        concat: "\n\n"
+      Date:
+        selector: $metainfo/small[2]/text()
```
bnkai, line 27: using a non indexed selector in cases like this seems safer
```suggestion
        selector: $metainfo/small[contains(span,"Posted")]/text()
```

scrapers/KinkBomb.yml
```diff
@@ -0,0 +1,48 @@
...
+            - replace:
+                - regex: .com
+                  with: ""
+      Studio:
+        Name:
+          selector: $details//h4[@class="studio-header-name"]/text()
+          postProcess:
+            - replace:
+                - regex: .com
+                  with: ""
```
bnkai, line 43: You are using the same thing for Studio/Performers, In this case a yaml anchor helps a lot and makes sure you won't update only one of selectors by mistake
Also the `.` is a special regex char so it needs to be escaped
```suggestion
      Performers: &studioAttr
        Name:
          selector: $details//h4[@class="studio-header-name"]/text()
          postProcess:
            - replace:
                - regex: \.com
                  with: ""
      Studio: *studioAttr
```

NozomiClambie:
Thanks for the notes, first time taking a stab at it :)

PR #776, opened 2021-11-17
AVEntertainments xPath scene scraper

bnkai:
As stated in the scraper the search by id is a bit picky. It needs the exact id eg `XXX-01`

### DIFF
--- a/scrapers/AVE.yml
+++ b/scrapers/AVE.yml
@@ -35,7 +35,6 @@ xPathScrapers:
       Title: $searchTitle/a
       URL: $searchTitle/a/@href
       Image: $searchRes/div[@class="single-slider-product__image"]/a/img/@src
- 
   sceneScraper:
     common:
       $singleInfo: //div[@class="single-info"]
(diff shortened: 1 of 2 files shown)


scrapers/AVE.yml
```diff
@@ -0,0 +1,68 @@
...
+      Title: $searchTitle/a
+      URL: $searchTitle/a/@href
+      Image: $searchRes/div[@class="single-slider-product__image"]/a/img/@src
+  sceneScraper:
+    common:
+      $singleInfo: //div[@class="single-info"]
+      $performer: //div[@class="single-info"][contains(span,"Starring")]/span[@class="value"]/a
+    scene:
+      Title:
+        selector: $singleInfo[contains(span,"Item#")]/span[@class="tag-title"]
```
Belleyy, line 44: I think `Item#` should be just `Item`, found a scene with a space so the selector give nothing.

https://www.aventertainments.com/ppv/new_detail.aspx?ProID=13757&languageID=1&VODTypeID=1

bnkai: Nice catch. Yeah didnt notice that. I updated the selector using starts-with and plain `Item`

PR #777, opened 2021-11-18
Added Babes&Stars (babesandstars.com)

Failead0r:
Random website with lot of performer's measurements.


Failead0r:
Forgot performer's details.

PR #782, opened 2021-11-20
Adultempire sceneByName Searching

hackerlicious:
Adds sceneByName searches for AdultEmpire.

### DIFF
--- a/scrapers/AdultEmpire.yml
+++ b/scrapers/AdultEmpire.yml
@@ -25,7 +25,18 @@ sceneByQueryFragment:
 xPathScrapers:
   sceneSearch:
     scene:
-      Title: //div[@id="content"]//div[@class="item-title"]/a/@title
+      Title:
+        selector: //div[@id="content"]//div[@class="item-title"]/a/@href
+        postProcess:
+          - replace:
+            - regex: .+?(\d+)/(.+)
+              with: $2 ($1)
+            - regex: "-"
+              with: " "
+            - regex: "/"
+              with: ""
+            - regex: ".html"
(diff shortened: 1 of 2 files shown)


scrapers/AdultEmpire.yml
```diff
@@ -12,8 +12,40 @@ sceneByURL:
...
+        postProcess:
+          - replace:
+            - regex: .+?(\d+)/(.+)
+              with: $2 ($1)
+            - regex: "-"
+              with: " "
+            - regex: "/"
+              with: ""
+            - regex: ".html"
+              with: ""
```
bnkai, line 39: i have been playing with something like this
```yaml
      Title:
        selector: //div[@class="item-title"]/a/@*[local-name()="href" or local-name()="title"]
        concat: "|"
        postProcess:
          - replace:
            - regex: '/(\d+)/[^|]+\|([^|]+)'
              with: "$2 ($1)"
        split: "|"
```
It seems to work ok and the titles look better imo

bnkai:
There is a known issue with stash and searching. When there multiple duplicate results (images in this case) some are removed and the searches are messed up. You ll notice that if you do a generic search and see that the search results and the images returned dont line up
In this case 
```
  queryURL: https://www.adultdvdempire.com/allsearch/search?q={}
```
uses the search all endpoint that returns results with the same image url for `vod` or dvds` even though the  adultempire `id` is different
Should we use the `dvd`  or `vod` instead of `allsearch` and leave a comment for the user to adjust as needed?
At least till we resolve the stash issue

PR #786, opened 2021-11-22
Update GammaEntertainment to inc. DisruptiveFilms

NozomiClambie:


Belleyy:
You need to update `SCRAPERS-LIST.md` with this new site.

bnkai:
The site uses the new layout from adult time.
Adding the url to the newly merged  `AdultEmpireNew` scraper instead of Gamma is adviced since that gets the performers and tags as well

NozomiClambie:
Probably a better way than how I went about it lol but new request to use `AdultEmpireNew` for DisruptiveFilms.com: https://github.com/stashapp/CommunityScrapers/pull/788

PR #790, opened 2021-11-23
[ThePornDB] Add date to search results at sceneSearch

IAmKontrast:
Add missing date field to scene search results to simplify the selection of the correct scene.

### DIFF
--- a/scrapers/ThePornDB.yml
+++ b/scrapers/ThePornDB.yml
@@ -139,4 +139,4 @@ driver:
       Value: stashjson/1.0.0
     #- Key: Authorization # Uncomment and add a valid API Key after the `Bearer ` part
       #Value: Bearer zUotW1dT5ESmpIpMnccUNczf8q4C9Thzn07ZqygE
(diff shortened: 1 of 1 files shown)


bnkai:
There is an issue with stash when the search returns non unique entries. That's why the title has the porndb id in it to make it unique.
That means that if the search returns the same date more than once you will only have it listed/used once in the results.
I wouldnt expect a results having the same date so that could be ok

Another option would be to adjust the title to include the date there
```yaml
  sceneSearch:
    scene:
      Title:
        # Prevent having duplicated title so it won't get ignored by stash
        selector: data.#.[title,date,_id]
        postProcess:
          - replace:
              - regex: \"
                with: ""
              - regex: \,
                with: "] ["
              - regex: ^\[([^\]]+)]
                with: $1
```

bnkai:
Update:
There is an open PR that will fix the current issue.
Leave the date as you have it, just update the `Last Updated Date`  in the yaml file if possible

PR #796, opened 2021-12-01
FuckedHard18 Scraper

Caleripo:
Adds a FuckedHard18.com scraper as requested in #73.
The scraper is tested and validated. However, this is the first scaper that I add to the repo so if someone could double-check that would be appreciated.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -416,6 +416,7 @@ freeusemilf.com|PaperStreetMedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 freshoutofhighschool.com|NewSensationsNetworkSites.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 frolicme.com|FrolicMe.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fuckedfeet.com|VNAGirls.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
+fuckedhard18.com|FuckedHard18.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fuckfiesta.com|Hustler.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
 fuckingawesome.com|FuckingAwesome.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fuckinghardcore.com|Hustler.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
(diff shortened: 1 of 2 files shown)


SCRAPERS-LIST.md
```diff
@@ -416,6 +416,7 @@ freeusemilf.com|PaperStreetMedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 freshoutofhighschool.com|NewSensationsNetworkSites.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 frolicme.com|FrolicMe.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 fuckedfeet.com|VNAGirls.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
+fuckedhard18.com|FuckedHard18.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
```
Belleyy, line 419: Your scraper only support scene so you should remove the ❌ for Gallery & Movies
`fuckedhard18.com|FuckedHard18.yml|:heavy_check_mark:|:x:|-|-|-|-`

scrapers/FuckedHard18.yml
```diff
@@ -0,0 +1,34 @@
...
+  - action: scrapeXPath
+    url:
+      - fuckedhard18.com/membersarea/gallery.php
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: &name
+        selector: //a[@class="model_category_link"][contains(@href,'sets.php')]/text()
+      Details:
```
Belleyy, line 12: Remove it if it's not used.

scrapers/FuckedHard18.yml
```diff
@@ -0,0 +1,34 @@
...
+      Date:
+        selector: //td[@class="date"]/text()
+        postProcess:
+          - replace:
+            - regex: Added 
+              with: ""
+          - parseDate: 01/02/2006 
+      Image: 
+        selector: //a[@class="jqModal"]/img[@class="thumbs"]/@src[1]
```
bnkai, line 21: This returns all the `idividual clip` images. ( of which stash needs only 1 so gets the  first)
If we only want the first (i also cant find an alternative for a better quality image), to be more exact
```suggestion
        selector: (//a[@class="jqModal"]/img[@class="thumbs"])[1]/@src
```
should be used instead

Caleripo:
Thanks a lot guys! Still a lot to learn for me but trying to contribute to this project!

scrapers/FuckedHard18.yml
Caleripo: Thanks! Yes, unfortunately the data on the website is quite poor also for title and tags. In most cases the thumb generated by stash is better.

PR #797, opened 2021-12-02
Update README.md (Update Stable/Dev)

Belleyy:
I'm not really happy with the term i used... So help me 🥲


bnkai:
@Belleyy i think just leaving a single image now is ok (simpler is better)
New users would have the latest stable anyway

PR #799, opened 2021-12-03
Unset birthdate if it can't be parsed to YYYY-MM-DD

7dJx1qP:
I found an IAFD profile that had a birthdate value of `??/??/2000` which is the value the scraper returned. If a valid YYYY-MM-DD birthdate can't be scraped, then it should not be returned since stash won't let you save invalid date values.

### DIFF
--- a/scrapers/IAFD.py
+++ b/scrapers/IAFD.py
@@ -462,6 +462,12 @@ def performer_from_tree(tree):
             p.birthdate = datetime.datetime.strptime(p.birthdate, iafd_date).strftime(stash_date)
         except:
             p.birthdate = None
+            if performer_birthdate[0].lower() != "no data":
+                dob =  f'D.O.B. : {performer_birthdate[0]}\n'
+                try:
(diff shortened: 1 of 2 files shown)


bnkai:
From what i can think we can either leave it as is so that users can at least keep the year (most users would probaly do something like 2000-1-1) or remove it altogether like you did...
I think in those cases you would simple scrape from another scraper or use google to get the actual date. So removing is probably better
Shall we wait for a little more feedback?

7dJx1qP:
Waiting for more feedback is fine. Users wanting to keep the year portion is a good point. I only noticed this behavior because I had a plugin that was making a scrape gql call and using the data to automatically create performers so it was failing there. But I can just update the plugin to remove the invalid date rather than do it in the scraper.

bnkai:
Feedback from @Belleyy  in the  discord channel
We could leave the format check for the date and if it fails set date to None as it is now AND add the DOB text string to the details part of the performer

PR #802, opened 2021-12-07
add Danni.com

JaseNZC:
Another site to add to the growing list :-)

### DIFF
--- a/scrapers/Danni.yml
+++ b/scrapers/Danni.yml
@@ -1,0 +1,25 @@
+name: Danni
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - danni.com/scenes/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="scene-title"]
+      Performers:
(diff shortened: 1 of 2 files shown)


scrapers/Danni.yml
```diff
@@ -0,0 +1,25 @@
...
+        selector: //div[@class="scene-title"]
+      Performers:
+        Name: //div[@class='danni-card-model-list-container']/a
+      Tags:
+        Name: //div[@class="scene-tags"]/a
+      Image:
+        selector: //script[contains(text(),"let sources")]/text()
+        postProcess:
+          - replace:
+              - regex: .*?(https?:\/\/\S*\.jpg).*
```
bnkai, line 20: just a question about the regex, doesnt really affect anythin
`.*?(https?:\/\/\S*\.jpg).*`
whats the use of `?` after the `*` for? beeing lazy with the ? or greedy without it does it make any difference here?

JaseNZC: Me being overzealous with things. Have pruned it back. :-)

bnkai: Yeah it works either way and you could leave it, i just wasnt sure if it was needed for a special case or something...

PR #804, opened 2021-12-07
Hentaied.com: Various fixes

ryosaeba75:


scrapers/Hentaied.yml
```diff
@@ -7,27 +7,36 @@ sceneByURL:
...
+          postProcess:
+            - replace:
+                - regex: 'Exclude'
+                  with: ''
       Details:
         selector: //div[@class="cont" or @class="column"]/$desc
         concat: "\n\n"
+        postProcess:
+          - replace:
+            - regex: 'Description|\(Read Less\)|\(Read more\)'
```
bnkai, line 35: For the `Description` regex i would use something more restrictive in case it is actually present in the details
eg `^Description\n+`

ryosaeba75: Updated in 07cb7d2

PR #805, opened 2021-12-08
Update Nympho.yml

JaseNZC:
Was a fix from here.
#123 
Is a bit thrown together so feel free to adjust as you see fit.

### DIFF
--- a/scrapers/Nympho.yml
+++ b/scrapers/Nympho.yml
@@ -3,11 +3,11 @@ name: Nympho
 sceneByURL:
   - action: scrapeXPath
     url:
-      - allanal.com/view/
-      - analonly.com/view/
-      - nympho.com/view/
-      - swallowed.com/view/
-      - trueanal.com/scenes/
+      - allanal.com/
+      - analonly.com/
+      - nympho.com/
+      - swallowed.com/
+      - trueanal.com/
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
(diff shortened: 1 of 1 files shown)


JaseNZC:
Changed incase they migrate all sites scenes patch.

bnkai:
For the studio can now use
```yaml
      Studio:
        Name: //a[contains(@class,"navbar-brand")]/img/@title
```
No need to have the whole regex and weird (:-P) map.

JaseNZC:
> For the studio we can now use
> 
> ```yaml
>       Studio:
>         Name: //a[contains(@class,"navbar-brand")]/img/@title
> ```
> 
> No need to have the whole regex and weird (:-P) map.

Updated , even better and more cleaner :-)

scrapers/Nympho.yml
```diff
@@ -20,22 +20,15 @@ xPathScrapers:
...
         postProcess:
           - replace:
               - regex: (\d+)(?:[dhnrst]+)\s+(.+)
                 with: "$1 $2"
           - parseDate: 2 Jan 2006
-      Image: //video[contains(@id,"ypp-player")]/@poster|//div[@id="main-player"]/@data-screenshot
-# Last Updated December 06, 2021
+      Image: //video[contains(@id,"ypp-player")]/@poster|//div[@id="main-player"]/@data-screenshot|//meta[@property="og:image"]/@content
+driver:
+  useCDP: true
```
bnkai, line 33: I didnt notice that before...
Why was cdp added here?

JaseNZC: Have removed it, Think I had it in there as sometimes it would block you. I think it looks more like a cookie thing.

PR #809, opened 2021-12-10
Add Pornbox.com support

ryosaeba75:


ryosaeba75:
I'm not sure if legalporno and pornworld scrappers should be removed as they are listed in pornbox. There was some scenes not listed on legalporno.com that where listed in pornbox.com (under the legalporno studio), that's the main reason for this scrapper. There are others studio (Bang Bros, Private for example) that have dedicated scrappers but are also listed on pornbox.com, but I haven't look at them.

bnkai:
> I'm not sure if legalporno and pornworld scrappers should be removed as they are listed in pornbox. There was some scenes not listed on legalporno.com that where listed in pornbox.com (under the legalporno studio), that's the main reason for this scrapper. There are others studio (Bang Bros, Private for example) that have dedicated scrappers but are also listed on pornbox.com, but I haven't look at them.

We can have multiple sources so that the user can use the url he downloaded the scene from (if he has it). Most users prefer original sources (tube/aggregator sites can have different meta)  so its up to the user to select the scraper he wants.
What we try to avoid is more than one scrapers for the same url (which doesnt apply to the quoted cases) since that isnt supported by stash and can lead to troubleshooting issues.

PR #810, opened 2021-12-11
Add StripzVR.yml

philpw99:
For stripzvr.com. Can create scenes and movies.
* The info is limited due to their web page design.
* Their date format is not consistent. Some is like "1 Apr '21", some is like "1 April '21". And they use a special character: ’ (U+2019).
* Cannot find tags in the webpages.


bnkai:
Their dates are indeed a mess 
eg <https://www.stripzvr.com/kristina-fox/private-lapdance/> -> `Xmas Day` should be regex replaced.
The above scene is btw the older i can access, after that i get to a paywall is that the same for you @philpw99 ?

philpw99:
You are very thorough. I hadn't reach videos that far. The "X Mas" entry is added. Thank you for finding it!
Yup, after viewing like 10 to 20 videos, it will auto-direct you to a paywall. So far I only scraped like 10 videos, so haven't encountered a paywall page in the scraper yet. It's not suitable for mass scraping.

philpw99:
Confirmed. Xmas Day is the last video which will hit a paywall.
Also in my memory is that if I browsed many videos I would also hit a paywall.

PR #812, opened 2021-12-13
Add performer and scene scrapers for ATK Girlfriends

bacchus-enorches:


bacchus-enorches:
Large parts of their site are broken or missing, but these should work for the bits that aren&rsquo;t.

bacchus-enorches:
Took some fragment scraper inspiration from #668 which I did not realize already existed.

bnkai:
> Took some fragment scraper inspiration from #668 which I did not realize already existed.

I'll close the other PR once this is ready so take whatever you need from there

bacchus-enorches:
@bnkai go ahead and merge this if you want; model page scraper needs to be Python to preserve some state so it&rsquo;ll take a minute.

PR #814, opened 2021-12-16
Data18.com has restructured their site

freeagent1384:
Data18 has redone their site structure.  They mention on the page that it's in "Final Bata -  Fully operative but pending some functions and design. Thank you for your patience and understanding."  For the time being this update fixes most issues but the Details could use some help.  I'm collecting across a few nodes and they aren't concating together in the correct order.  Any input would be appreciated.


scrapers/data18.yml
```diff
@@ -2,34 +2,24 @@ name: data18
...
+        selector: //div[b[text()="Story"]]/text() |  //div[b[text()="Story"]]/a/text()
+        concat: ' '
         postProcess:
           - replace:
-              - regex: "^Story:"
-                with:
-              - regex: �
-                with: "'"
+            - regex: "^- "
+              with:
```
bnkai, line 22: For the xpath selectors the order you do the union matters.
In this case it is not needed though.
You could go with 
either
```yaml
      Details:
        selector: //div[b[text()="Story"]]//text()
        concat: ' '
        postProcess:
          - replace:
            - regex: "^Story - "
              with:
```
or even simpler since stash gets the inner text

```yaml
      Details:
        selector: //div[b[text()="Story"]]
        postProcess:
          - replace:
            - regex: "^Story - "
              with:
```
not tested with a lot of scenes so might need some adjustment

scrapers/data18.yml
```diff
@@ -2,34 +2,24 @@ name: data18
...
     url:
-      - data18.com/content
+      - data18.com/scenes
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
     common:
-      $performer: //p[b[text()='Starring:']]/a
-      $studio: //p[b[text()='Site:' or text()='Studio:']]/a
+      $performer: //b[text()='Pornstars / Cast']/../..//a[@class='bold gen']
```
bnkai, line 10: In a few scenes the performer doesnt have a page so we get a span instead of an anchor
```suggestion
      $performer: //b[text()='Pornstars / Cast']/../..//*[local-name()='a' or local-name()= 'span'][@class='bold gen']
```
It shouldnt affect the URL as we will get none in that case

scrapers/data18.yml
```diff
@@ -2,34 +2,24 @@ name: data18
...
+      $studio: (//b[text()="Studio"] | //b[text()="Network"])/../a[last()]
```
bnkai, line 11: from the selector itt seems  we prefer the network name if both are present ?

scrapers/data18.yml
```diff
@@ -2,34 +2,24 @@ name: data18
...
     scene:
-      Title: //h1/text()
+      Title: //span/following-sibling::h1/a/text()
```
bnkai, line 13: are there scenes with more than one `h1` ? (did a test in a few and `//h1/a/text()` seemed enough)

scrapers/data18.yml
```diff
@@ -2,34 +2,24 @@ name: data18
 sceneByURL:
   - action: scrapeXPath
     url:
-      - data18.com/content
+      - data18.com/scenes
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
     common:
```
bnkai, line 9: we can add Movies also
```yaml
    common:
      $movie: //b[text()="Movie:"]/..//a
```
and get the name url  like
```yaml
      Movies:
        Name: $movie/text()
        URL: $movie/@href
```

freeagent1384:
I implemented a bunch of improvements but a for nocked me offline last week. As soon as I'm back online I'll push a new commit. Sorry for delay

scrapers/data18.yml
```diff
@@ -1,35 +1,31 @@
...
-                with:
-              # When day of month is not specified, the scraper defaults to <Month> 1st, <Year>.
-              # Comment the next two lines if you want to skip partial dates instead.
-              - regex: ([A-Za-z]+) (\d{4})
-                with: "$1 1 $2"
-          - parseDate: January 2 2006
+        selector: //time/@datetime
       Details:
-        selector: //p[b[text()='Story:']]
+        selector: //div[b[text()="Story"]] |  //div[b[text()="Story"]]/a | //b[contains(text(),"Movie Description")]/../text()
```
bnkai, line 23: `//div[b[text()="Story"]] |  //div[b[text()="Story"]]/a`

the second part `//div[b[text()="Story"]]/a` is not needed as stash will get all the inner text from the whole div
it even appends the  a (usually performers) at the end of the description due to the concat, so it should be removed

bnkai:
Merging so that we can get some more feedback (looks ok but data18 is too big of a site to check :-) )

PR #815, opened 2021-12-16
multiscrape: pass url to scrape_performer

DefinitelyNotSpicy:
If I don't pass the URL I get an error about fragments not being supported instead of the results.

Example Query for the playground:
```query scrapePerformer($scraper_id: ID!, $performer: ScrapedPerformerInput!) {
  scrapePerformer(scraper_id: $scraper_id, scraped_performer: $performer) {
    name
    url
    gender
    twitter
    instagram
    birthdate
    ethnicity
    country
    eye_color
    height
    measurements
    fake_tits
    career_length
    tattoos
    piercings
    aliases
    images
  }
}
```

Variables: 
```
{
  "scraper_id": "Pornhub",
  "performer": {"name": "Riley Reid", 
   "url": "https://www.pornhub.com/pornstar/riley-reid"}
}
```

If you take the URL out it should fail.

### DIFF
--- a/scrapers/multiscrape.py
+++ b/scrapers/multiscrape.py
@@ -203,7 +203,7 @@ class multiscrape:
                     if spl is not None:
                         for spli in spl:
                             if spli["name"].lower()==name.lower():
-                                r=self.scrape_performer(s, {"name":spli["name"]})
+                                r=self.scrape_performer(s, {"name":spli["name"], "url":spli["url"]})
                                 if r is not None:
                                     scraper_cache[s]=r
                                     found=True
--- a/scrapers/multiscrape.yml
+++ b/scrapers/multiscrape.yml
@@ -14,4 +14,4 @@ performerByName:
       - multiscrape.py
       - query
 
-# Last Updated June 09, 2021
+# Last Updated December 16, 2021


bnkai:
First of this is the PULL REQUEST section you should  have made an issue instead
Also the above query is used when you have the url, what you would use for serch queries 
(provided that the scraper supports searchByName eg the pornhub scraper that was merged today)
is 
```
query {
  scrapePerformerList(scraper_id:  "Pornhub", query: "Riley Reid" ) {
    name
    url
  }
}
```
or the newer
```
query {
  scrapeSinglePerformer(source: {scraper_id:  "Pornhub"}, input: {query: "Riley Reid" }) {
    name
    url
  }
}
````
Getting the url from the above you can then use your query

bnkai:
Having a quick look at the code it seems that the `scrapePerformerList` query is used properly. Are you sure you have latest pornhub scraper that supports scraping?

DefinitelyNotSpicy:
Yes, scrapePerformerlist is used correctly. But the only thing that's passed from that to scrapePerformer is the name. As far as I understand it needs the url, too, which is what this pull request does.

bnkai:
Yeah  i need to have a better look. Are any other scrapers working then ?

DefinitelyNotSpicy:
Other Scrapers like Babepedia work as much as as Pornhub without the fix meaning they get a result from `scrapePerformerList` but not from `scrapePerformer`. I do have the newest Pornhub scraper.

I should have described this in more detail. This function is the culprit:
https://github.com/stashapp/CommunityScrapers/blob/6383fa5b829de79a065c6dee52eb08c5ca9e4b31/scrapers/multiscrape.py#L187
It iterates over the configured fields and scrapers. In line 202 it calls `scrape_performer_list` with the scraper and the name of the performer. Then it iterates over the results which contain multiple performers. It takes the result whose name is equal to the sought performer and calls `scrape_performer` but only passes an object with the name of the result to it.
https://github.com/stashapp/CommunityScrapers/blob/6383fa5b829de79a065c6dee52eb08c5ca9e4b31/scrapers/multiscrape.py#L206
But that's not enough cause the query `scrapePerformer` need the URL too

bnkai:
Yeah it seems you are right. 
Sorry i was looking at some other things while checking this out.

DefinitelyNotSpicy:
I'm glad I could help.

PR #819, opened 2021-12-21
Improve HelixStudios scraper

MrShibe:
Adds sceneByName and performerByName functionality to Helix Studios scraper
Improves description scraping so that some descriptions are not missed
Adds canonicalURL field to scenes and performers

### DIFF
--- a/scrapers/HelixStudios.yml
+++ b/scrapers/HelixStudios.yml
@@ -41,7 +41,7 @@ xPathScrapers:
             - regex: ^
               with: "https://helixstudios.com"
       Image:
-        selector: $searchItem//div[contains(@class, "image-wrapper img-poster")]//img/@src
+        selector: $searchItem//div[contains(@class, "image-wrapper img-poster")]//img/@data-src
         
   sceneScraper:
     common:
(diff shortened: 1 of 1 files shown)


scrapers/HelixStudios.yml
```diff
@@ -2,16 +2,47 @@ name: "HelixStudios"
...
+    scene:
+      Title: $searchThumb//h4/text()
+      URL:
+        selector: $searchItem//div[contains(@class, "grid-item-wrapper")]/a/@href
+        postProcess:
+          - replace:
+            - regex: ^
+              with: "https://helixstudios.com"
+      Image:
+        selector: $searchItem//div[contains(@class, "image-wrapper img-poster")]//img/@src
```
bnkai, line 44: in some cases eg search for `bareback` the image `@src` returns this `data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==` instead of url
switching to `@data-src` seems to work better

MrShibe: Upon testing this, @data-src is definitely more reliable, so I've pushed another commit to incorporate this.

Thanks for the suggestion.

PR #821, opened 2021-12-22
add famedigital.com movies

ferengi82:
im not sure if this is the correct way to do it, but it works ;-)


bnkai:
> im not sure if this is the correct way to do it, but it works ;-)

The place was the correct one.
If possible for future reference, update the `Last Updated Date` in  the scraper yml file and the relevant `M` support column for the SCRAPERS_LIST file

PR #823, opened 2021-12-24
Fixed PornHub fragment scene scraper

jthrow0451:
Fixed PornHub fragment scene scraper not working with files that are JUST the video ID. e.g. "ph613244fcca3f8.mp4"


bnkai:
Yeah looks better, thanks.
For furure reference please also update  the last updated date in the scraper file

PR #824, opened 2021-12-25
Update PureTaboo scrapper

SuperScrapper:


bnkai:
Seems to work ok
For a faster alternative (no cdp+sleep) the Algolia_Adultime api scraper should work with all the above sites.
Merging this till we decide or site changes again :-)

PR #826, opened 2021-12-26
Add new Site: NaughtyAmericaVR.com

ferengi82:
hi,
a little help whit the tiltle-regex would be realy nice (im to dumb).
At the Studioname im not sure if with spaces in the name or without....
Rest seems to work fine.

### DIFF
--- a/scrapers/NaughtyAmericaVR.yml
+++ b/scrapers/NaughtyAmericaVR.yml
@@ -10,11 +10,7 @@ xPathScrapers:
       $sceneinfo: //div[@class="scene-info"]/div/div
     scene:
       Title:
-        selector: //title
-        postProcess:
-          - replace:
-            - regex: ^[^|]+|[ \t]*
-            with:
+        selector: $sceneinfo/div/p[@class="scene-title"]
       Date:
         selector: $sceneinfo/p[@class="scene_date"]/text()
         postProcess:
--- a/scrapers/NaughtyAmericaVR.yml
+++ b/scrapers/NaughtyAmericaVR.yml
@@ -1,0 +1,36 @@
+name: NaughtyAmericaVR
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - naughtyamericavr.com/scene/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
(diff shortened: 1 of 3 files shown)


scrapers/NaughtyAmericaVR.yml
```diff
@@ -0,0 +1,36 @@
...
+  sceneScraper:
+    common:
+      $sceneinfo: //div[@class="scene-info"]/div/div
+    scene:
+      Title:
+        selector: //title
+        postProcess:
+          - replace:
+            - regex: ^[^|]+|[ \t]*
+            with:
```
bnkai, line 17: from a quick look
```suggestion
      Title:
        selector: //title
        postProcess:
          - replace:
              - regex: \s*\|[^|]*$
                with:
```
Is that the title btw?
I see in the main page that  the grey-title is probably used
```yaml
      Title:
        selector: //div[@class="grey-title"]
        postProcess:
          - replace:
              - regex:  \s*VR\sPorn\sVideos\s*$
                with:
```

ferengi82: im not at my pc at the moment, but i will give it a try.
i think ist de //title not grey-title, you can see it when you compare the text with the poster-image.
grey-title is more like a subtitle i think

ferengi82: changed the whole thing. im not 100% happy, but the title alone does not work, it seams its more like an movie title ...
https://www.naughtyamericavr.com/fantasy/2-chicks-same-time-porn-videos?&custom1=30796 (at least 5 times 2 Chicks same time)

bnkai: > changed the whole thing. im not 100% happy, but the title alone does not work, it seams its more like an movie title ... https://www.naughtyamericavr.com/fantasy/2-chicks-same-time-porn-videos?&custom1=30796 (at least 5 times 2 Chicks same time)

Yeah actually choosing a title from that site is not easy.
The latest selector you chose is more like `perfomer + perfomer +  ... in movie_name`
Should we  leave the `grey-title` i mention above commented out in the code as an alternative?

ferengi82: sure, why not.

PR #828, opened 2021-12-26
Add Kristen Bjorn scraper

Jacintocuenca:


scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      Tags:
+        Name: //a[@role="button"]/text()
+      Performers:
+        Name: $performer//h5/text()
+        URL:
+          selector: $performer/@href
+          postProcess: &appendDomain
+            - replace:
+                - regex: ^(.*)$
+                  with: http://kristenbjorn.com$1
```
bnkai, line 35: simpler
```suggestion
                - regex: ^
                  with: http://kristenbjorn.com
```

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      - kristenbjorn.com/web/model/gay-porn-star/
+    scraper: performerScraper
+
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performer: //div[@class="scene-item"]/a[@class="scene-box"]
+    scene:
+      Title: //article/div[1]//h2/text()
+      Details: //div[@class="med-text"]/p/text()
```
bnkai, line 25: There are multiple `text()` in some `<p>` due to some `<span>`s
Doing something like this
```yaml
      Details:
        selector: //div[@class="med-text"]/p/text()
        concat: " "
```

or even better 
```suggestion
      Details: //div[@class="med-text"]/p
``` 
is needed to get all of  the description

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+          postProcess: &appendDomain
+            - replace:
+                - regex: ^(.*)$
+                  with: http://kristenbjorn.com$1
+      Image: &imageAttr //meta[@property="og:image"]/@content
+      Studio:
+        Name:
+          fixed: Kristen Bjorn
+      Movies:
+        Name: //article/div[5]//h2/text()
```
bnkai, line 41: adding a class to the mix makes it safer (less chance of a false match)
```suggestion
        Name: //article/div[@class="StoreHeaderNewContainer"][2]//h2/text()
```

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      - kristenbjorn.com/gay-porn-star/
+      - kristenbjorn.com/web/model/gay-porn-star/
+    scraper: performerScraper
+
+xPathScrapers:
+  sceneScraper:
+    common:
+      $performer: //div[@class="scene-item"]/a[@class="scene-box"]
+    scene:
+      Title: //article/div[1]//h2/text()
```
bnkai, line 24:
```suggestion
      Title: //article/div[@class="StoreHeaderNewContainer"][1]//h2/text()
```

adding a class to the mix makes it safer (less chance of a false match)

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      Duration:
+        selector: //span[text() = "Running Time:"]/following-sibling::text()
+        postProcess:
+          - replace:
+              - regex: \sh(ou)?rs?\s?
+                with: ':'
+              - regex: \smin(ute)?s?
+                with: ':00'
+              - regex: :$
+                with: :00:00
```
bnkai, line 66: this doesnt relay work,  set log to debug and check when you do some scrapes
i tried the below
```yaml
          - replace:
              - regex: ^(\d+)\s*(?i:h)\D+$ # movies with only hours
                with: $1:00:00
              - regex: ^(\d+)\s*(?i:m)\D+$ # movies with only minutes
                with: '$1:00'
              - regex: ^(\d+)\s*(?i:h)\D+(\d+)\s*(?i:m)\D+$ # movies with hours and minutes
                with: '$1:$2:00'
```
and it seems to work for the 3 different cases listed, if  there are more cases we need to check that again

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      Synopsis: //div[@class="more"]/text()
```
bnkai, line 67:
```suggestion
      Synopsis: //div[@class="more"]
```
you need the whole div othewise you only get a partial synopsis

scrapers/KristenBjorn.yml
```diff
@@ -0,0 +1,114 @@
...
+      Name: //article/div[1]/span/text()
+      Date:
+        selector: //span[text()="Release Date:"]/following-sibling::text()
+        postProcess:
+          - replace:
+              - regex: /19([0-9])
+                with: /$1
+              - regex: /20([0-9])
+                with: /$1
+          - parseDate: 01/02/06
```
bnkai, line 56: i have a feeling that the dates will be a pain to get correctly
do you have any samples with a 19xx date ?
i found another case where the date is `1/02/06` so the below regex should  be added as well
```yaml
              - regex: ^(\d/\d\d/\d\d)$
                with: 0$1
          - parseDate: 01/02/06
```

scrapers/KristenBjorn.yml
Jacintocuenca: Thanks! This seems to work too with the movies I was trying.

scrapers/KristenBjorn.yml
Jacintocuenca: I couldn't find samples with 19xx. I put it to be safe, but we can always add it back if I find any of such cases, so I'm removing it.

PR #842, opened 2022-01-04
Add Meninosonline

gimmeliina:


Belleyy:
You need to update `SCRAPERS-LIST.md`

scrapers/Meninosonline.yml
```diff
@@ -0,0 +1,38 @@
...
+  sceneScraper:
+    scene:
+      Title: 
+        selector: //div[@class="col-sm-12"]/h3
+        postProcess:
+          - replace:
+              - regex: .* \(
+                with: ""
+              - regex: \) .*
+                with: ""
```
bnkai, line 17: This is not getting a good title
For older scenes the h3 doesnt have any `()` so you get non consistent titles

scrapers/Meninosonline.yml
```diff
@@ -0,0 +1,38 @@
...
+      Date:
+        selector: //div[@class="col-sm-12"]/h3
+        postProcess:
+          - replace:
+              - regex: .* -.
+                with: ""
+          - parseDate: 2006-01-02
```
bnkai, line 24: the regex used doesnt seem safe and the format used for parseDate is wrong
```suggestion
        postProcess:
          - replace:
              - regex: .*(\d{2}/\d{2}/\d{4})$
                with: $1
          - parseDate: 02/01/2006
```

scrapers/Meninosonline.yml
```diff
@@ -0,0 +1,38 @@
...
+      Studio:
+        Name:
+          fixed: meninosonline
+      Performers:
+        Name: 
+          selector: //h4[@class="d-1"]
+          postProcess:
+            - replace:
+              - regex: "[^a-zA-Z-._~]" # clean filename so that it can construct a valid url
+                with: " " # "%20"
```
bnkai, line 36: regex could remove non Ascii letters that could be present in names better use something more specific eg
```yaml
              - regex: \s*\(.+$
                with:
```

scrapers/Meninosonline.yml
gimmeliina: Date format is language specific. I changed url to en site.

bnkai:
@gimmeliina  i forgot about this one... Do you have a better selector for the title ? (i would imagine you would have scraped more scenes by now)

PR #845, opened 2022-01-04
Fix Helix Studios Title

gimmeliina:


scrapers/HelixStudios.yml
```diff
@@ -47,7 +47,12 @@ xPathScrapers:
     common:
       $info: //div[@class="text"]
     scene:
-      Title: //title/text()
+      Title:
+        selector: //title/text()
+        postProcess:
+          - replace:
+            - regex:  \- Helix Studios.*
+              with: $1
```
bnkai, line 55: `$1` doesnt match anything, that's for matching groups
since you want to clean the match a plain `with:` or `with: ""` should be used instead

PR #847, opened 2022-01-05
Adding an onlyfans scraper using the DIGITALCRIMINAL onlyfans tool

Tweeticoats:
This is a fragment based scraper that uses sqlite databases left behind after running the DIGITALCRIMINAL onlyfans scraper https://github.com/DIGITALCRIMINAL/OnlyFans

This script needs python and requests.

This script connects back to stash to find the current path of the video.
If your instance has a password you need to edit the script to include the api key by removing the # and adding it there.
#   "ApiKey":"xxxxxxxxxx",

Once it has the path it looks for user_data.db and returns the metadata from there.


bnkai:
just from a quick look can you adjust the code to get the config settings from `py_common` ?
`py_common/graphql.py` already deals with the basic configuration so you could use that directly?
(adding a new getScenePath based on the getGalleryPath should be all that is needed)

bnkai:
Looks good (couldnt actually try it)

PR #849, opened 2022-01-06
Fix vixenNetwork scraper

bnkai:
Now uses the graphql endpoint for all sites

### DIFF
--- a/scrapers/vixenNetwork.py
+++ b/scrapers/vixenNetwork.py
@@ -279,14 +279,14 @@ if search_query and "search" in sys.argv:
     if filter:
         log.info(f"Filter: {filter} applied")
 
-    log.debug(f"Query: '{search_query}'")
+    log.debug(f"Query: '{search_query[1:]}'")
 
     for x in studios:
         if filter:
             if x.id.lower() in filter:
                 #log.debug(f"[Filter] {x.id} ignored")
                 continue
-        s = x.getSearchResult(search_query)
+        s = x.getSearchResult(search_query[1:])
         # merge all list into one
         if s:
             lst.extend(s)
--- a/scrapers/vixenNetwork.py
+++ b/scrapers/vixenNetwork.py
@@ -4,147 +4,292 @@ from urllib.parse import urlparse
 
 import requests
 
-import py_common.log as log
+try:
+    import py_common.log as log
+except ModuleNotFoundError:
+    print(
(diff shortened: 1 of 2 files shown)


scrapers/vixenNetwork.py
```diff
@@ -1,126 +1,150 @@
...
+        Site('Tushy'),
+        Site('Tushy Raw'),
+        Site('Slayed'),
+        Site('Vixen')
+}
+
+frag = json.loads(sys.stdin.read())
+url = frag.get("url")
+if url is None:
+  log.error(f"No URL given")
```
Belleyy, line 137: don't need the `f`

Belleyy: Just thinking that the whole block `if url is None:` is useless...
You can't use this scraper if you don't put a URL so this condition will never happens.

bnkai:
scraper update with code by @Belleyy to support scrapeByName

for the sceneByName
 If your search query contains the studio name, the scraper will only search on that site.
`Ariana Vixen Cecilia` will only search on Vixen for `Ariana Cecilia`
 If your search query starts with a '$' all sites will be searched
`'$Ariana Vixen Cecilia` will search in all sites for `Ariana Vixen Cecilia`

scrapers/vixenNetwork.py
```diff
@@ -1,126 +1,298 @@
...
+
+    #  Only search on specific site if the studio name is in the search query
+    # ('Ariana Vixen Cecilia' will search only on Vixen)
+
+    # if the first character is $, filter will be ignored.
+    if search_query[0] != "$":
+        # make sure longer matches are filtered first
+        studios_sorted = sortByLength(studios)
+        for x in studios_sorted:
+            if x.id.lower() not in search_query:
```
Belleyy, line 274:
```suggestion
            if x.id.lower() in search_query:
```

scrapers/vixenNetwork.py
```diff
@@ -1,126 +1,298 @@
...
+        search_query = search_query[1:]
+
+    if filter:
+        log.info(f"Filter: {filter} applied")
+
+    log.debug(f"Query: '{search_query}'")
+
+    for x in studios:
+        if filter:
+            if x.id.lower() in filter:
```
Belleyy, line 289:
```suggestion
            if x.id.lower() not in filter:
```

tolotos68:
Hi, just a question - the scraper doesn't get any tags, right? Is there a chance of using the member's section (if one has an account)?

PR #853, opened 2022-01-10
Update IFeelMyself.py

VinceEatWorld:
I added this to be able to scrape the scene by providing the image url. This is more reliable than the search function and easier to find.

### DIFF
--- a/scrapers/IFeelMyself.py
+++ b/scrapers/IFeelMyself.py
@@ -11,9 +11,9 @@ def readJSONInput():
     return json.loads(input)
 
 def extract_info(table,cover_url=None):
+    description = None
     if table.find(class_= ["blog_wide_new_text","entryBlurb"]):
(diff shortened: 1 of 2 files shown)


scrapers/IFeelMyself.py
```diff
@@ -10,17 +10,20 @@ def readJSONInput():
     input = sys.stdin.read()
     return json.loads(input)
 
-def extract_info(table):
-    description=table.find(class_= ["blog_wide_new_text","entryBlurb"]).get_text(strip=True).replace("\x92","'")
+def extract_info(table,cover_url=None):
+    if table.find(class_= ["blog_wide_new_text","entryBlurb"]):
+        description=table.find(class_= ["blog_wide_new_text","entryBlurb"]).get_text(strip=True).replace("\x92","'")
+    description = None
```
bnkai, line 16: this makes the `if` above irrelevant, shouldnt it go above the `if table.find` instead?

PR #854, opened 2022-01-11
Add WAPdB python scraper

Releck:
Adds a scraper for warashi asian pornstars (WAPdB).

Notes:
- Tested on over 100 female profiles and a dozen male ones.
- French/Japanese return partial data as some data is in raw `<p>` tags which were processed in English.
- I could only find one mini profile that had AV activity listed and it was unknown so it isn't supported.
- Characteristics like piercing and tattoos are checked twice as they are sometimes present only in the "distinctive physical characteristics" field or tags
- Some pages will return the default picture. These pictures use the same URL path as the normal ones and there's no way to tell which ones these are. My guess is these were manually changed.
- Some performers have multiple twitter handles listed. This scraper returns the first one.
- Tested on python 3.10 but should work on 3.8 and up. Requires requests and beautifulsoup.

### DIFF
--- a/scrapers/WAPdB.py
+++ b/scrapers/WAPdB.py
@@ -1,0 +1,222 @@
+import json
+import io
+import sys
+import requests
+from datetime import datetime
+from bs4 import BeautifulSoup
+
+
+def process_name(name):
+    name_map = {
+        'Ô': 'ou',
+        'ô': 'ou',
+        'û': 'uu',
+        'Û': 'uu',
+        'î': 'ii',
+        'Î': 'ii'
+    }
+    for k, v in name_map.items():
+        name = name.replace(k, v)
+    return name.title()
+
+
+def get_gender(url):
+    if 'female-pornstar' in url:
+        return 'female'
+    if 'male-pornstar' in url:
+        return 'male'
+
+
+def scrape_performer(url):
+    resp = requests.get(url)
+    if resp.ok:
+        soup = BeautifulSoup(resp.text, 'html.parser')
+        if soup.find('div', {'id': 'casting-profil-mini-infos'}):
+            return scrape_mini_profile(soup, url)
+        else:
+            return scrape_full_profile(soup, url)
+
+
+def scrape_mini_profile(soup, url):
+    performer = {}
+    birthdate_prefix = 'birthdate: '
+    birthplace_prefix = 'birthplace: '
+    measurements_prefix = 'measurements: '
+    height_prefix = 'height: '
+
+    performer['url'] = url
+    if gender := get_gender(url):
+        performer['gender'] = gender
+
+    if soup.find(text=lambda t: 'pornstar is not yet in our database' in t):
+        print('Performer not in database', file=sys.stderr)
+        return
+
+    if profile := soup.find('div', {'id': 'casting-profil-mini-infos'}):
+        if alphabet_name := profile.find('meta', {'itemprop': 'name'}):
+            name = alphabet_name.attrs['content']
+            performer['name'] = process_name(name)
+        if additional_name := profile.find('meta', {'itemprop': 'additionalName'}):
+            japanese_name = additional_name.attrs['content']
+            performer['aliases'] = japanese_name
+
+    if details_node := soup.find('div', {'id': 'casting-profil-mini-infos-details'}):
+        if birthdate_node := details_node.find('p', text=lambda t: birthdate_prefix in str(t)):
+            birthdate_full = birthdate_node.text.split(birthdate_prefix)[1]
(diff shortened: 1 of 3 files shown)


Belleyy:
```
ERRO[2022-01-14 18:04:07] [Scrape / WAPdB]     performer['measurements'] = measurements_node.text.split(measurements_prefix)[1]
ERRO[2022-01-14 18:04:07] [Scrape / WAPdB]   File "C:\Users\Winter\AppData\Local\Programs\Python\Python39\lib\site-packages\bs4\element.py", line 921, in __getattr__
ERRO[2022-01-14 18:04:07] [Scrape / WAPdB]     raise AttributeError(
ERRO[2022-01-14 18:04:07] [Scrape / WAPdB] AttributeError: 'NavigableString' object has no attribute 'text'
```
So i got this error (Tested on Fukada Eimi & Sora Shiina).
Removing the `.text` no longer create this error. (For `measurements` & `career_length`)

Releck:
Tested both those performers and they both work without any changes. What version of beautifulsoup are you using? I tested with 4.10.0.

Belleyy:
I'm using the v4.9.3 for bs4 and python v3.9.1

bnkai:
4.10 has this 
```
* NavigableString and its subclasses now implement the get_text()
  method, as well as the properties .strings and
  .stripped_strings. These methods will either return the string
  itself, or nothing, so the only reason to use this is when iterating
  over a list of mixed Tag and NavigableString objects. [bug=1904309]
```
so we would need to add a comment to the import requiring version 4.10
or change the code to  work with older versions eg
use  `findAll` instead of `find` and iterate through eache result keeping the first non `NavigableString`  one using `isInstance`

For obvious reasons i would just add a comment for the version

scrapers/WAPdB.py
```diff
@@ -0,0 +1,222 @@
...
+        if image_node := image_container_node.find('img', {'itemprop': 'image'}):
+            image_url = image_node.attrs['src']
+            if '/WAPdB-img/par-defaut/' not in image_url:
+                performer['image'] = f'http://warashi-asian-pornstars.fr{image_url}'
+
+    if len(country_nodes := soup.find_all('span', {'itemprop': 'addressCountry'})) > 1:
+        country = country_nodes[1].text
+        performer['country'] = country
+        if country == 'Japan':
+            performer['ethnicity'] = 'Japanese'
```
bnkai, line 137: for ethnicity we use `asian` 
i think its safe to use it as a fixed value insteas as all the performers in the site are asians ?

scrapers/WAPdB.py
```diff
@@ -0,0 +1,222 @@
...
+        if birthplace_node := details_node.find('p', text=lambda t: birthplace_prefix in str(t)):
+            birthplace_full = birthplace_node.text.split(birthplace_prefix)[1]
+            if ', ' in birthplace_full:
+                birthplace = birthplace_full.split(', ')[0]
+            else:
+                birthplace = birthplace_full
+            if birthplace != 'unknown':
+                performer['country'] = birthplace
+            if birthplace == 'Japan':
+                performer['Ethnicity'] = 'Japanese'
```
bnkai, line 77: same here
for ethnicity we use `asian` 
i think its safe to use it as a fixed value insteas as all the performers in the site are asians ?

bnkai:
the birthplace/country seems to be missing from the full profile?

scrapers/WAPdB.py
Releck: Wasn't sure what the convention was as "asian" where I'm from means people from the Indian subcontinent. There are some American and French performers on the site so it's not safe to make it a static value.

Releck:
> the birthplace/country seems to be missing from the full profile?

Can you give some examples? Some full profiles don't list the country (Satou Haruki) but others that list them (Tachibana Nao) are picked up.

bnkai:
> > the birthplace/country seems to be missing from the full profile?
> 
> Can you give some examples? Some full profiles don't list the country (Satou Haruki) but others that list them (Tachibana Nao) are picked up.
Yeah i'll have to find the performers i tested :-( I will let you know
What happened from what i remember was the country was not updated even though the birth place was available. I'll update when i have the performers.

scrapers/WAPdB.py
bnkai: Yeah the `asian performers database` in the title made me think that...

bnkai:
Update:
I can't seem to find the performer cases mentioned above, everything seems ok
Ok to merge @Releck  ?

Releck:
I've added a check that should make it clearer for users if beautifulsoup is out of date. If you're happy with that feel free to merge @bnkai.

PR #861, opened 2022-01-16
PornstarPlatinum scraper (Xpath)

Belleyy:
Not the best way to get a performer name...

### DIFF
--- a/scrapers/PornstarPlatinum.yml
+++ b/scrapers/PornstarPlatinum.yml
@@ -1,0 +1,36 @@
+name: "PornstarPlatinum"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - pornstarplatinum.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $panel: //div[@class="panel-content"]
+    scene:
+      Title: //div[contains(@class,"video-slider-meta")]/h3/text()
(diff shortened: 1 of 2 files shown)


scrapers/PornstarPlatinum.yml
```diff
@@ -0,0 +1,30 @@
...
+      Title: //div[contains(@class,"video-slider-meta")]/h3/text()
+      Date: $panel/div[@class="video-author"]/span[2]/text()
+      Details: $panel/p/text()
+      Performers:
+        Name: 
+          selector: //div[contains(@class,"video-slider-meta")]/h3/text()
+          postProcess:
+            - replace:
+              - regex: ([a-zA-Z]+\s[a-zA-Z]+)*.+[Ii][Nn]+.+
+                with: "$1"
```
bnkai, line 21: There also cases with `with` instead of in, cases with multiple performers using & and a couple of others i've seen
What about something like this ?
```yaml
          selector: //div[contains(@class,"video-slider-meta")]/h3/text()
          postProcess:
             - replace:
                - regex: ^
                  with: "https://www.pornstarplatinum.com/tour/index.php?search="
                - regex: \s
                  with: "+"
             - subScraper: //div[@id="videos-list"]//span[@class="marker left"]
```
Do a search for the scene title and then get the `black` span
This limits us to one performer per scene but its closer to what the site considers performers i think
The original selector used for this scraper doesnt seem to get more than one performers for the scene anyway?

PR #865, opened 2022-01-20
create Tripforfuck

JaseNZC:
This site was requested.

Seems you can move around in the /members area at will to find scenes you want to scrape.

Well I was able to

### DIFF
--- a/scrapers/TripForFuck.yml
+++ b/scrapers/TripForFuck.yml
@@ -13,6 +13,8 @@ xPathScrapers:
         selector: //div[@class="d-flex small movie-status"]/p
         postProcess:
           - replace:
+              - regex: (\d),(\d)
+                with: $1$2
               - regex: (\d+).+
                 with: $1
           - subtractDays: true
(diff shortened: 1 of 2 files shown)


scrapers/TripForFuck.yml
```diff
@@ -0,0 +1,34 @@
...
+  sceneScraper:
+    scene:
+      Title:
+        selector: //h1[@class="h3 mb-0"]/text()
+      Date:
+        selector: //div[@class="d-flex small movie-status"]/p
+        postProcess:
+          - replace:
+              - regex: (\d+).+
+                with: $1
```
bnkai, line 17: just a minor issue here
for old scenes eg https://www.tripforfuck.com/member/movie/001/index.html where days have more than 3 digits there is a , as a separator

something like 
```yaml
          - replace:
              - regex: (\d),(\d)
                with: $1$2
              - regex: (\d+).+
                with: $1
```
or 
```yaml
          - replace:
              - regex: ","
                with: ""
              - regex: (\d+).+
                with: $1
```
should fix the issue

JaseNZC: Did not notice that, Yup that fix's it for 3 digits.

PR #858, opened 2022-01-15
Add Jellyfin/Emby scraper

masterr2001:
A python scraper that uses the Jellyfin/Emby API to look for Scenes, Performers and Movies via URL, Query or Fragments. Needs the URL, API-Key and User from Jellyfin set in jellyfin.py and the URLs in jellyfin.yml adopted to your local Jelly/Emby Instance

PR #879, opened 2022-01-26
Add Jellyfin/Emby scraper

masterr2001:
As suggested by @bnkai I re-created the Pull Request on the most recent main branch (Original was #858)
A python scraper that uses the Jellyfin/Emby API to look for Scenes, Performers and Movies via URL, Query or Fragments. Needs the URL, API-Key and User from Jellyfin set in jellyfin.py and the URLs in jellyfin.yml adopted to your local Jelly/Emby Instance.

Regarding the question for the URL-Scrapers:
For Scenes and Movies the URL Schema from Jellyfin looks like this: http://[hostname]/Users/[userid]/Items/[itemid]
For Performers the URL Schema looks like this: http://[hostname]/emby/Persons/[Performername]

The reason for implementing the URL-Scrapers was my personal workflow when initially filling my Stash-Instance with content. I have all movies in Jellyfin already (based on nfo-files) including the Performers and Pictures. When I initially added all my directories I scraped all of the Scenes with this Jellyfin Scraper (based on the Name). Over this I got the URL to every scene. I then wrote an Stash-Plugin (which I am able to contribute as well) to mass-create Movies from all the Scenes that have an URL set according to the Jellyfin Schema.
For the Performers I just thought that some people might want to use the Person-Schema from Jelly/Emby to Scrape the Actors rather than searching for them.


bnkai:
Regarding URL scraping I wasnt sure if there was a way to get the the URLs directly from the Jellyfin UI. For example I tried the URL from a performer's page and it errored out.
@masterr2001 any last change or good to merge ?

masterr2001:
Good to merge. Thank you very much bnkai.

PR #884, opened 2022-01-29
Fix performer/bio URL for TheNude scraper

Teda1:


scrapers/TheNude.yml
```diff
@@ -35,13 +36,13 @@ xPathScrapers:
...
               - regex: Caucasian
-                with: "white"
+                with: "Caucasian"
               - regex: Black
-                with: "black"
-              - regex: Latin
-                with: "hispanic"
+                with: "Black"
+              - regex: Latina
+                with: "Latin"
```
bnkai, line 45: The regex replace's where meant to keep the same ethnicity as the first performer scraper we had (freeones)
Some of the changes dont even change anything... I would either leave them as they were or remove all the postProcess

Teda1: No problem. I forget I made other changes to this file locally I didn't mean to push.

scrapers/TheNude.yml
```diff
@@ -23,7 +23,8 @@ xPathScrapers:
   performerScraper:
     performer:
       Name: (//meta[@itemprop="name"])[1]/@content
-      URL: (//meta[@itemprop="url"])[1]/@content
+      URL:
+        selector: //meta[@itemprop="url"]/@content
```
bnkai, line 27: was there an issue with the previous selector?
this now fetches more than one results (you can check the stash log (set to debug) when scraping)

scrapers/TheNude.yml
```diff
@@ -78,7 +79,13 @@ xPathScrapers:
       Piercings: 
         selector: //li/span[@class="list-quest"][contains(text(),'Piercings')]/../text()
       Image:
-        selector: (//meta[@itemprop="image"])[1]/@content
+        selector: //meta[@itemprop="url"]/@content
```
bnkai, line 82: the selector as mentioned above gets 2 results
using the `(//meta[@itemprop="url"]/@content)[1]` instead should get us only the first

scrapers/TheNude.yml
Teda1: On my end, the previous selector was never populating URL with this scrapper. My stash instance runs in a docker container in Unraid. Changing the XPath to your suggestion led the URL to not be provided again. What behavior are you seeing on your end with the old XPath and my change?

scrapers/TheNude.yml
Teda1: This suggestion here also breaks the image selection.

bnkai: damn.... idk why it happens. for me it seems to work but not all of the times...

PR #887, opened 2022-01-31
add Pissplay

JaseNZC:
This update removes bruceandmorgan.com and replaces it with their other site PissPlay.com which contains more meta data and is easier to scrap.

I will leave it to you if you want to delete bruceandmorgan.yml


bnkai:
> This update removes bruceandmorgan.com and replaces it with their other site PissPlay.com which contains more meta data and is easier to scrap.
> 
> I will leave it to you if you want to delete bruceandmorgan.yml


For cases like this we leave the existing scrapers in case some user has urls from those sites. It is then his choice to use the existing URL or look for a new pissplay one with more meta.

JaseNZC:
> > This update removes bruceandmorgan.com and replaces it with their other site PissPlay.com which contains more meta data and is easier to scrap.
> > I will leave it to you if you want to delete bruceandmorgan.yml
> 
> For cases like this we leave the existing scrapers in case some user has urls from those sites. It is then his choice to use the existing URL or look for a new pissplay one with more meta.

Yup no problems sounds like a good idea :-)

scrapers/PissPlay.yml
```diff
@@ -0,0 +1,32 @@
...
+        selector: //div[contains(@class, 'video_date')]
+        postProcess:
+          - parseDate: 2 Jan 2006
+      Performers:
+        Name:
+          fixed: Bruce and Morgan
+      Tags:
+        Name: //div[@id="video_tags"]/a
+      Details:
+        selector: //div[@id="video_description"]/text()
```
bnkai, line 22: are you sure about this?
most scenes i see (<https://pissplay.com/videos/myas-piss-bath>,<https://pissplay.com/videos/black-flag-review>,<https://pissplay.com/videos/playing-with-my-reflection>) have `<p>s` in the div
```yaml
        selector: //div[@id="video_description"]
```
or
```yaml
        selector: //div[@id="video_description"]/p
        concat: "\n\n"
```
seem to work

JaseNZC: God damit, that's my fault I did actually have the one I changed it to same as yours but committed from the wrong tab I had open which was an earlier work in progress. Farkin typical lol

bnkai:
@JaseNZC  can you restore the deleted file from this PR (and the deleted line in the list)?

JaseNZC:
> @JaseNZC can you restore the deleted file from this PR (and the deleted line in the list)?

Hopefully that worked :-)

PR #890, opened 2022-02-01
Add Boobpedia xPath performer scraper

bnkai:
tweaked withoutpant's original scraper from a stash's pull request
requested in #73

- The search functionality is limited by the way the boobpedia api treats white spaces (we cant urlreplace in byName queries in stash so nothing can be done about it atm) and also some results may not be performers. Any search query that  has spaces needs them replaced with `+`. If for example you need to search for `First Last` then use `First+Last`
- boobpedia doesnt seem to have a fixed order for the meta that has to do with metric sizes eg kg/lb, m/feet. I adjusted the regexes to work despite that, let me know if it doesnt work in some cases
- details keeps just the `<p>` part of the descriptions(should be the first paragraph actually), again there are too many variations and i thought that was the safest to use


VonTittyslappen:
thanks @bnkai :)

PR #892, opened 2022-02-03
Add AdultFilmIndex parser

themese8:
Hello, we are building a new adult filmography database on AdultFilmIndex.com. We combine the metadata from multiple adult film sources to make a new, comprehensive adult DVD index and database.

We have made a stash-specific API endpoints for scraping the metadata for movies and scenes. The API is in beta, and we have included a key for accessing it, this key is available for all stash users.

This PR includes the parsers for scenes and movies using AdultFilmIndex API. We have tested the parser locally and have determined it working.

### DIFF
--- a/scrapers/AdultFilmIndex.yml
+++ b/scrapers/AdultFilmIndex.yml
@@ -24,7 +24,7 @@ movieByURL:
     queryURL: https://adultfilmindex.com/api/v1/stash/movie/{url}
     queryURLReplace:
       url:
-        - regex: '.+/movie/(.*)/(.*).+'
+        - regex: '.+/movie/([^/]+)/.+$'
           with: "${1}"
 jsonScrapers:
   sceneSearch:
@@ -78,4 +78,4 @@ driver:
       Value: stashjson/1.0.0
     - Key: Authorization # Beta key, enabled and active for now
       Value: Bearer 4vY0iwSUVPH5cGAX1AUZarJ8pbuDUK53
-# Last Updated February 03, 2022
+# Last Updated February 04, 2022
--- a/scrapers/AdultFilmIndex.yml
+++ b/scrapers/AdultFilmIndex.yml
@@ -30,7 +30,7 @@ jsonScrapers:
   sceneSearch:
     scene:
       Title: data.#.title
-      date: data.#.release_date
+      Date: data.#.release_date
       Image: data.#.thumbnail
       URL: data.#.url
       Details: data.#.description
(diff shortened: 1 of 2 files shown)


scrapers/AdultFilmIndex.yml
```diff
@@ -0,0 +1,75 @@
...
+  sceneScraper:
+    scene:
+      Title: data.title
+      Details: data.description
+      URL: data.url
+      Image: data.thumbnail
+      Date:
+        selector: data.movie.release_date
+        postProcess:
+          - parseDate: 2006-01-02
```
bnkai, line 46: it seems that we need some pp to remove the `T.*` part of the date
```suggestion
          - replace:
              - regex: T.*$
                with: ""
          - parseDate: 2006-01-02
```

scrapers/AdultFilmIndex.yml
```diff
@@ -0,0 +1,75 @@
...
+  movieScraper:
+    movie:
+      Name: data.title
+      Synopsis: data.description
+      URL: data.url
+      Duration: data.runtime
+      Date:
+        selector: data.release_date
+        postProcess:
+          - parseDate: 2006-01-02
```
bnkai, line 64: same as above
```suggestion
          - replace:
              - regex: T.*$
                with: ""
          - parseDate: 2006-01-02
```

themese8:
Thanks for the review, updated the PR

scrapers/AdultFilmIndex.yml
```diff
@@ -0,0 +1,75 @@
...
+  scraper: sceneScraper
+movieByURL:
+  - action: scrapeJson
+    url:
+      - https://adultfilmindex.com/movie/
+    scraper: movieScraper
+    queryURL: https://adultfilmindex.com/api/v1/stash/movie/{url}
+    queryURLReplace:
+      url:
+        - regex: '.+/movie/(.*)/(.*).+'
```
bnkai, line 27: althought it seems to work the regex seems a bit weird, eg
<https://regex101.com/r/LgAWOB/1>
if you only want to keep the movie id i would suggest something like this `.+/movie/([^/]+)/.+$`
instead <https://regex101.com/r/pLE432/1>

themese8: You are right, that's much better regex, updated the PR

MichelFolco:
Cant download and save images anymore I get the error message when pressing save after scrapping:  processing front image: http error 418

PR #893, opened 2022-02-04
Add Domai and GoddessNudes support

DampToast:
This adds support for Domai and GoddessNudes (SARJ sites).

It also fixes an incorrect UUID for Erotic Beauty.


bnkai:
A minor issue, domai,goddesnudes and eroticbeauty now that i see it dont support scenes. Its probably not worth it to change the yaml anchors so i'll just update the scrapers list instead.

PR #895, opened 2022-02-05
Scrapper to copy scene data to gallery

Teda1:
For anyone looking for an easier way to tag their galleries. Here is a scrapper I wrote to copy scene metadata to associated galleries. If no gallery is associated with the scene, setting `find_gallery` to true will enable the scrapper to look for gallery zips found within the scene's directory and associate the gallery to the scene as well as tag the gallery with the scene's data. Do not enable `find_gallery` if scenes don't exist in their own directory.

### DIFF
--- a/scrapers/CopyToGallery.py
+++ b/scrapers/CopyToGallery.py
@@ -5,11 +5,6 @@ import os
 import py_common.graphql as graphql
 import py_common.log as log
 
-SERVER_IP = "http://localhost:9999"
-# API key (Settings > Configuration > Authentication)
-STASH_API = ""
-SERVER_URL = SERVER_IP + "/graphql"
-
 find_gallery = False
 
 def call_graphql(query, variables=None):
--- a/scrapers/CopyToGallery.py
+++ b/scrapers/CopyToGallery.py
@@ -1,7 +1,6 @@
 import json
 import sys
 import os
-import requests
 
 import py_common.graphql as graphql
 import py_common.log as log
--- a/scrapers/CopyToGallery.py
+++ b/scrapers/CopyToGallery.py
@@ -1,0 +1,146 @@
+import json
+import sys
+import os
+import requests
+
+import py_common.graphql as graphql
+import py_common.log as log
+
+SERVER_IP = "http://localhost:9999"
+# API key (Settings > Configuration > Authentication)
+STASH_API = ""
+SERVER_URL = SERVER_IP + "/graphql"
+
+def call_graphql(query, variables=None):
+    headers = {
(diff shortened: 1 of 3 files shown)


scrapers/CopyToGallery.py
```diff
@@ -0,0 +1,151 @@
...
+        elif response.status_code == 401:
+            log.debug("401 returned from callGraphQL")
+            # debug("[ERROR][GraphQL] HTTP Error 401, Unauthorised.")
+            return None
+        else:
+            raise ConnectionError("GraphQL query failed:{} - {}".format(response.status_code, response.content))
+    except Exception as err:
+        log.error(err)
+        # debug(err)
+        return None
```
bnkai, line 46: from a quick look the above can be replaced by
```python
find_gallery = False

def call_graphql(query, variables=None):
    return graphql.callGraphQL(query, variables)
```
Using the imported callGraphQL function makes the code cleaner and already includes the stash configuration from `config.py` in the `py_common` folder

scrapers/CopyToGallery.py
```diff
@@ -0,0 +1,151 @@
...
+                    id
+                    title
+                  }
+                }
+                """
+    variables = {
+        "input": input
+    }
+    result = call_graphql(query, variables)
+    log.debug("updateGallery callGraphQL result " + str(result))
```
bnkai, line 79: As the scraper doesnt print anything by default i think we can switch this to `info` instead of debug so that the user is notified which galleries were updated.
Maybe make the print a little more user friendly also ? eg
```python
    if result:
        g_id = result['galleryUpdate'].get('id')
        g_title = result['galleryUpdate'].get('title')
        log.info(f"updated Gallery ({g_id}): {g_title}")
```

scrapers/CopyToGallery.py
Teda1: Perfect, thanks for bringing this to my attention.

scrapers/CopyToGallery.py
Teda1: No problem.

scrapers/CopyToGallery.py
```diff
@@ -0,0 +1,124 @@
...
+import sys
+import os
+
+import py_common.graphql as graphql
+import py_common.log as log
+
+SERVER_IP = "http://localhost:9999"
+# API key (Settings > Configuration > Authentication)
+STASH_API = ""
+SERVER_URL = SERVER_IP + "/graphql"
```
bnkai, line 11: you forgot to remove this
the `graphql.callGraphQL` function gets those from `py_common/config.py` so they are not needed/used here

PR #896, opened 2022-02-08
Add performer support to CorbinFisher scraper

athleld:
validator passes
tested on a couple pages

### DIFF
--- a/scrapers/CorbinFisher.yml
+++ b/scrapers/CorbinFisher.yml
@@ -4,6 +4,13 @@ sceneByURL:
     url:
       - corbinfisher.com
     scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - corbinfisher.com/tour/models
+    scraper: performerScraper
+
 xPathScrapers:
   sceneScraper:
     scene:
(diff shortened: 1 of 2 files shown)


scrapers/CorbinFisher.yml
```diff
@@ -27,4 +34,23 @@ xPathScrapers:
...
+      EyeColor: //span[contains(text(),"EYES:")]/following-sibling::text()
+      Weight:
+        selector: //span[contains(text(),"WEIGHT:")]/following-sibling::text()
+        postProcess:
+          - lbToKg: true
+      Height:
+        selector: //span[contains(text(),"HEIGHT:")]/following-sibling::text()
+        postProcess:
+          - feetToCm: true
+      Measurements: //span[contains(text(),"COCK SIZE:")]/following-sibling::text()
```
bnkai, line 53: If you dont care about .5 precision you can use the below
```yaml
        selector: //span[contains(text(),"COCK SIZE:")]/following-sibling::text()
        postProcess:
          - replace:
              - regex: ^
                with: "0'"
          - feetToCm: true
```
to have the size in cm
With the above  5 = 5.5 = 5.xx inches -> 13 cm. This is mostly a hack though so not sure if you want it

athleld:
Added the suggestion because that's what a lot of other scrapers seem to do with the same info.

PR #898, opened 2022-02-08
Add spermmania scene scraper

Muldec:
There's not much information on site pages, but the one that are present are scraped


bnkai:
@Muldec is there a way to get a scene link? The main page seems to only have join links

Muldec:
Sure @bnkai. Here's some : 
https://www.spermmania.com/fhg/1331ff2a/movie/152_YamazakiHonoka_AoiShino_RYF6
https://www.spermmania.com/fhg/6a69bd72/movie/147_AlexisCrystal_A7B5

Didn't find a way to specify that pattern in the url between the hostname and the /movie/ so I only left the hostname

PR #899, opened 2022-02-11
Fix issue with missing scene covers in SARJ scraper

DampToast:
The API returned links to covers that didn't exist.  This checks if they exist before passing the image link onwards.

### DIFF
--- a/scrapers/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC.py
@@ -245,9 +245,11 @@ def map_model(base_url, model):
 
 studios = {
         '2163551D11D0439686AD9D291C8DFD71': ('ALS Scan', 'alsscan.com'),
-        '5592E33324211E3FF640800200C93111': ('Erotic Beauty', 'eroticbeauty.com'),
+        'D0E7E33329311E3BB6E0800200C93255': ('Domai', 'domai.com'),
+        'FDA021004E3411DF98790800200C9A66': ('Erotic Beauty', 'eroticbeauty.com'),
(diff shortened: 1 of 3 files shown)


bnkai:
This looks good. Does it make sense to actually download, b64 encode and pass through the actual image since we are already doing the request?

scrapers/SARJ-LLC.py
```diff
@@ -196,7 +196,23 @@ def scrape_movie(base_url, date, name):
...
+            res['Image'] = f"https://www.{studio[1]}{image_part}"
+            try:
+                response = requests.get(res['Image'], headers={
+                    'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0'
+                }, timeout=(3, 6))
+            except requests.exceptions.RequestException as e:
+                log.error(f"Error fetching URL {res['Image']}: {e.strerror}")
+
+            if response.status_code < 400:
+                break
```
bnkai, line 212:
```python
                if response.status_code < 400:
                    res['Image'] = f"data:image/jpeg;base64,{base64.b64encode(response.content).decode()}"
                    response.close()
                    break
```
Something like that seems to work (you will need to import `base64` and for some reason remove the `log.debug(f"Send output: {output}"` line at the end of the scraper)

DampToast: I figured it out then came back here and found it was already done.  I'll put amend the PR shortly

DampToast:
Should be the better way now.  Also just did some auto-formatting

PR #901, opened 2022-02-13
Add tags to PervCity.yml

n0bodyyouknow:
So far I only found tags on scenes from PervCity.com, but theoretically it should work on other sites associated with this scraper if there are any.

### DIFF
--- a/scrapers/PervCity.yml
+++ b/scrapers/PervCity.yml
@@ -52,4 +52,6 @@ xPathScrapers:
                 oraloverdose: OralOverdose
                 pervcity: PervCity
                 upherasshole: UpHerAsshole
-# Last Updated October 14, 2020
+      Tags:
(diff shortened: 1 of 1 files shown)


bnkai:
> So far I only found tags on scenes from PervCity.com, but theoretically it should work on other sites associated with this scraper if there are any.

I could only find scenes with tags in pervcity also. Btw i think we need to update the details also
```yaml
      Details: $sceneinfo//p/text()|$sceneinfo//h3[@class="description"]
```
From a couple of scenes with tags i found
<https://pervcity.com/trailers/Busty-MILF-Syren-De-Mer-Creampied-by-Maestro-Claudio.html>
https://pervcity.com/trailers/Naughty-Brunette-Hime-Marie-Craves-Anal-Sex.html

PR #902, opened 2022-02-14
Add SexMex xPath scene and gallery scraper

Emilo2:
Date scraping is bit iffy, but seems to work pretty well. 

It uses subscraper to find the date from the search results by using the scene title and selecting the first item

### DIFF
--- a/scrapers/SexMex.yml
+++ b/scrapers/SexMex.yml
@@ -27,25 +27,12 @@ xPathScrapers:
           - replace:
               - regex: cover.jpg$
                 with: 1.jpg
-      Date: &date
-        selector: *title
-        postProcess:
-          - replace:
-              - regex: "[^a-zA-Z\\d\\-._~]" # clean title so that it can construct a valid url
-                with: "+" # "%20"
-              - regex: \.+
-                with: "+"
-              - regex: (.*)
-                with: https://sexmex.xxx/tour/search.php?query=$1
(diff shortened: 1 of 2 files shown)


scrapers/SexMex.yml
```diff
@@ -0,0 +1,51 @@
...
+    url:
+      - sexmex.xxx
+    scraper: sceneScraper
+galleryByURL: *byUrl
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: &title //div[@class="panel-body"]/h4//text()
+      Details: &details //meta[@name="description"]/@content
```
bnkai, line 14: This gets only the synopsis  that is displayed in the main scenes page 
The full description seems to be ok with something like this 
```yaml
      Details: &details //div[@class="panel-body"]/p[not(@class)]
```

bnkai:
> Date scraping is bit iffy, but seems to work pretty well.
> 
> It uses subscraper to find the date from the search results by using the scene title and selecting the first item

Date sometimes is wrong due to some weird titles/search results. Idk if its better to have  a wrong date or no date at all.
(No other way to get it though as a site scrape is required for that)

scrapers/SexMex.yml
Emilo2: Yeah that's true. I only looked up scenes with such short descriptions that didn't notice this.

PR #903, opened 2022-02-17
add VirtualPee

JaseNZC:
another one to add.


scrapers/VirtualPee.yml
```diff
@@ -0,0 +1,31 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //h2[@class="video_title"]/text()
+        postProcess:
+          - replace:
+              - regex: "with"
+                with: ""
```
bnkai, line 15: minor nitpick the selector matches the comma as a second result (it is ignored by stash)
i would use something like this ? 
```yaml
        selector: //h2[@class="video_title"]/text()[1]
        postProcess:
          - replace:
              - regex: '\swith.*$'
                with: ""
```
to make sure that everything after " with" is removed from the title

JaseNZC: Sorted :-)

PR #906, opened 2022-02-18
check messages in addition to posts for dc-onlyfans scraper

DefinitelyNotSpicy:
The scraper only used to check the posts not the messages. This also checks the messages table so all files can be found.


bnkai:
@DefinitelyNotSpicy thanks for the contribution. Closing this to merge #921 instead

PR #912, opened 2022-02-20
Scene support for SeeMomSuck

BuffyMilkers:
Some video pages require JavaScript to be enabled for image to load. I don't know how to get the images in those cases. When it is not required, this works.

### DIFF
--- a/scrapers/SeeMomSuck
+++ b/scrapers/SeeMomSuck
@@ -1,0 +1,28 @@
+name: SeeMomSuck
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - seemomsuck.com/videos
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="videoblock"]/h2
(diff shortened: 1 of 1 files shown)


bnkai:
> Some video pages require JavaScript to be enabled for image to load. I don't know how to get the images in those cases. When it is not required, this works.

If js parsing is needed then we usually switch to CDP
That means you would have to add a 
```yaml
driver:
  useCDP: true
```
section at the bottom of your scraper and retest the selectors as parsing may have altered some of them

BuffyMilkers:
> > Some video pages require JavaScript to be enabled for image to load. I don't know how to get the images in those cases. When it is not required, this works.
> 
> If js parsing is needed then we usually switch to CDP That means you would have to add a
> 
> ```yaml
> driver:
>   useCDP: true
> ```
> 
> section at the bottom of your scraper and retest the selectors as parsing may have altered some of them

Thanks for this info, I've go the updated version. Do i just edit the original commit, because I have no clue how to use github properly?

PR #919, opened 2022-02-24
Add JavHub xPath scene scraper

bnkai:
The site seems to have duplicate scenes with different dates
Titles seem to be also performer names for most of the scenes  so for the search i list the duration (in the place of details) so that matching a scene is easier


Belleyy:
It work, some scenes are under login (https://tour.javhub.com/join) so the search look to be working but clicking won't give anything.
Don't know how to fix that 🤷‍♂️

bnkai:
> It work, some scenes are under login (`https://tour.javhub.com/join`) so the search look to be working but clicking won't give anything. Don't know how to fix that man_shrugging

I noticed that but i was a bit lazy to fix that.
I updated the scraper removing all search results that have join links

PR #922, opened 2022-02-25
Add MyCherryCrush XPath Scene Scraper

sevenfifteen:
This is my first time making a scraper let me know if there's anything that can be done better!

### DIFF
--- a/scrapers/MyCherryCrush.yml
+++ b/scrapers/MyCherryCrush.yml
@@ -54,6 +54,6 @@ driver:
   headers:
     - Key: User-Agent
       Value: stash/1.0.0
-    - Key: Authorization
+#   - Key: Authorization
 #     Value: Basic Key
 # Last Updated February 24, 2022
--- a/scrapers/MyCherryCrush.yml
+++ b/scrapers/MyCherryCrush.yml
@@ -1,0 +1,59 @@
+name: MyCherryCrush
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://mycherrycrush.com/
+    scraper: sceneScraper
+
+sceneByName:
+  action: scrapeXPath
+  scraper: sceneSearch
+  queryURL: "https://mycherrycrush.com/members/blog/?s={}"
+
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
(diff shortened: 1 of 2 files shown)


scrapers/MyCherryCrush.yml
```diff
@@ -0,0 +1,59 @@
...
+        postProcess:
+          - parseDate: January 2, 2006
+      Tags: 
+        Name: //span[@class='entry-category']/a
+      Performers:
+        Name:
+          fixed: Cherry Crush
+      Studio:
+        Name:
+          fixed: MyCherryCrush.com
```
Belleyy, line 34: The studio should be `MyCherryCrush`

sevenfifteen: > The studio should be `MyCherryCrush`

Shoot, well in StashDB it is MyCherryCrush.com, should I still change it?

bnkai: With studio aliases available in stash it doesnt really matter as any user can create an alias to match the scraper
The same thing for stashdb matches as the first time it is matched from stashdb it can be linked from there.
@sevenfifteen  any name is fine, if stashdb already has the studio just use that name

bnkai:
Is there a way to test this without being a member? I don't see any `free` scenes.
It looks ok though.
Is only the autnetication header enough? Can you also give an example of how the authenitcation value would be ?
eg 
```yaml
    - Key: Authorization
      Value: Basic XXXXXXX
```
or 
```yaml
    - Key: Authorization
      Value: XXXXXXX
```
?

sevenfifteen:
> Is there a way to test this without being a member? I don't see any `free` scenes. It looks ok though. Is only the autnetication header enough? Can you also give an example of how the authenitcation value would be ? eg
> 
> ```yaml
>     - Key: Authorization
>       Value: Basic XXXXXXX
> ```
> 
> or
> 
> ```yaml
>     - Key: Authorization
>       Value: XXXXXXX
> ```
> 
> ?

Unfortunately, you can't see anything without being a member. But it looks just like the first example you gave, basic is part of it. The site just prompts you for a super basic login. I've scraped about half of my collection with this scraper so far.

This is the login prompt.
image

PR #925, opened 2022-02-26
Add scraper for crashpadseries.com

jackchack:
Ran the validations, all passed


scrapers/CrashpadSeries.yml
```diff
@@ -0,0 +1,33 @@
+name: CrashpadSeries 
+debug:
+  printHTML: true
```
JaseNZC, line 3: Don't think it needs debug still there.

scrapers/CrashpadSeries.yml
```diff
@@ -0,0 +1,33 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://crashpadseries.com
```
bnkai, line 7: If there is not specific reason we usually do it like this
```suggestion
      - crashpadseries.com
```

scrapers/CrashpadSeries.yml
jackchack: woops! removed

scrapers/CrashpadSeries.yml
```diff
@@ -0,0 +1,33 @@
...
+          - replace:
+            - regex: .+(?:"datePublished":")([^T]+).+
+              with: $1
+          - parseDate: 2006-01-02
+      Performers:
+        Name: //div[@id="primary-container"]//div[contains(@class,"model-thumb")]//h4
+      Tags:
+        Name: //div[@id="primary-container"]//a[contains(@class,"content-icon")]/@data-original-title
+      Details:
+        selector: //div[@id="primary-container"]//div[@class="ep-description"]
```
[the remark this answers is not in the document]
jackchack: agree, looks like there is a span of tags I missed, fixed

scrapers/CrashpadSeries.yml
jackchack: fixed

jackchack:
@bnkai @JaseNZC thanks for the comments, I have resolved them

PR #926, opened 2022-02-27
Use cloudscraper for Teamskeet

bnkai:
From discord it seems that the api endpoint used by teamskeet uses now cloudflare
I replaced requests with cloudscraper, it will need some testing though

### DIFF
--- a/scrapers/TeamskeetAPI.py
+++ b/scrapers/TeamskeetAPI.py
@@ -6,10 +6,16 @@ import sys
 from datetime import datetime
 
 try:
-    import requests
+    import py_common.log as log
 except ModuleNotFoundError:
-    print("You need to install the requests module. (https://docs.python-requests.org/en/latest/user/install/)", file=sys.stderr)
-    print("If you have pip (normally installed with python), run this command in a terminal (cmd): pip install requests", file=sys.stderr)
+    print("You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)", file=sys.stderr)
+    sys.exit()
+
+try:
+    import cloudscraper
+except ModuleNotFoundError:
+    print("You need to install the cloudscraper module. (https://pypi.org/project/cloudscraper/)", file=sys.stderr)
+    print("If you have pip (normally installed with python), run this command in a terminal (cmd): pip install cloudscraper", file=sys.stderr)
     sys.exit()
 
 USERFOLDER_PATH = str(pathlib.Path(__file__).parent.parent.absolute())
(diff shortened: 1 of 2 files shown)


scrapers/TeamskeetAPI.py
```diff
@@ -39,63 +33,77 @@ def save_json(api_json, url):
...
+    try:
+        scene_api_json_check = r.json().get('found')
+        if scene_api_json_check:
+            scene_api_json = r.json()['_source']
+        else:
+            log.error('Scene not found (Wrong ID?)')
+            sys.exit(1)
+
+    except:
+        log.debug(f"invalid input {r.content}")
```
Belleyy, line 97:
```suggestion
        if "Please Wait... | Cloudflare" in r.text:
            log.error("Protected by Cloudflare.")
        else:
            log.debug(f"invalid input {r.content}")
```
I am not a fan of printing the whole html into the log.

scrapers/TeamskeetAPI.py
```diff
@@ -39,63 +33,77 @@ def save_json(api_json, url):
...
         'Referer': 'https://www.teamskeet.com/'
     }
-
+    scraper = cloudscraper.create_scraper()
     # Send to the API
     r = ""
     try:
-        r = requests.get(api_url, headers=headers, timeout=(3, 5))
+        r = scraper.get(api_url, headers=headers, timeout=(3, 5))
     except requests.exceptions.RequestException:
```
Belleyy, line 80:
```suggestion
    except cloudscraper.exceptions.CloudflareException:
```
It's not request for sure, but not sure if this one is the correct exception.

scrapers/TeamskeetAPI.py
```diff
@@ -39,63 +33,77 @@ def save_json(api_json, url):
...
-        print("An error has occurred with Requests", file=sys.stderr)
-        print(f"Request status: `{r.status_code}`", file=sys.stderr)
-        print(f"Check your TeamskeetJSON.log for more details", file=sys.stderr)
+        log.error("An error has occurred with Requests")
+        log.error(f"Request status: `{r.status_code}`")
+        log.error(f"Check your TeamskeetJSON.log for more details")
```
Belleyy, line 83:
```suggestion
        log.error("Check your TeamskeetJSON.log for more details")
```

scrapers/TeamskeetAPI.py
bnkai: yikes..yes i only used that to see why the scraper was blocking, i was going to remove that later on.
should we skip the `r.content` print in the else also and leave a simple `log.error("invalid input")` ?

Belleyy: I would keep the `log.error("invalid input")`

scrapers/TeamskeetAPI.py
Belleyy: yea sure

bnkai:
Updated with latest suggestions
I tried to get CF to block me uising the scraper. It took 5 scrapes within one minute to do that and a couple of minutes (without retyring inbetween) to get unblocked.

PR #931, opened 2022-03-02
add Tenshigao.com

JaseNZC:
Requested via Discord.


bnkai:
Just added a URL for the performers since the selector was easy and tweaked the image url postprocessing in case there are full urls

PR #934, opened 2022-03-03
Fix ArchAngelVideo.yml

sevenfifteen:
Fixed the scraper to work correctly on the site. Looks like the site may have changed none of the old links worked.

### DIFF
--- a/scrapers/ArchAngelVideo.yml
+++ b/scrapers/ArchAngelVideo.yml
@@ -12,7 +12,7 @@ xPathScrapers:
       Title: $videoContent//div[@class='title clear']/h2
       Details: //div[@class='description']/p
       Date:
-        selector: $videoContent//div[@class='info']/p/text()[1]
+        selector: $videoContent//div[@class='info']/p/text()[contains(.,"Added:")]
         postProcess:
           - replace:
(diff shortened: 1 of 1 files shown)


scrapers/ArchAngelVideo.yml
```diff
@@ -7,34 +7,29 @@ sceneByURL:
...
+      $videoContent: //div[@class="midwrap clear"]
     scene:
-      Title: $videoContent//h4/a/text()
-      Details: //div[@class="item-info"]/p[@class="description"]
+      Title: $videoContent//div[@class='title clear']/h2
+      Details: //div[@class='description']/p
       Date:
-        selector: $videoContent//ul[@class="item-meta"]//li[1]/text()
-        concat: " "
+        selector: $videoContent//div[@class='info']/p/text()[1]
```
bnkai, line 15: maybe just to be safe use 
```yaml
        selector: $videoContent//div[@class='info']/p/text()[contains(.,"Added:")]
```
instead of the first text ?

sevenfifteen: Yeah, that's a better selector, just changed it.

PR #936, opened 2022-03-06
Adding scraper: upskirtjerk.yml

duck-butter:
Adds a scraper for the site UpskirtJerk.com and also adds a reference to the scraper in SCRAPERS-LIST.md

Fills title, date, studio (as "Upskirt Jerk"), performers, tags, description (details), and cover image just fine

### DIFF
--- a/scrapers/upskirtjerk.yml
+++ b/scrapers/upskirtjerk.yml
@@ -1,0 +1,29 @@
+name: Upskirt Jerk
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - upskirtjerk.com/videoentry/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+     Title:
+        selector: //h1[@class='page_title']/text()
+     Date:
+       selector: //span[@class="day"]/span/text()|//span[@class="day"]/text()
+       concat: "/"
+       postProcess:
+         - replace:
+             - regex: ^(\w+)/(\d+)$ # for dates with current year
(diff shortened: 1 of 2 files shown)


scrapers/upskirtjerk.yml
```diff
@@ -0,0 +1,29 @@
...
+    scene:
+     Title:
+        selector: //h1[@class='page_title']/text()
+     Date:
+       selector: //span[@class="day"]/span/text()|//span[@class="day"]/text()
+       concat: "/"
+       postProcess:
+         - replace:
+             - regex: ^(\w+)/(\d+)$ # for dates with current year
+               with: $1/21/$2 # the year part needs to be updated every year
```
bnkai, line 18:
```sugesstion
        with: $1/22/$2 # the year part needs to be updated every year
```
2022...

scrapers/upskirtjerk.yml
```diff
@@ -0,0 +1,29 @@
...
+         - parseDate: Jan/06/02
+     Performers:
+      Name: //span[@class="meta_modelcategory meta_category"]//a
+     Details: //meta[@property="og:description"]/@content
```
bnkai, line 22: The meta descriptions are partial, not complete
```yaml
     Details: //div[@class="post_excerpt"]/p
```
seems to work, we need to check though if some scenes have multiple `<p>` and need a concat

duck-butter: I've checked a number of scenes, including newer and older scenes. It looks like they always have everything on a single `<p>` for video descriptions

PR #937, opened 2022-03-09
add DollsHub

JaseNZC:
One of these sites was requested, it's made up of a few other sites under the same banner so thought I might as well add the lot of them.

### DIFF
--- a/scrapers/DollsHub.yml
+++ b/scrapers/DollsHub.yml
@@ -18,8 +18,9 @@ xPathScrapers:
         selector: //div[@class="col-sm-12 col-lg-12 col-md-12 scence-title-name"]/h2
       Performers:
         Name: //li/span/a
+        URL:  //li/span/a/@href
       Details:
-        selector: //h4[contains(text(),"Description")]/following-sibling::text()
+        selector: //h4[contains(text(),"Description")]/following-sibling::p
       Image:
         selector: //video[@class="thevideo2"]/@data-poster
       Studio:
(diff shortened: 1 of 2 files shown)


scrapers/DollsHub.yml
```diff
@@ -0,0 +1,41 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="col-sm-12 col-lg-12 col-md-12 scence-title-name"]/h2
+      Performers:
+        Name: //li/span/a
+      Details:
+        selector: //h4[contains(text(),"Description")]/following-sibling::text()
```
bnkai, line 22: The text selector get some more comments as results (not included in stash though)
Getting the `<p>` instead returns only a single result
```suggestion
        selector: //h4[contains(text(),"Description")]/following-sibling::p
```

scrapers/DollsHub.yml
```diff
@@ -0,0 +1,41 @@
...
+      - nastypublicsex.com
+      - oldsfuckdolls.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //div[@class="col-sm-12 col-lg-12 col-md-12 scence-title-name"]/h2
+      Performers:
+        Name: //li/span/a
```
bnkai, line 20: most performers seem to be single named
adding the URL makes it easier to distinguish them later 
```suggestion
        Name: //li/span/a
        URL:  //li/span/a/@href
```

JaseNZC: All sorted , Good point in adding the url.

PR #939, opened 2022-03-11
add fragileslave.com url scraper

xxlbug:
Adds support for urls from fragileslave.com (title, date, img, tags, details).

Since it's my first scraper, please review a little bit more carefully :-)

### DIFF
--- a/scrapers/FragileSlave.yml
+++ b/scrapers/FragileSlave.yml
@@ -29,13 +29,12 @@ xPathScrapers:
               - regex: .*<img.*src0_1x="(.*)".*
                 with: $1
       Studio:
-        Name: 
+        Name:
           fixed: Fragile Slave
       Tags:
-        Name: 
+        Name:
           selector: $details/a/text()
       Performers:
-        Name: 
+        Name:
           selector: //span[@class="tour_update_models"]/a/text()
-
 # Last Updated March 14, 2022
--- a/scrapers/FragileSlave.yml
+++ b/scrapers/FragileSlave.yml
@@ -1,0 +1,46 @@
+name: "Fragile Slave"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - fragileslave.com/
+    scraper: sceneScraper
+
+xPathScrapers:
+  sceneScraper:
+    common:
+      $details: //div[@class="videodetails"]/p[@class="date"]
+    scene:
+      Title:
+        selector: //div[@class="pagetitle"]//h1/text()
+      Details:
+        selector: //div[@class="videocontent"]/p/text()
+      Date:
+        selector: $details[1]/text()
+        postProcess:
+          - replace:
+              - regex: (\s*\|.*)
+                with:
+              - regex: (\d{2})\/(\d{2})\/(\d{4})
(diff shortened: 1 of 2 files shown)


scrapers/FragileSlave.yml
```diff
@@ -0,0 +1,46 @@
...
+        selector: //div[@class="videocontent"]/p/text()
+      Date:
+        selector: $details[1]/text()
+        postProcess:
+          - replace:
+              - regex: (\s*\|.*)
+                with:
+              - regex: (\d{2})\/(\d{2})\/(\d{4})
+                with: $3-$1-$2
+          - parseDate: 2020-02-01
```
bnkai, line 25: The selector matches the commas as well. Adding a `[1]` at the  end gets us the correct text only
The parseDate should be `          - parseDate: 2006-01-02` 
Tbf even the second regex is not needed
```yaml
      Date:
        selector: $details[1]/text()[1]
        postProcess:
          - replace:
              - regex: (\s*\|.*)
                with:
          - parseDate: 01/02/2006
 ```
is enough

File: scrapers/FragileSlave.yml [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+      Image:
+        selector: //div[@class='trailerblock'][1]/script
+        postProcess:
+          - replace:
+              - regex: .*<img.*src0_3x="(.*\.jpg)".*
+                with: $1
```
bnkai, line 31: In some scenes the `script` is not present `https://www.fragileslave.com/updates/trailers/Strapped-down-and-Climaxing.html` so you get no image
The images you get are too big imho, we should get the 1x or 2x by default
The regex needs to be more specific to work for both cases
The below seems to work ok for me
```yaml
      Image:
        # replace src0_1x in the below lines with src0_2x or src0_3x if you need higher res covers
        selector: //div[@class='trailerblock'][1]/script|//img[@class="update_thumb thumbs stdimage"]/@src0_1x
        postProcess:
          - replace:
              - regex: .*<img.*src0_1x="([^"]+\.jpg)".*
                with: $1
```

scrapers/FragileSlave.yml
```diff
@@ -0,0 +1,46 @@
...
+      Studio:
+        Name: 
+          selector: //head/title/text()
+          postProcess:
+            - replace:
+                - regex: (^[\w]+ [\w]+) - .*
+                  with: $1
```
bnkai, line 38: Is there any other studio? If not a plain
```yaml
      Studio:
        Name:
          fixed: Fragile Slave
```
should be better

scrapers/FragileSlave.yml
xxlbug: corrected, thank you for the date suggestion

scrapers/FragileSlave.yml
xxlbug: streamlined it even more and reduced the used image resolution

scrapers/FragileSlave.yml
xxlbug: not that i know of, so lets go with your solution :-) Thank you for the review!

PR #940, opened 2022-03-13
Create PinkVisual.yml

xiosensei:


scrapers/PinkVisual.yml
```diff
@@ -0,0 +1,32 @@
...
+  - action: scrapeXPath
+    url:
+      - www.ipinkvisualpass.com/es/sx/pc/episode
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: 
+        selector: //div[@class="summary"]/h3[@itemprop="name"]/text()
+      Details: //span[@itemprop="description"]/text()
```
bnkai, line 12: This matches details from the `other scenes` as well
The below gets a single match
```suggestion
      Details: //p[@class="description"]/span[@itemprop="description"]/text()
```

scrapers/PinkVisual.yml
```diff
@@ -0,0 +1,32 @@
...
+      Performers:
+        Name: //p[@class="actors"]/span/a/span/text()
+      Image:
+        selector: //div[@class="collage"]/img[@itemprop="image"]/@src
```
bnkai, line 16: same here, we need to be sure that the image we get has the 01.jpg and not just replace if it exists
```yaml
        selector: //div[@class="collage"]/img[@itemprop="image"][contains(@src,"-01.jpg")]/@src
```

PR #952, opened 2022-03-18
Move Sayuncle to Teamskeet API

gimmeliina:


bnkai:
While it works ok i think i am getting more cloudflare blocks due to the API endpoint being protected compared to the original/old scraper. @gimmeliina  does it work good enough for you to replace the old scraper ? If not we could comment out the url in the yml file so that users can uncomment it(and delete the old scraper) if they want to use this scraper.

gimmeliina:
> While it works ok i think i am getting more cloudflare blocks due to the API endpoint being protected compared to the original/old scraper. @gimmeliina does it work good enough for you to replace the old scraper ? If not we could comment out the url in the yml file so that users can uncomment it(and delete the old scraper) if they want to use this scraper.

I haven't got any Cloudfare blocks yet, but lets see...

hurricaneporter:
While working on this, is there a good way to integrate a scene name scraper?

gimmeliina:
I Was wondering if TeamskeetAPI should refactored to PaperStreetMediaAPI to cover all Paper Street Media sites. Now we have several copies of API and regular scrapers.

This PR implements multi site support by simple if else arguments, but it is not practical for greater number of sites.

PR #953, opened 2022-03-19
Fix data18 (studio/date/tags)

bnkai:
Fixes studio and date due to data18 changes.
Tweaks tag selector to get tags from old scenes as well (no `<div>`)


JaseNZC:
Yup that works fine

PR #955, opened 2022-03-21
Add Mature.nl

Emilo2:
Includes scene and gallery scraping.

### DIFF
--- a/scrapers/MatureNL.yml
+++ b/scrapers/MatureNL.yml
@@ -1,0 +1,43 @@
+name: "Mature.nl"
+sceneByURL: &byUrl
+  - action: scrapeXPath
+    url:
+      - mature.nl/
+    scraper: sceneScraper
+galleryByURL: *byUrl
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: &title //div[@class="box"]/h1/text()
+      Details: &details //meta[@name="description"]/@content
+      Tags: &tags
+        Name: //div[@class="box-cnt"]/div[@class="mar-t"]/a[not(@class)]/text()
+      Performers: &performers
+        Name: 
+          selector: //div[@class="name"]/span[@class="col-accent"]/text()
+          postProcess:
+          - replace: 
+            - regex: ( \(EU\))$
+              with: 
+      Image:
+        selector: //span[@id="spnPageUpdateTrailer"]/a/img/@data-src
+      Date: &date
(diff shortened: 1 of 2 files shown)


scrapers/MatureNL.yml
```diff
@@ -0,0 +1,43 @@
...
+    url:
+      - mature.nl/
+    scraper: sceneScraper
+galleryByURL: *byUrl
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: &title //div[@class="box"]/h1/text()
+      Details: &details //meta[@name="description"]/@content
```
bnkai, line 13: It seems that some scenes (only found one atm) have an actual description that should be prefered over the meta
sample `https://www.mature.nl/en/update/14026/spying-on-my-older-and-taller-stepsister`
Based on the above the following selector seems to work ok. I am not sure if the `synopsis` part is there in other scenes though. We would have to find some more with actual descriptions
```yaml
      Details: &details //div/span[text()="Synopsis:"]/following-sibling::text()|//meta[@name="description"]/@content
```

scrapers/MatureNL.yml
```diff
@@ -0,0 +1,43 @@
...
+        Name: 
+          selector: //div[@class="name"]/span[@class="col-accent"]/text()
+          postProcess:
+          - replace: 
+            - regex: ( \(EU\))$
+              with: 
+      Image:
+        selector: //span[@id="spnPageUpdateTrailer"]/a/img/@data-src
+      Date: &date
+        selector: //div[@class="box-cnt"]/div[@class="mar-t"]/text()
```
bnkai, line 26: the date selector matches 2 divs (the second being the view/watch link) and the text a few more commas (you can have a look at the console with log set to debug when scraping)
The below selector seems to get us only the needed text
```yaml
        selector:  //div[@class="box-cnt"]/div[@class="mar-t"][not (contains(a, "the full"))]/text()[1]
```

scrapers/MatureNL.yml
Emilo2: Yeah found scenes with the Synopsis part with [Google](https://www.google.com/search?q=site%3Amature.nl+%22synopsis%3A%22) quite easily. 

Your suggestion seems to be good and works with both kinds of pages. I'll add the change.

scrapers/MatureNL.yml
Emilo2: You suggestion works great

PR #961, opened 2022-03-23
Add WillTileXXX

hphpanon:
I missed updating the last-updated date in #958. If that should be a separate PR to keep merges clean, let me know and I'll remove it from this PR and open a separate one.

As far as the date handling, the HTML element with the date has erroneous data that, as best I can tell, I'd need to regex out in order to pass into `parseDate` so it seemed like it'd make sense to cut out the step of passing it into the parser. If there's benefit to throwing into that postprocessor regardless, please let me know.

### DIFF
--- a/scrapers/WillTileXXX.yml
+++ b/scrapers/WillTileXXX.yml
@@ -1,0 +1,30 @@
+name: WillTileXXX
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.willtilexxx.com/updates
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Studio:
+        Name:
+          fixed: WillTileXXX
+      Title: //span[@class="update_title"]
+      Details: //span[@class="latest_update_description"]
+      Performers:
+        Name:
+          selector: //span[@class="tour_update_models"]/a
+      Date:
+        selector: //span[@class="availdate"]/text()[1]
+        postProcess:
+          - parseDate: 01/02/2006
+      Tags:
+        Name: //span[@class="update_tags"]/a
+      Image:
+        selector: //img[@class[contains(., "large_update_thumb")]]/@src
+        postProcess:
+          - replace:
+              - regex: ^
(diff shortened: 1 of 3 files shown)


bnkai:
> I missed updating the last-updated date in #958. If that should be a separate PR to keep merges clean, let me know and I'll remove it from this PR and open a separate one.

I forgot to check that too :-)
Leave it as is, it shoud be ok.

> 
> As far as the date handling, the HTML element with the date has erroneous data that, as best I can tell, I'd need to regex out in order to pass into `parseDate` so it seemed like it'd make sense to cut out the step of passing it into the parser. If there's benefit to throwing into that postprocessor regardless, please let me know.

In general parsedate is still useful as it checks for out of range dates that can pass through the regex
For this case changing the selector a bit is enough to skip the regex-replace. More below...

scrapers/WillTileXXX.yml
```diff
@@ -0,0 +1,31 @@
...
+      Details: //span[@class="latest_update_description"]
+      Performers:
+        Name:
+          selector: //span[@class="tour_update_models"]/a
+      Date:
+        selector: //span[@class="availdate"]
+        postProcess:
+          - replace:
+              - regex: (\d{2})\/(\d{2})\/(\d{4})(.*)
+                with: $3-$1-$2
```
bnkai, line 22:
```yaml
      Date:
        selector: //span[@class="availdate"]/text()[1]
        postProcess:
          - parseDate: 01/02/2006
```
Something like this is enough

scrapers/WillTileXXX.yml
```diff
@@ -0,0 +1,31 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - https://www.willtilexxx.com/updates
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Studio:
+        Name: WillTileXXX
```
bnkai, line 11: you forgot the `fixed`
```yaml
      Studio:
        Name:
          fixed: WillTileXXX
```

PR #793, opened 2021-11-29
Algolia Scraper (Adultime, filthykings, zerotolerancefilms, ...)

Belleyy:
- To create a scraper with it, write the name of the domain after the `Algolia.py`
```yml
    script:
      - python
      - Algolia.py
      - SITENAME
```
- URL can be incorrect. I think this issue will be for Adultime, other should work correctly.
- Studio can be inexact. Again mostly for Adultime.

- Change the config to a `.ini`. You can remove the `Adultime_key.txt` if you used `AdultimeAPI.py`.
- Added some more **py_common**. (GraphQL interface & Config)
- Edited the `log.py` and using it.

- Add support for (Conflict with #792 probably): 
  - disruptivefilms
  - filthykings
  - mypervyfamily
  - touchmywife
  - zerotolerancefilms

- Removed `IGNORE_TAGS` var since Stash do it.

PR #962, opened 2022-03-23
Add option to scrape all performers in Algolia API scenes, not just females

hphpanon:
I couldn't find an explanation in https://github.com/stashapp/CommunityScrapers/pull/793 nor the git history of Algolia.py as to why this is behaving this way. On the few PureTaboo pages I tested against, it seemed capable of retrieving the male performers just fine.

As far as updating dates, should I update all of the `Algolia_*.yml` configs to reflect this change, since they all depend on this file? Alternatively/additionally, it appears [the Scraper Checker Plugin](https://github.com/stashapp/CommunityScripts/blob/main/plugins/GHScraper_Checker/GHScraper_Checker.py) will only fetch .yml files, so if the main reason we'd want to update the dates is for that script, changing those dates would seemingly just cause unnecessary fetches.

### DIFF
--- a/scrapers/Algolia.py
+++ b/scrapers/Algolia.py
@@ -385,11 +385,10 @@ def scraping_json(api_json, url=None):
     # Performer
     perf = []
     for x in api_json.get('actors'):
-        if x.get('gender') == "female":
-            perf.append({
-                "name": x.get('name').strip(),
-                "gender": x.get('gender')
-            })
+        perf.append({
+            "name": x.get('name').strip(),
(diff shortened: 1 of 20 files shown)


Belleyy:
I normally put the a option to only get female performer, probably forget it 😕 
And yes you should update every algolia yml.

bnkai:
I think most users prefer just female (thats probably why we didnt notice that till now) , we should add a config option like we have in the other scrapers though with all as a default (since we can now add meta to stashdb)

PR #963, opened 2022-03-25
Update tag and performer parsing for VNAGirls site

hphpanon:
Spot-checked a few linked sites to verify they have seemingly migrated to a `tag1,tag2` format vs `tag1, tag2`.

Also, something I spent a lot of time debugging but ultimately couldn't solve is the Performer Name parsing. It appears common/normal for these sites to have the performer names formatted as `First1 Last1,&nbsp;First2 Last2`, which, with the split set to `","`, causes them to get parsed as:

* `First1 Last1`
* `&nbsp;First2 Last2`

Nothing I tried (`normalize-space`, trying to use regex to rip out `nbsp;`s) worked. Is there something I'm missing, or is the inconsistent formatting of the site too much for the relatively strict XPath (and/or Stash's use of it) to play nice with?

### DIFF
--- a/scrapers/VNAGirls.yml
+++ b/scrapers/VNAGirls.yml
@@ -45,13 +45,13 @@ xPathScrapers:
       Performers:
         Name:
           selector: //div[@class='customcontent']/h3/text()
-          split: ","
-          postProcess:
+          postProcess: &pp
             - replace:
                 - regex: \x{0020}|\x{00A0} # unicode SP, NBSP
                   with: " "
                 - regex: '\s*,\s*'
                   with: ","
+          split: ","
       Title:
         selector: //div[@class='customcontent']/h1/text()
       Details:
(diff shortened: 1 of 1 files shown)


JaseNZC:
Yeah their site's are very inconsistent and you can't cover them all off. 
You would have to split out the ones that don't conform.
The majority work ok and if somebody really wanted to they could just manually add the performers name in.

bnkai:
A more generic that should work with most cases would be 
```yaml
          postProcess:
            - replace:
                - regex: \x{0020}|\x{00A0} # unicode SP, NBSP
                  with: " "
                - regex: '\s*,\s*'
                  with: ","
          split: ","
``` 
I found a scene in random that has both the unicode SP character and the unicode NBSP in the PERFORMERS section though
`https://girlgirlmania.com/videos/34619/aubrey-gold-nina-elle-nicky-ferrarilesbian-pussy-eating-milfs`
The above should work for both of those cases as well as cases  with plain space

@hphpanon do you have some sample urls with the tags you mention ?
I think we can use the above both in performers and tags, not sure if the NBSP character you mention is the same

hphpanon:
@bnkai Thanks for the regex. Verified it worked correctly on the romemajor URL I was testing against (`https://www.romemajor.com/videos/32839/rome-major-dee-williams-interracial-hardcore-bg-bbc-milf-pawg`) to properly parse out the performers without excess whitespace as well as continue to parse the no-whitespace tags from that page. Tried it against the URL you provided too and that also worked (tags and performers both parsing properly). I had the `'\s*,\s*'` in what I was testing but didn't have the `\x{0020}|\x{00A0}` one.

bnkai:
Yeah i think just replacing those two characters with a normal space is the safest to do atm. IDK if they use some other special character. If we find something else we can add it later on.

PR #970, opened 2022-04-04
Add xSinsVR scraper

Emilo2:
Combined scene and gallery scrapers for xsinsvr.com and billiestar.com

### DIFF
--- a/scrapers/xSinsVR.yml
+++ b/scrapers/xSinsVR.yml
@@ -2,8 +2,8 @@ name: "xSinsVR"
 sceneByURL: &byUrl
   - action: scrapeXPath
     url:
-      - xsinsvr.com/video
       - billiestar.com/video
+      - xsinsvr.com/video
     scraper: sceneScraper
 galleryByURL: *byUrl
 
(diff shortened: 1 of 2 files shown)


scrapers/xSinsVR.yml
```diff
@@ -0,0 +1,37 @@
...
+  sceneScraper:
+    scene:
+      Title: &title //div[contains(@class,"video-player-container__info")]/div/h1
+      Date: &date
+        selector: //time/text()
+        postProcess:
+          - parseDate: Jan 02, 2006
+      Details: &details
+        selector: //li/div[@class="small"]/p/text()
+        concat: " "
```
bnkai, line 20: The text can be inside the p or a span in the p or an anchor among others
eg
`https://xsinsvr.com/video/leveling-up`
`https://billiestar.com/video/double-anal-2`

Using either 
```yaml
        selector: //li/div[@class="small"]
```
or 
```yaml
        selector: //li/div[@class="small"]/p//text()
        concat: " "
```
should work

Emilo2: Found even better solution

```yaml
        selector: //li/div[@class="small"]//p
        concat: "\n\n"
```

This will properly separate the paragraphs and include all the texts in the <p> tags and their children.
Works on your example links.

PR #974, opened 2022-04-13
Adding a performer scraper for helloladyboy

kozobot:


scrapers/VegasDreamsLLC.yml
```diff
@@ -34,5 +39,26 @@ xPathScrapers:
...
+      Height:
+        selector: //*/span[@class="model-details" and text()="Height:"]/../text()
+        postProcess:
+          - feetToCm: true
+      Weight:
+        selector: //*/span[@class="model-details" and text()="Weight:"]/../text()
+        postProcess:
+          - replace:
+              - regex: .*([0-9]{2,3}).*
+                with: $1
```
bnkai, line 55: regex doesn't work <https://regex101.com/r/9Y7Emp/1>
why not use something simpler like 
```yaml
              - regex: ([^\d]+)
                with: ""
```
to clean the weight?

bnkai:
Can you update the [scrapers list](https://github.com/stashapp/CommunityScrapers/blob/master/SCRAPERS-LIST.md) to add a P tick to the helloladyboy.com url?

PR #977, opened 2022-04-20
add grooby.club scraper

jessqic:


scrapers/GroobyClub.yml
```diff
@@ -0,0 +1,43 @@
...
+      Performers:
+        Name: //div[@class="trailer_videoinfo"]//b[contains(.,"Featuring")]/following-sibling::a/text()[1]
+      Studio:
+        Name: //meta[@name="author"]/@content
+      Image:
+        selector: //div[@class="videohere"]/img[@class="thumbs stdimage"]/@src
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https://grooby.club/"
```
jessqic, line 42: I would like to use the supplied site instead but I'm not sure how to do it in stash scraper syntax

bnkai: The way to do it is using the xpath `union` (`|`) to get  a selector that includes the full url and a selector with the image.
Then using stash's `concat` and a regex replace construct a full image url 
For example (not tested on all sites) using your image selector
```yaml
      Image:
        selector: //base/@href|//div[@class="videohere"]/img[@class="thumbs stdimage"]/@src
        concat: "/"
        postProcess:
          - replace:
              - regex: (//tour/)
                with: ""
```

- the `/base/@href`  seems to have the base url with the `tour/` included  (not checked against all sites)
- using the union and concating we get a full url with an extra `/tour//` part that we regex replace
- the image selector isn't valid for the first site `asianamericantgirls` (needs to be checked for other sites as  well). If a selector that isnt valid for every site is available then you can just add more than one(as long as only one is found per site so that the constructed url is ok) using the union mentioned above

scrapers/GroobyClub.yml
```diff
@@ -0,0 +1,43 @@
...
+# Last Updated Apr 20, 2022
```
bnkai, line 43: It  should be `April` not `Apr`

scrapers/GroobyClub.yml
```diff
@@ -0,0 +1,43 @@
...
+    scene:
+      Title: //div[@class="trailer_videoinfo"]//h3/text()
+      Date:
+        selector: //div[@class="trailer_videoinfo"]//b[contains(.,"Added")]/following-sibling::text()[1]
+        postProcess:
+          - replace:
+              - regex: ^-
+                with: ""
+          - parseDate: Jan 2, 2006
+      Details: //div[@class="trailer_videoinfo"]/p[4]/text()
```
bnkai, line 32: Does not work for all sites as
- not all sites have 4 paragraphs
- some sites have extra `<l>`'s before the `<p>`'s
```yaml
      Details: //div[@class="trailer_videoinfo"]//p[not(b)]
```
may work better, not tested on all sites though

jessqic:
updated:
- fixed description selector
- support jwplayer thumbnails too

PR #978, opened 2022-04-21
[dc-onlyfans] use log module for logging

stg-annon:


scrapers/dc-onlyfans.py
```diff
@@ -9,7 +9,7 @@
     import py_common.graphql as graphql
     import py_common.log as log
 except ModuleNotFoundError:
-    print("You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)", file=sys.stderr)
+    log.error("You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)")
```
bnkai, line 12: If the module is not found then no `log` will be available.
The `print` needs to stay here

bnkai:
If possible please update the .yml file with the latest updated date as well

scrapers/dc-onlyfans.py
stg-annon: ahh yeah, fair enough

PR #981, opened 2022-04-27
Reality lovers

kozobot:
Adding a scraper for RealityLovers

### DIFF
--- a/scrapers/RealityLovers.py
+++ b/scrapers/RealityLovers.py
@@ -1,0 +1,107 @@
+import json
+import sys
+import re
+import requests
+import time
+import base64
+
+# Set these cookies to the values from your RealityLovers browser session
+cookies = []
+
+# create our session with cookies
+session = requests.session()
+for cookie in cookies:
+    session.cookies.set_cookie(cookie)
+
+try:
+    import py_common.log as log
+except ModuleNotFoundError:
+    print(
+        "You need to download the folder 'py_common' from the community repo (CommunityScrapers/tree/master/scrapers/py_common)",
+        file=sys.stderr)
+    sys.exit(1)
+
+try:
+    from lxml import html
+except ModuleNotFoundError:
+    print("You need to install the lxml module. (https://lxml.de/installation.html#installation)", file=sys.stderr)
+    print("If you have pip (normally installed with python), run this command in a terminal (cmd): pip install lxml",
+          file=sys.stderr)
+    sys.exit()
+
+
+#  --------------------------------------------
+
+# This is a scraper for: RealityLovers sites
+#
+# These fields will be populated if available:
+# Title, URL, Image, Date
+#
+
+
+# Get the scene by the fragment.  The title is used as the search field.  Should return the JSON response.
+# This is a script instead of a JSON scraper since the search is a POST
+def sceneByName():
+    # read the input.  A title or name must be passed in
+    inp = json.loads(sys.stdin.read())
+    log.trace("Input: " + json.dumps(inp))
+    query_value = inp['title'] if 'title' in inp else inp['name']
+    if not query_value:
+        log.error('No title or name Entered')
+    log.trace("Query Value: " + query_value)
+
+    # call the query url based on the input and validate the response code
+    data = {
+        "sortBy": "MOST_RELEVANT",
+        "searchQuery": query_value,
+        "offset": 0,
+        "isInitialLoad": True,
+        "videoView": "MEDIUM",
+        "device": "DESKTOP"
+    }
+    scraped_scenes = session.post(f"https://realitylovers.com/videos/search?tc={time.time()}",
+                                  data=data)
+    log.debug("Called: " + scraped_scenes.url + " with body: " + json.dumps(data))
+    if scraped_scenes.status_code >= 400:
+        log.error('HTTP Error: %s' % scraped_scenes.status_code)
+
+    # get the data we can scrape directly from the page
+    scenes = json.loads(scraped_scenes.content)
+    log.debug("Scenes: " + json.dumps(scenes))
+    results = []
+    for scene in scenes['contents']:
+        # Parse the date published.  Get rid of the 'st' (like in 1st) via a regex. ex: "Sep 27th 2018"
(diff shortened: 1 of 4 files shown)


scrapers/RealityLovers.py
```diff
@@ -0,0 +1,195 @@
...
+    # get the url specified in the input and validate the response code
+    scraped = session.get(inp['url'])
+    if scraped.status_code >= 400:
+        log.error('HTTP Error: %s' % scraped.status_code)
+    log.trace('Scraped the url: ' + inp["url"])
+
+    # get the data we can scrape directly from the page
+    tree = html.fromstring(scraped.content)
+    title = "".join(tree.xpath('//*[@class="video-detail-name"]/text()')).strip()
+    details = re.sub(r'\n\s*\n', '\n\n', "\n\n".join(tree.xpath('//*[@itemprop="description"]//text()')).strip())
```
bnkai, line 104: The details get a few extra tabs as well as a Read More and `...`
It seems we only need to get the first text from the p and the first text from the span
With limited testing the below seems to work
```python
    details = tree.xpath('//*[@itemprop="description"]/text()[1]')[0].strip()
    details = details + " " + tree.xpath('//*[@itemprop="description"]/span/text()')[0].strip()
```

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,195 @@
...
+    # create our output
+    return {
+        'title': title,
+        'tags': [{
+            'name': x
+        } for x in tags],
+        'details': details,
+        'image': image_url,
+        'studio': {
+            'name': "Reality Lovers"
```
bnkai, line 121: you are using the scraper for tsvirtuallovers as well

getting the studio name as
```python
    studio = tree.xpath('//span[@class="icon header-logo"]/text()')[0]
```
and using it here works for both sites

kozobot: Good find on the header logo text.

scrapers/RealityLovers.py
kozobot: Updated with your suggestion

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,197 @@
+import json
+import sys
+import re
+import requests
+import time
+import base64
+
+# Set these cookies to the values from your RealityLovers browser session
```
bnkai, line 8: Is setting the cookies from browser actually needed ? I left it as it is and it seems to work ok?

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,197 @@
...
+def sceneByURL():
+    # read the input.  A URL must be passed in for the sceneByURL call
+    inp = json.loads(sys.stdin.read())
+    if not inp['url']:
+        log.error('No URL Entered')
+
+    # get the url specified in the input and validate the response code
+    scraped = session.get(inp['url'])
+    if scraped.status_code >= 400:
+        log.error('HTTP Error: %s' % scraped.status_code)
```
bnkai, line 98:
```python
    if scraped.status_code >= 400:
        log.error('HTTP Error: %s' % scraped.status_code)
        return {}
```
If we get an http error we should return an empty scene as all the below code will fail

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,197 @@
...
+def performerByURL():
+    # read the input.  A URL must be passed in for the sceneByURL call
+    inp = json.loads(sys.stdin.read())
+    if not inp['url']:
+        log.error('No URL Entered')
+
+    # get the url specified in the input and validate the response code
+    scraped = session.get(inp['url'])
+    if scraped.status_code >= 400:
+        log.error('HTTP Error: %s' % scraped.status_code)
```
bnkai, line 47:
```python
    if scraped.status_code >= 400:
        log.error('HTTP Error: %s' % scraped.status_code)
        return {}
```
same as the scene part, http error means below code will fail

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,197 @@
...
+
+# Get the scene by the fragment.  The title is used as the search field.  Should return the JSON response.
+# This is a script instead of a JSON scraper since the search is a POST
+def sceneByName():
+    # read the input.  A title or name must be passed in
+    inp = json.loads(sys.stdin.read())
+    log.trace("Input: " + json.dumps(inp))
+    query_value = inp['title'] if 'title' in inp else inp['name']
+    if not query_value:
+        log.error('No title or name Entered')
```
bnkai, line 139: we should probably return an empty array here and thus skip the search

scrapers/RealityLovers.py
```diff
@@ -0,0 +1,197 @@
...
+        "isInitialLoad": True,
+        "videoView": "MEDIUM",
+        "device": "DESKTOP"
+    }
+    scraped_scenes = session.post(f"https://realitylovers.com/videos/search?tc={time.time()}",
+                                  data=data)
+    log.debug("Called: " + scraped_scenes.url + " with body: " + json.dumps(data))
+    if scraped_scenes.status_code >= 400:
+        log.error('HTTP Error: %s' % scraped_scenes.status_code)
+
```
bnkai, line 156: same as scenes/performers here. If there is an http error return immediately

scrapers/RealityLovers.py
kozobot: Not sure why I needed these to make it work originally.  Removed and retested, it works without them. 

I left the session declaration where it was so I wouldn't have to refactor a bunch.

scrapers/RealityLovers.py
kozobot: There was code duplication here as well.  Moved to a function and fixed the return.

scrapers/RealityLovers.py
kozobot: Fixed, see above comment.

scrapers/RealityLovers.py
kozobot: Fixed

scrapers/RealityLovers.py
kozobot: Fixed

PR #982, opened 2022-04-27
Add scrapers for mariskax.com

cambiph:
Add sceneByUrl and performerByUrl scraper for Mariskax.com

### DIFF
--- a/scrapers/mariskax.yml
+++ b/scrapers/mariskax.yml
@@ -1,0 +1,42 @@
+name: "MariskaX"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - mariskax.com
+    scraper: sceneScraper
+
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - mariskax.com/model
+    scraper: performerScraper  
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h1[@class="title"]
+      Date:
+        selector: //span[@class="post-date"]
+        postProcess:
+          - replace:
+              - regex: (\d+)(st|nd|rd|th)
+                with: "$1"
+          - parseDate: 2 Jan 2006
+      Image: 
+        selector: //video[@id="ypp-player"]/@poster
+        postProcess:
+          - replace:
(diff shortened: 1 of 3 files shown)


scrapers/mariskax.yml
```diff
@@ -0,0 +1,42 @@
...
+        postProcess:
+          - replace:
+            - regex: (?:background-image:url\(\")(.+)(?:\"\));?
+              with: $1
+      Studio:
+        Name: 
+          fixed: MariskaX
+      Performers:
+        Name:
+          selector: //div[@class="content-meta"]/h4[@class="models"]/a
```
bnkai, line 36: Since you are using the performer scraper you could add the URL along with the name. You can then scrape for the performer image without having to copy/paste it
```yaml
        Name: //div[@class="content-meta"]/h4[@class="models"]/a
        URL: //div[@class="content-meta"]/h4[@class="models"]/a/@href
```

cambiph: I don't think I completely understand. Do you mean that I should add Name and Image to the sceneScraper like so:
```
xPathScrapers:
  sceneScraper:
    scene:
      Title: //h1[@class="title"]
      Date:
        selector: //span[@class="post-date"]
        postProcess:
          - replace:
              - regex: (\d+)(st|nd|rd|th)
                with: "$1"
          - parseDate: 2 Jan 2006
      Image: 
        selector: //video[@id="ypp-player"]/@poster
        postProcess:
          - replace:
            - regex: (?:background-image:url\(\")(.+)(?:\"\));?
              with: $1
      Studio:
        Name: 
          fixed: MariskaX
      Performers:
        Name:
          selector: //div[@class="content-meta"]/h4[@class="models"]/a
        Image: 
          selector: //div[@class="model-photo"]/img/@data-src  
        Url:
          selector: //div[@class="content-meta"]/h4[@class="models"]/a/@href   ```

And then remove the performerScraper?

bnkai: No i just meant adding the `URL:` part
I will update the scraper and merge

PR #983, opened 2022-04-29
Strip commas from tags

cardiac-cycle:
lesbea & bellesa (and perhaps others) are including trailing commas in the content of the HTML element for the tag - so you end up with a parsed tag like "Socks,". This strips them out and hopefully shouldn't be destructive for any sites that aren't doing this.


cardiac-cycle:
need to adjust date first

PR #984, opened 2022-04-29
Strip commas from tags

cardiac-cycle:
lesbea & bellesa (and perhaps others) are including trailing commas in the content of the HTML element for the tag - so you end up with a parsed tag like "Socks,". This strips them out and hopefully shouldn't be destructive for any sites that aren't doing this.

### DIFF
--- a/scrapers/RealityKingsOL.yml
+++ b/scrapers/RealityKingsOL.yml
@@ -30,7 +30,12 @@ xPathScrapers:
           - parseDate: January 2, 2006
       Details: //span/div[text()="Description:"]/following-sibling::text()
       Tags:
-        Name: //span[div[text()='Categories:']]/a/text()[1]
+        Name:
+          selector: //span[div[text()='Categories:']]/a/text()[1]
+          postProcess:
+            - replace:
+                - regex: \,
+                  with: ""
       Performers:
         Name: //h1/following-sibling::div//a[contains(@href,"/model")]/text()
       Studio:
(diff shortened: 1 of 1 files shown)


scrapers/RealityKingsOL.yml
```diff
@@ -30,7 +30,12 @@ xPathScrapers:
           - parseDate: January 2, 2006
       Details: //span/div[text()="Description:"]/following-sibling::text()
       Tags:
-        Name: //span[div[text()='Categories:']]/a/text()[1]
+        Name:
+          selector: //span[div[text()='Categories:']]/a/text()[1]
+          postProcess:
+            - replace:
+                - regex: \,
```
bnkai, line 37: Tags in most sites don't seem to have commas as part of them so removing all commas might be ok
Just to be safe can't we only replace leading or trailing commas though?
Something like 
```yaml
                - regex: \,$|^\,
```
?

cardiac-cycle: Good point. Might get thrown if there's whitespace on the ends of the tags - is whitespace ever an issue on tag ends?

How about
`
                - regex: \,\s*$|^\s*\,`

Or even

`                - regex: \s*\,?\s*$|^\s*\,?\s*`

That's spaces, a possible comma and then more spaces - so good against " , tag , " or "  tag  " but doesn't break "Tit, Large"

(Am I overthinking this?)

bnkai: i think the first you mention (`- regex: \,\s*$|^\s*\,`) is enough. Even my suggestion was just to be safe...

PR #985, opened 2022-04-29
Add performer scrapers for Kpopping

bacchus-enorches:


scrapers/Kpopping.yml
```diff
@@ -0,0 +1,80 @@
...
+      DeathDate:
+        selector: $summary//div[@class="equal"][contains(., "Passed away:")]/following-sibling::div[1]/text()
+        postProcess:
+          - parseDate: Jan 2, 2006
+      Country: $summary//strong[contains(., "Country:")]/following-sibling::p/text()
+      Height:
+        selector: $summary//div[@class="equal"][contains(., "Height:")]/following-sibling::div[1]/text()
+        postProcess:
+          - replace:
+              - regex: '\d''\d"[\s\xA0]*\((\d+)[\s\xA0]*cm\)'
```
bnkai, line 54: The regex doesnt work with double digit eg 5''10, a \d+ instead of \d could fix that. Even then it feels a bit too complex
A simpler
```yaml
              - regex: '.*\((\d+)[^\)]*cm\).*'
 ```
should work fine since we are only interested for what is in the ()

File: scrapers/Kpopping.yml [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+                with: $1
+      Weight:
+        selector: $summary//div[@class="equal"][contains(., "Weight:")]/following-sibling::div[1]/text()
+        postProcess:
+          - replace:
+              - regex: '\d+(?:\.\d+)?[\s\xA0]*lbs[\s\xA0]*\((\d+)[\s\xA0]*kg\)'
```
bnkai, line 60:
```yaml
              - regex: '.*\((\d+)[^\)]*kg\).*'
```
simpler and should work as well

bacchus-enorches:
Good looking out. Tbh parens don't really matter either as long as the unit is there.

bnkai:
> Good looking out. Tbh parens don't really matter either as long as the unit is there.

If you mean in the specifix regex they are needed so that `\d+` gets all the digits eg  <https://regex101.com/r/IHD9no/1>

bacchus-enorches:
I'm matching against `[\s\xA0]*` as the unit separator rather than `[^\)]*`. That way even if the display changes, it should still match. See https://regex101.com/r/aYgC1d/1.

scrapers/Kpopping.yml
```diff
@@ -0,0 +1,80 @@
...
+        postProcess:
+          - replace:
+              - regex: " †"
+                with: ""
+      Gender:
+        selector: //ol[@class="breadcrumb"]/li[2]/a/text()
+        postProcess:
+          - map:
+              Girl groups: female
+              Boy groups: male
```
bnkai, line 27: A couple more entries for the map
```yaml
              Female idols: female
              Male idols: male
```

bnkai:
> I'm matching against `[\s\xA0]*` as the unit separator rather than `[^\)]*`. That way even if the display changes, it should still match. See https://regex101.com/r/aYgC1d/1.

Yeah that should work as well

PR #987, opened 2022-05-02
Add support for MrDeepFakes

bacchus-enorches:
A couple hot takes in here since it doesn&rsquo;t seem like there&rsquo;s any prior art regarding deepfakes. In general, I&rsquo;ve opted for over&#x2010;scraping so curators can just ignore what they don&rsquo;t want.

- Both the faked celebrity faces and original pornstars are scraped as performers without distinction.
If it&rsquo;s acceptable to add a `deepfake` tag to the deepfaked performers, that would be an easy workaround, but there&rsquo;s definitely potential for false positives with that approach.
- Verified uploaders, which I&rsquo;m interpreting as the closest thing to content creators here, are scraped as studios.

If there&rsquo;s an XPath way to do gallery search by name, that should be easy to add as it&rsquo;s analogous to scene search.

Also, I can&rsquo;t tell whether or not using a subscraper to grab a whole performer by URL from a scene/gallery, rather than just string values, is possible with XPath. But if it is, it should be dead simple since all the information is there.

### DIFF
--- a/scrapers/MrDeepFakes.yml
+++ b/scrapers/MrDeepFakes.yml
@@ -1,0 +1,100 @@
+name: MrDeepFakes
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - https://mrdeepfakes.com/celebrities/
+      - https://mrdeepfakes.com/pornstars/
+    scraper: performerScraper
+sceneByName:
+  action: scrapeXPath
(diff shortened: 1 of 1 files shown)


feederbox826:
Nevermind, will just pull changes and rebase to avoid such a messy merge

feederbox826:
Thanks, and sorry for the long wait

PR #988, opened 2022-05-05
HQPorner

DogmaDragon:
Decided to start learning xPath and noticed that there was a request https://github.com/stashapp/CommunityScrapers/issues/73#issuecomment-1076214178 for this one, so here it is, a little barebone but it works. 

Couldn't figure out how to extract the image from the iframe. 

Feedback is welcome!

### DIFF
--- a/scrapers/HQPorner.yml
+++ b/scrapers/HQPorner.yml
@@ -1,0 +1,29 @@
+name: "HQPorner"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - hqporner.com/hdporn
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+    #Title: //h1[@class='main-h1']/text()
+      Title: 
+        selector: //title
+        postProcess:
+          - replace:
+              - regex: (.*)-.*
+                with: $1
+      Tags:
+        Name: //a[@class='tag-link click-trigger']/text()
+      Performers:
+        Name:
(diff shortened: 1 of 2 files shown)


bnkai:
I updated the last updated date to pass the validation and added the image part (tested only for a couple of scenes)

To get the image we need to use a `subscraper` and get the poster from the script text

DogmaDragon:
Tested getting the image it works well with the latest player they use, but not with a different one. 
E.g. https://hqporner.com/hdporn/77657-Kink_-_Petite_Whore_Inserts_Her_Small_Hand_In_A_Butt_-_Everything_Butt.html
From what I understand it's because it doesn't pass the poster via script it just adds it in the <video poster="">. 

Not sure how much of a problem this is as it seems that only a minority of older videos use that player. 

Thank you for the information and example.

scrapers/HQPorner.yml
```diff
@@ -0,0 +1,41 @@
...
+            - map:
+                HQPORNER: HQPorner
+      Image:
+        selector: //iframe[@width="560"]/@src
+        postProcess:
+          - replace: &https
+              - regex: ^//
+                with: "https://"
+          - subScraper:
+              selector: //script[contains(.,"poster=")]
```
bnkai, line 35:
```yaml
              selector: //video/@poster|//script[contains(.,"poster=")]
```
This should cover the `older` player case then. @DogmaDragon  can you verify for the scenes you mention?

DogmaDragon: Tested and it works.

PR #992, opened 2022-05-17
Add dlsite scraper

Releck:
Adds an xpath scraper for dlsite.com. I've only supported Japanese and English but Chinese and Korean should have partial data. The reason I've used text selectors is because the main table can omit/add information so ordinal paths won't be consistent.

### DIFF
--- a/scrapers/dlsite.yml
+++ b/scrapers/dlsite.yml
@@ -27,7 +27,7 @@ galleryByFragment:
   action: scrapeXPath
   queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html
   #queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html/?locale=en_US
-  scraper: galleryScraper
+  scraper: galleryManiaxScraper
 galleryByURL:
   - action: scrapeXPath
     url:
--- a/scrapers/dlsite.yml
+++ b/scrapers/dlsite.yml
@@ -1,0 +1,100 @@
+name: DLsite
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html
+  #queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html/?locale=en_US
+  scraper: sceneScraper
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - dlsite.com/maniax/work/
+    scraper: sceneScraper
+    queryURL: "{url}"
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/fsr/=/language/jp/sex_category%5B0%5D/male/keyword/{}/order%5B0%5D/trend/per_page/30/from/fs.header
+  scraper: sceneQueryScraper
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
+  scraper: sceneScraper
+galleryByFragment:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html
+  #queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html/?locale=en_US
+  scraper: galleryScraper
+galleryByURL:
+  - action: scrapeXPath
+    url:
+      - dlsite.com/maniax/work/
+    scraper: galleryScraper
+    queryURL: "{url}"
+
+xPathScrapers:
+  sceneQueryScraper:
+    scene:
+      Title: //dd[@class='work_name']/div[@class='multiline_truncate']/a/text()
+      URL: //dd[@class='work_name']/div[@class='multiline_truncate']/a/@href
+      Image: //a[@class='work_thumb_inner']/img/@src
+  sceneScraper:
+    scene:
+      Title: &titleSel //h1[@id='work_name']/text()
+      Date: &dateSel
+        selector: "//th[text() = '販売日']/following-sibling::td/a/text()|//th[text() = 'Release date']/following-sibling::td/a/text()"
+        postProcess:
+          - replace:
+              - regex: "Jan"
+                with: "01"
+              - regex: "Feb"
+                with: "02"
+              - regex: "Mar"
+                with: "03"
+              - regex: "Apr"
+                with: "04"
+              - regex: "May"
+                with: "05"
+              - regex: "Jun"
+                with: "06"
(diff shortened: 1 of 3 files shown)


Releck:
Hi. Is there any chance of a review for this PR?

bnkai:
> Hi. Is there any chance of a review for this PR?

Sorry forgot a few scrapers... (I tend to check the easier to merge first)

scrapers/dlsite.yml
```diff
@@ -0,0 +1,100 @@
...
+      Title: //dd[@class='work_name']/div[@class='multiline_truncate']/a/text()
+      URL: //dd[@class='work_name']/div[@class='multiline_truncate']/a/@href
+      Image: //a[@class='work_thumb_inner']/img/@src
+  sceneScraper:
+    scene:
+      Title: &titleSel //h1[@id='work_name']/text()
+      Date: &dateSel
+        selector: "//th[text() = '販売日']/following-sibling::td/a/text()|//th[text() = 'Release date']/following-sibling::td/a/text()"
+        postProcess:
+          - replace:
```
bnkai, line 45: An extra cleaning is needed 
```yaml
              - regex: \s.+$
                with: ""
```
as in some cases somethig like ` 0` or ` 0時` ends ups in the date
eg random sample `https://www.dlsite.com/maniax/work/=/product_id/RJ394653.html`

scrapers/dlsite.yml
```diff
@@ -0,0 +1,100 @@
...
+              - regex: "Oct"
+                with: "10"
+              - regex: "Nov"
+                with: "11"
+              - regex: "Dec"
+                with: "12"
+              - regex: (\d{2})/(\d{2})/(\d{4})
+                with: "${3}年${1}月${2}日"
+          - parseDate: 2006年01月02日
+      Details: &detailsSel //div[@class='work_parts_area']/descendant-or-self::*
```
bnkai, line 73: Get all of the details (some have extras in the class)
```yaml
      Details: &detailsSel
        selector: //div[contains(@class,"work_parts_area")]
        concat: "\n\n"
```

scrapers/dlsite.yml
```diff
@@ -0,0 +1,100 @@
...
+  scraper: sceneScraper
+galleryByFragment:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html
+  #queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html/?locale=en_US
+  scraper: galleryScraper
+galleryByURL:
+  - action: scrapeXPath
+    url:
+      - dlsite.com/maniax/work/
```
bnkai, line 29:
```yaml
      - dlsite.com/maniax/work/
      - dlsite.com/books/work/
 ```
Add galleries from the books section as well ? Eg `https://www.dlsite.com/books/work/=/product_id/BJ560145.html/?locale=en_US` or `https://www.dlsite.com/books/work/=/product_id/BJ560145.html`

File: scrapers/dlsite.yml (hunk 1 above)
Comment (Releck) [reply scrapers/dlsite.yml:45]: Thanks. Hadn't seen a release that included the hour before.

File: scrapers/dlsite.yml [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+    scraper: sceneScraper
+    queryURL: "{url}"
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/fsr/=/language/jp/sex_category%5B0%5D/male/keyword/{}/order%5B0%5D/trend/per_page/30/from/fs.header
+  scraper: sceneQueryScraper
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
```
bnkai, line 19:
```yaml
  queryURL: "{url}"
  #queryURLReplace:
  #    url:
  #      - regex: $
  #        with: "/?locale=en_US"
```
using a url replace we can get the en page of the query results

scrapers/dlsite.yml
```diff
@@ -0,0 +1,100 @@
+name: DLsite
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html
+  #queryURL: https://www.dlsite.com/maniax/work/=/product_id/{title}.html/?locale=en_US
+  scraper: sceneScraper
```
bnkai, line 6: This expects that the title is exactly the same as the sites title?
Won't that create too much trouble especially for users that dont know that a specific title is needed?
Shouldnt we remove all the the `sceneByFragment` section since there is a `sceneByName/sceneByQueryFragment` that works better?

Releck: The expectation was that someone would have the title as RJxxxx. I was mainly thinking of identify users.

scrapers/dlsite.yml
Releck: Sure I can include that. Might need a different galleryScraper though as there's some massaging needed around the 著者 field instead of the 作者 one that maniax use.

scrapers/dlsite.yml
bnkai: I am not familiar with the site. Do the downloaded files have the id/RJxxx as their filename or is the user supposed to name the files like that? It should be ok, just add an extra comment of the expected title/filename for the scenebyfragment  to work

scrapers/dlsite.yml
bnkai: Did a quick test with the URL added, didnt even notice the extra `massaging` needed...
If it needs a lot of extra work its fine as it is (no books)

scrapers/dlsite.yml
Releck: It would be reasonable for a user to have a file that was just RJxxxx as the filename as that's how works from DLsite are generally referred to (similar to JAV codes). Happy to leave a comment clarifying that it'll need the RJ code as the title.

scrapers/dlsite.yml
Releck: It shouldn't be too much effort. Happy to get this as feature complete as possible.

Releck:
I've made the changes based off the review. Some other notes:
- The gallery fragment scraper uses maniax only. This is because users are more likely to use maniax over books as maniax has unique content not found elsewhere. Like the scene scraper it'll require the RJ code as the title but it's expected that a user would have `RJxxx.zip` as a gallery.
- I've fixed an issue where some releases had no hyperlink for the date so it wasn't getting picked up.

Lastly thanks @bnkai for the thorough review.

PR #995, opened 2022-05-20
Add xPath scraper for japornxxx.com

CarlNs92891:
Except for the image everything works from my testing. Any assistance is greatly appreciated.

### DIFF
--- a/scrapers/JapornXXX.yml
+++ b/scrapers/JapornXXX.yml
@@ -8,24 +8,23 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: //*[@id="player-content"]/div/div[1]/div[3]/h2
-      Details: //*[@id="player-content"]/div/div[1]/div[3]/div[1]/div/div[2]/p/span
+      Title: //*[@id="player-content"]//h2
+      Details: //*[@id="player-content"]//div[@class="description"]/p/span
       Date:
-        selector: //*[@id="player-content"]/div/div[1]/div[3]/div[1]/div/div[2]
+        selector: //*[@id="player-content"]//strong[text()="Publish Date:"]/following-sibling::text()[1]
         postProcess:
-          - replace: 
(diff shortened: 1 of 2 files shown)


bnkai:
The image part is loaded using a java script so using something like this is needed
```yaml
      Image:
        selector: //script[not(@type)][contains(.,"VodPlayer(")]
        postProcess:
          - replace:
              - regex: '.+image: "([^"]+).+'
                with: $1
```
to get the `<script>` part and then extract the image url

scrapers/JapornXXX.yml
```diff
@@ -0,0 +1,31 @@
...
+  - action: scrapeXPath
+    url:
+      - japornxxx.com
+    scraper: sceneScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //*[@id="player-content"]/div/div[1]/div[3]/h2
+      Details: //*[@id="player-content"]/div/div[1]/div[3]/div[1]/div/div[2]/p/span
```
bnkai, line 12: same here
```yaml
      Details: //*[@id="player-content"]//div[@class="description"]/p/span
```

scrapers/JapornXXX.yml
```diff
@@ -0,0 +1,31 @@
...
+      Date:
+        selector: //*[@id="player-content"]/div/div[1]/div[3]/div[1]/div/div[2]
+        postProcess:
+          - replace: 
+            - regex: '^(?:Length:)\s+(?:[0-9]\d{1}):(?:[0-9]\d{1})\s+(?:Publish Date:)\s+([0-9]\d{0,1})\s+(Jan(?:uary)?|Feb(?:ruary)?|Mar(?:ch)?|Apr(?:il)?|May|Jun(?:e)?|Jul(?:y)?|Aug(?:ust)?|Sep(?:tember)?|Oct(?:ober)?|(Nov|Dec)(?:ember)?)\s+(19\d{2}|20\d{2}).*'
+              with: "$1 $2, $4"
+          - parseDate: 2 January, 2006
```
bnkai, line 19: Instead of getting the whole div just get the text part.
Doing that is safer and doesnt need any regex replace
```yaml
      Date:
        selector: //*[@id="player-content"]//strong[text()="Publish Date:"]/following-sibling::text()[1]
        postProcess:
          - parseDate: 2 January 2006
```

bnkai:
Tags can also be scraped using 
```yaml
      Tags:
         Name: //*[@id="player-content"]//div[@class="tags"]//a
```

CarlNs92891:
Thank you so much for your detailed feedback! The file has been tested and updated.

CarlNs92891:
Thank you so much for your detailed feedback @bnkai! The file has been tested and updated.

PR #998, opened 2022-05-21
Add sceneByFragment for PissPlay

DogmaDragon:
Only supports the format where the filename matches the title name in the URL.

Requested here https://github.com/stashapp/CommunityScrapers/issues/73#issuecomment-1080512271

### DIFF
--- a/scrapers/PissPlay.yml
+++ b/scrapers/PissPlay.yml
@@ -4,6 +4,16 @@ sceneByURL:
     url:
       - pissplay.com
     scraper: sceneScraper
+    
+sceneByFragment:
+  action: scrapeXPath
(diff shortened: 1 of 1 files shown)


bnkai:
Just double checking
For this to work the file must have the exact name of the title.
Do the original files have that form? With the same separators as in the url?
I wouldnt want the majority of the users complaining that we only get 404s

DogmaDragon:
E.g. https://pissplay.com/videos/**title-of-the-video**
The filename has to be **title-of-the-video**.ext

I don't have an example of the original file structure as I don't have access to the site. Only basing the scraper on the information provided in the request. 

I can wait until someone who knows the structure can chime in and update the regex to match that.

PR #999, opened 2022-05-21
Add ShinysBoundSluTS xPath scene scraper

DogmaDragon:

PR #1000, opened 2022-05-21
Add ShinyBound xPath scene scraper

DogmaDragon:


bnkai:
It seems that you are using the exact same selectors as #999 
Both scrapers can/should be merged into one
```yaml
name: "ShinyBound"
sceneByURL:
  - action: scrapeXPath
    url:
      - shinybound.com
      - shinysboundsluts.com
    scraper: sceneScraper

xPathScrapers:
  sceneScraper:
    common:
      $logo: //div[@class="logo"]/a/@href
    scene:
      Title: //span[@class='update_title']
      Details:
        selector: //span[@class='latest_update_description']
      Date:
        selector: //span[@class='availdate']
        postProcess:
          - parseDate: 01/02/2006
      Tags:
        Name: //span[@class='update_tags']/a
      Performers:
        Name:
          selector: //span[@class='tour_update_models']/a
      Studio:
        Name:
          selector: $logo
          postProcess:
            - replace:
                - regex: https://([^.]+)\..+
                  with: $1
            - map:
                shinysboundsluts: ShinysBoundSluTS
                shinybound: ShinyBound
      Image:
        selector: $logo|//img[@class='stdimage promo_thumb left thumbs']/@src
        concat: "/"
# Last Updated May 21, 2022
```

DogmaDragon:
Verified everything works. 

I personally prefer individual files for each of my scrapers, but I'll keep that in might and merge them when possible to consolidate before making a PR.

bnkai:
> I personally prefer individual files for each of my scrapers, but I'll keep that in might and merge them when possible to consolidate before making a PR.

In most cases separate files are prefered. In this case though all the selectors are exactly the same so it is more convenient (imho) to use 1 file.

PR #1006, opened 2022-05-31
Update iFeelMyself.py

VinceEatWorld:
Changed the 'replace' function to unicodedata.normalize() function to clean up the unicode in the descriptions. Added a 'replace' function to change unicode \x92 in some titles to '

### DIFF
--- a/scrapers/IFeelMyself.py
+++ b/scrapers/IFeelMyself.py
@@ -2,6 +2,7 @@ import json
 import re
 import sys
 from datetime import datetime
+import unicodedata
 
 try:
     from mechanicalsoup import StatefulBrowser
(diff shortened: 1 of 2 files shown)


VinceEatWorld:
Does anyone know why unicode is not supported in the text fields? Firefox and chrome simple ignore unicode in the title an or descriptions, while safari generates a funny looking symbol instead.

bnkai:
> Does anyone know why unicode is not supported in the text fields? Firefox and chrome simple ignore unicode in the title an or descriptions, while safari generates a funny looking symbol instead.

I dont the issue is the UI.
For example using this scraper
```yaml
name: CodeTable
sceneByURL:
  - action: scrapeXPath
    url:
      - codetable.net
    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $unicode: //div[@class="bigSymbol left"]
    scene:
      Title: $unicode
      Performers:
        Name: $unicode
      Studio:
        Name: $unicode
      Details: $unicode
```
and trying `https://www.codetable.net/hex/92` or even `https://www.codetable.net/decimal/9819` the unicode chars render fine
Maybe there is an issue with how the text is passed through a script scraper? Does your console log (set to debug) while scraping print the unicode character correctly?

I am merging the PR btw, you can still reply here though if you have some more info/example for the unicode issue.

PR #1008, opened 2022-06-02
Update Backroomcastingcouch.yml

JaseNZC:
A rework of Backroomcastingcouch for missing items.

### DIFF
--- a/scrapers/Backroomcastingcouch.yml
+++ b/scrapers/Backroomcastingcouch.yml
@@ -1,21 +1,27 @@
-name: backroomcastingcouch
-performerByURL:
+name: BackRoomCastingCouch
+sceneByURL:
   - action: scrapeXPath
     url:
-      # Get Data either from model page or a video. Both provide same info
-      - backroomcastingcouch.com/site/models
-      - backroomcastingcouch.com/site/trailers
-
-    scraper: performerScraper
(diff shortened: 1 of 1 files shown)


bnkai:
From a quick view isnt the scraper now a scene scraper instead of a performer one? 
The existing one got the performer name and the image (not a lot of meta :-( )
Shouldnt the scene scraper get the `scene` image instead of the performer image and just get the performer name/url that we can then use for a performer scraper?
(haven't actually tested the scraper yet)

JaseNZC:
> From a quick view isnt the scraper now a scene scraper instead of a performer one? The existing one got the performer name and the image (not a lot of meta :-( ) Shouldnt the scene scraper get the `scene` image instead of the performer image and just get the performer name/url that we can then use for a performer scraper? (haven't actually tested the scraper yet)

Ahhhh I did not notice there was a scraper (ExploitedX) already in play and the other was just for performer.
Worth merging the two ??, if not I will just withdraw this one :-)

bnkai:
> Ahhhh I did not notice there was a scraper (ExploitedX) already in play and the other was just for performer.
> Worth merging the two ??, if not I will just withdraw this one :-)

If the ExploitedX scene scraper is working fine then merging the  Backroomcastingcouch performer scraper into it should be ok?

PR #1010, opened 2022-06-02
[IAFD] use py_common.log for logging

stg-annon:


scrapers/IAFD.py
```diff
@@ -420,25 +420,27 @@ def scrape(url, retries=0):
     try:
         scraped = scraper.get(url, timeout=(3,7))
     except Exception as e:
-        log("scrape error %s" % e)
+        log.error(f"scrape error {e}")
+        sys.exit(1)
```
bnkai, line 424: With the exit added here the retry code fails if there is a timeout.
Adding an extra exception check for the requests timeout exception is needed
for example
```python
    except requests.exceptions.Timeout as exc_time:
        log.debug(f"Timeout: {exc_time}")
        return scrape(url, retries+1)
    except Exception as e:
        log.error(f"scrape error exception {e}")
        exit(1)
```
To make it easier to check you can reduce the timeout from `3,7` to `3,3`
(an import of the requests module is also needed)

Not sure if the same needs to be applied in the scrape_image function as well

bnkai: Yeah sorry, didnt make it clear. The retry code was not working before as well. I noticed it due to this PR (couldnt see any retries after a timeout)

PR #1018, opened 2022-06-08
[Bang.com] add movie and performer scraper

reynn:


reynn:
Updated @Belleyy thanks for the reminder

PR #1019, opened 2022-06-08
[AdultPrime] add basic scraping for AP

stg-annon:


scrapers/AdultPrime.yml
```diff
@@ -0,0 +1,30 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - adultprime.com/studios/video/
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //h2[contains(@class,"update-info-title")]/text()
+      Details: //p[contains(@class,"ap-limited-description-text")]/text()
```
bnkai, line 11: This gets only a part of the description, a post process and concat is needed.
Is the description good/useful to scrape though? Unless i am doing something wrong it seems that the description is meant for the studio and not the scenes, so for every scene of the same studio we get the same description?
eg `https://adultprime.com/studios/video/122995` `https://adultprime.com/studios/video/123113`

scrapers/AdultPrime.yml
```diff
@@ -0,0 +1,30 @@
...
+        selector: //div[starts-with(@style,"background-image:")]/@style
+        postProcess:
+          - replace:
+              - regex: .+url\((.+)\).+
+                with: $1
+      Studio: 
+        Name: //p[contains(@class,"update-info-line")]/b/a[contains(@href,"/studio/")]/text()
+      Tags:
+        Name:
+          selector: //p[@class="update-info-line regular hidden-xs"]/text()
```
bnkai, line 26: Again not sure if we have the same links 
For example for `https://adultprime.com/studios/video/123094` or the above `https://adultprime.com/studios/video/122995` i had to use this selector
```
//p[contains(@class,"update-info-line regular")][contains(b,"Niches:")]/text()
```
As the `hidden-xs` seems to be missing from the class ?

scrapers/AdultPrime.yml
stg-annon: removed as requested; the site might have changed since I threw this together, noticed it wasn't in the repo so I added it.

however it seems that scenes either have a custom description or are given a generic one there are some that are unique
e.x. `https://adultprime.com/studios/video/27075`

PR #1022, opened 2022-06-10
[ModelMediaUS] adding an alt scraper that works with jerkaoke, delphinefilms and mmasia

reynn:


scrapers/ModelMediaUS.yml
```diff
@@ -24,5 +30,28 @@ xPathScrapers:
...
+      Performers:
+        Name: 
+          selector: //div[contains(.,"Cast")]/following-sibling::div//a/text()
+      Tags:
+        Name: $container//h1/following-sibling::div//a[contains(@href,"videos?tag")]/span/text()
+      Image:
+        selector: //iframe[@id="video-player"]/@src
+        postProcess:
+          - subScraper:
+              selector: //stream/@poster
```
bnkai, line 53: This seems to work for the modelmediaassia site but gets access denied for the other sites?
```yaml
      Image: //div[contains(@class,"mus-z-index-2")]/img[contains(@class,"mus-img-cover")]/@src
```
Seems to work for delphine and jerkaoke
From a quick test it seems modelmediaasia needs to be moved to the `sceneScraper`

PR #1024, opened 2022-06-11
[Hegre] Fix scene URL

Flashy78:
Their site now lists scenes at /films instead of /movies, and the title text has a new class on it.

Added perform bio scraping.

### DIFF
--- a/scrapers/Hegre.yml
+++ b/scrapers/Hegre.yml
@@ -30,7 +30,7 @@ xPathScrapers:
       Title: //meta[@property="og:title"]/@content
       Details:
         selector: |
-          //div[@class="film-subtitle translated-text"]/text()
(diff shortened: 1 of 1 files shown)


scrapers/Hegre.yml
```diff
@@ -29,7 +30,7 @@ xPathScrapers:
       Title: //meta[@property="og:title"]/@content
       Details:
         selector: |
-          //div[@class="film-subtitle"]/text()
+          //div[@class="film-subtitle translated-text"]/text()
```
bnkai, line 33: Is this for all scenes?
Would it be safer to use something like
```yaml
          //div[contains(@class,"film-subtitle")]/text()
```

Flashy78: That makes more sense, thanks for the suggestion.

PR #1032, opened 2022-06-22
luxure.com xPath scraper

DogmaDragon:
Requested here https://github.com/stashapp/CommunityScrapers/issues/73#issuecomment-1156667416

Luxure.com have 4 different channels, but they are not mentioned anywhere on the scene/scene code. So using a fixed studio. 

They are also using some JS to block overflow descriptions. So the scraper only gets part of it. I tried to look at `disclaimer.js` but wasn't able to figure out how to bypass it.


bnkai:
Luxure seems to use a json api
It might be better to investigate how to get the data directly from the api via a script scraper.
They use a token that probably doesnt change (it is in one of the js files from the scene)
An example request 
```
curl -H "Accept: application/json" -H "token: mysexmobile" -H "Referer: https://luxure.com" -H "Origin: https://luxure.com"  -H "siteid: 103" "https://api.luxure.com/24api/v1/sites/103/freetour/videos/2498"
```
for `https://luxure.com/en/videos/2498-ready-to-play` gets the json of the scene with all meta (including timed tags we can use for markers)

DogmaDragon:
Closing as per @bnkai suggestion in hope that someone makes a script scraper. In the meantime, the requestor can use it.

PR #1033, opened 2022-06-22
wakeupnfuck.com xPath scraper

DogmaDragon:
Made by @bnkai and posted on Discord.


bnkai:
Just a note for myself when testing later on. The performer selector that includes male+female performers isnt always perfectly `splitable`. The alternative performer selector i posted in the original scraper works fine all the time, it only includes female performers though

DogmaDragon:
From my more in-depth testing, the male + female selector starts to work correctly around 200+ scene id. I think its failure rate is a bit too high to leave it as default.

One idea is to use that selector for details/description so that people could add it manually to their scenes. But I don't know if it's something people at mass would want.

PR #1038, opened 2022-06-25
AEBN xPath performer scraper

DogmaDragon:
Performer scraper for AEBN. 

Requested on Discord.

### DIFF
--- a/scrapers/AEBN.yml
+++ b/scrapers/AEBN.yml
@@ -113,4 +113,4 @@ xPathScrapers:
           - replace:
               - regex: ^([^?]+).*$
                 with: "https:$1"
-# Last Updated June 25, 2022
+# Last Updated July 06, 2022
--- a/scrapers/AEBN.yml
+++ b/scrapers/AEBN.yml
@@ -37,13 +37,13 @@ xPathScrapers:
         selector: //li[@class='section-detail-list-item-height']/text()
         postProcess:
           - replace:
-              - regex: .+\((.+)\)
+              - regex: .+\((\d+).+\)
                 with: $1
       Weight:
         selector: //li[@class='section-detail-list-item-weight']/text()
         postProcess:
           - replace:
-              - regex: .+\((.+)\)
+              - regex: .+\((\d+).+\)
                 with: $1
       EyeColor: 
         selector: //li[@class='section-detail-list-item-eye-color']/text()
--- a/scrapers/AEBN.yml
+++ b/scrapers/AEBN.yml
@@ -1,18 +1,70 @@
-name: AEBN
+name: "AEBN"
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - aebn.com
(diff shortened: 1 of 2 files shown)


scrapers/AEBN.yml
```diff
@@ -10,6 +15,53 @@ movieByURL:
...
     scraper: movieScraper
 xPathScrapers:
+  performerScraper:
+    performer:
+      Name: //div[@class='dts-section-page-heading-title']/h1
+      Gender:
+        selector: //li[@class="section-detail-list-item-gender"]/text()
+        postProcess:
+          - map:
+              TS: transgender_female
```
DogmaDragon, line 25: While this is not always correct, AEBN only has 3 gender categories Female, Male, and TS. And since most Trans content is from Transgender Females, I feel like it's better than leaving it unamtched.

scrapers/AEBN.yml
```diff
@@ -10,6 +15,53 @@ movieByURL:
...
+              - regex: .+:\s(.+)
+                with: $1
+              - regex: "Sept"
+                with: "Sep"
+          - parseDate: Jan 2, 2006
+      Height:
+        selector: //li[@class='section-detail-list-item-height']/text()
+        postProcess:
+          - replace:
+              - regex: .+\((.+)\)
```
bnkai, line 40:
```yaml
              - regex: .+\((\d+).+\)
```
We need to remove the units as well

scrapers/AEBN.yml
```diff
@@ -10,6 +15,53 @@ movieByURL:
...
+                with: $1
+      Weight:
+        selector: //li[@class='section-detail-list-item-weight']/text()
+        postProcess:
+          - replace:
+              - regex: .+\((.+)\)
```
bnkai, line 46: same here
```yaml
              - regex: .+\((\d+).+\)
```

scrapers/AEBN.yml
```diff
@@ -10,6 +15,53 @@ movieByURL:
...
+                with: $1
+      EyeColor: 
+        selector: //li[@class='section-detail-list-item-eye-color']/text()
+      Ethnicity:
+        selector: //li[@class='section-detail-list-item-ethnicity']/text()
+        postProcess:
+          - map:
+              White: Caucasian
```
bnkai, line 54: I think in some of the other scrapers we use the opposite? 
For example the babepedia scraper uses the below
```yaml
          - map:
              Asian: asian
              Caucasian: white
              Black: black
              Latin: hispanic
```

DogmaDragon: I'm basing the scraper entries based on what StashDB has defined here https://github.com/stashapp/stash-box/blob/bae0400549c51ec055920ea326ed736ba8f12987/graphql/schema/types/performer.graphql#L31

So that when people submit them to StashDB they don't have to manually enter the value due to it not getting matched.

PR #1047, opened 2022-06-30
LHF images for BTS + new URL

zosky:
- BTS scenes have no video element *(& main scenes have no img)*
  - added OR in image selector to handle eaither
- also this studio has rolled another site: LoveHerFilms
  - which seems to be all their sites rolled into one place

### DIFF
--- a/scrapers/LoveHerFeet.yml
+++ b/scrapers/LoveHerFeet.yml
@@ -23,12 +23,14 @@ xPathScrapers:
         postProcess:
           - parseDate: January 2, 2006
       Image:
-        selector: //div[@class='video']//img[@class='mainThumb']/@src0_3x or (//base/@href|//video/@poster)
-        concat: /
+        selector: //div[@class='video']//img[contains(@class,'mainThumb')]/@src0_3x|//video/@poster|//base/@href
+        concat: "|"
         postProcess:
           - replace:
-              - regex: ///tour
-                with:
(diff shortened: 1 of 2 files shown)


scrapers/LoveHerFeet.yml
```diff
@@ -22,7 +23,7 @@ xPathScrapers:
         postProcess:
           - parseDate: January 2, 2006
       Image:
-        selector: //base/@href|//video/@poster
+        selector: //div[@class='video']//img[@class='mainThumb']/@src0_3x or (//base/@href|//video/@poster)
```
bnkai, line 26: `or` doesn't work like that
The above selector with the concat and the postprocessing doesnt work for all cases (BTS and plain scenes) 
Also the img class in she loves black BTS is a little different)
I had to use something like 
```yaml
      Image:
        selector: //div[@class='video']//img[contains(@class,'mainThumb')]/@src0_3x|//video/@poster|//base/@href
        concat: "|"
        postProcess:
          - replace:
              - regex: ^(https[^|]+)\|.+ # mainThumb urls do not need anything
                with: $1
              - regex: ^/tour([^|]+)\|(.+) # video/@poster urls need a domain
                with: $2$1
```
The pasted Image part seems to work fine, can you verify for scenes and BTS videos from all sites in the scraper?

zosky: thanks for making sense of what i was aiming for.

PR #1048, opened 2022-06-30
Tweak data18 movie duration selector/pp

bnkai:
Tweaks the duration selector mentioned in #123 and fixes a typo in the details pp.
For movies that have a `proper` duration that is used and we then fallback to the plain  text to catch cases like `120 min`
Sample movies `https://www.data18.com/movies/1225101-raw-43` , `https://www.data18.com/movies/1231098-it-takes-three-6` mentioned in #123 work fine. Problematic movies `https://www.data18.com/movies/1239543-naughty-little-sister-6` also seem to work ok. If there is another case please mention a URL or the format used.

### DIFF
--- a/scrapers/data18.yml
+++ b/scrapers/data18.yml
@@ -78,4 +78,4 @@ xPathScrapers:
                 with:
       FrontImage: //a[@id='enlargecover']/@href
       BackImage: //a[text()='+Back']/@href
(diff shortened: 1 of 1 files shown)


michael-d-stone:
Thanks, @bnkai! Works as expected.

PR #1050, opened 2022-07-03
Add hypnotube.com xPath scene scraper

Triskelia:
Add scene url scraper for hypnotube.com
Contains mostly Hypnosis, JOIs, PMVs, Feminization, Sissification and Bimbofication videos

Scrapes : 
- Title
- Details
- Image (From curated thumbnail, as shown in the search section of the site)
- Studio
  - Name (poster's username)
  - URL (poster's hypnotube profile url)
- Date
- Tags (From both Tags and Categories sections on the page)

### DIFF
--- a/scrapers/Hypnotube.yml
+++ b/scrapers/Hypnotube.yml
@@ -1,0 +1,28 @@
+name: "Hypnotube"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - hypnotube.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $studio: //a[starts-with(@href,'https://hypnotube.com/user/')]
+    scene:
+      Title: //div[@class='item-tr-inner-col inner-col']/h1/text()
+      Details: //meta[@name='description']/@content
+      Image: //meta[@property='og:image']/@content
+      Studio:
+        Name: $studio/@title
+        URL: $studio/@href
+      Date:
+        selector: //div[@class='stats-container']/ul/li[3]/span[@class='sub-label']/text()
+        postProcess:
+          - replace:
+            - regex: \s\d{2}:\d{2}:\d{2}
+              with: ""
+      Tags:
+        Name: //div[@class='tags-block']/a/text()
+      URL: //link[rel='canonical']/@href
+
+# Last Updated July 03, 2022
--- a/scrapers/Hypnotube.yml
+++ b/scrapers/Hypnotube.yml
@@ -10,19 +10,24 @@ xPathScrapers:
       $studio: //a[starts-with(@href,'https://hypnotube.com/user/')]
     scene:
       Title: //div[@class='item-tr-inner-col inner-col']/h1/text()
-      Details: //meta[@name='description']/@content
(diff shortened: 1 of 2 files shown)


scrapers/Hypnotube.yml
```diff
@@ -0,0 +1,28 @@
...
+      - hypnotube.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $studio: //a[starts-with(@href,'https://hypnotube.com/user/')]
+    scene:
+      Title: //div[@class='item-tr-inner-col inner-col']/h1/text()
+      Details: //meta[@name='description']/@content
+      Image: //meta[@property='og:image']/@content
```
bnkai, line 14: Not working for me (at least with the scenes i checked)
I tried to pull the image from a js that is in the page. It seems to work ok for me.
```yaml
      Image:
        selector: //script[@type="text/javascript" and contains(text(),"twitter:image")]/text()
        postProcess:
          - replace:
              - regex: '.+"twitter:image" content="([^"]+)".+'
                with: $1
```

scrapers/Hypnotube.yml
```diff
@@ -0,0 +1,28 @@
...
+    url:
+      - hypnotube.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $studio: //a[starts-with(@href,'https://hypnotube.com/user/')]
+    scene:
+      Title: //div[@class='item-tr-inner-col inner-col']/h1/text()
+      Details: //meta[@name='description']/@content
```
bnkai, line 13: For longer descriptions details are truncated...
We could get the details from the same script that has the image. Some special chars would need to be replaced though. Is it worth it to check that out?

scrapers/Hypnotube.yml
```diff
@@ -0,0 +1,28 @@
...
+      Image: //meta[@property='og:image']/@content
+      Studio:
+        Name: $studio/@title
+        URL: $studio/@href
+      Date:
+        selector: //div[@class='stats-container']/ul/li[3]/span[@class='sub-label']/text()
+        postProcess:
+          - replace:
+            - regex: \s\d{2}:\d{2}:\d{2}
+              with: ""
```
bnkai, line 23: a little simpler
```yaml
      Date:
        selector: //li/span[@class="icon i-calendar"]/following-sibling::span
        postProcess:
          - replace:
              - regex: \s.*$
                with: ""
```

scrapers/Hypnotube.yml
Triskelia: Looks like they changed something on the website. Your method works well. I've added it in my last commit

scrapers/Hypnotube.yml
Triskelia: Indeed, I should have tested videos with longer description. In my last commit, I went for the actual description in the page (class "main-description"). That way, we don't need to replace special char as we would have to if we were to scrap it from the js.

scrapers/Hypnotube.yml
Triskelia: Looks good. Added it in my last commit

PR #1057, opened 2022-07-10
[dc-onlyfans] Add gallery scraping

DefinitelyNotSpicy:
This adds scraping for DC-Onlyfans galleries.

Works for me on Windows 11. I'm not sure if it would work on Unix cause I'm querying with the full path of the gallery using `.as_posix()`

### DIFF
--- a/scrapers/dc-onlyfans.py
+++ b/scrapers/dc-onlyfans.py
@@ -56,6 +56,42 @@ def lookup_scene(file,db,parent):
 
     return res
 
+def lookup_gallery(file,db,parent):
+    log.info(f"using database: {db.name}  {file.name}")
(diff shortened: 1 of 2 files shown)


bnkai:
From what i understand if `.as_posix()` works for you in windows it should be ok (the tricky part should be the windows part not linux).
I have no sample db to test so i will only check the code. If someone else can verify no issues with linux that would be great.

DefinitelyNotSpicy:
> From what i understand if .as_posix() works for you in windows it should be ok (the tricky part should be the windows part not linux).

You're right. I confused which os was using which slash. Since DC-Onlyfans already uses the posix slash on Windows it should work the same on Linux.

PR #1060, opened 2022-07-15
[Javlibrary python] fix image using search

Belleyy:
#123 

The image from the XPath already contains the protocol

### DIFF
--- a/scrapers/JavLibrary_python.py
+++ b/scrapers/JavLibrary_python.py
@@ -396,7 +396,7 @@ def jav_search_by_name(html, xpath):
             "title": jav_title[count],
             "url":
             f"https://www.javlibrary.com/en/{jav_url[count].replace('./', '')}",
-            "image": f"https:{jav_image[count]}"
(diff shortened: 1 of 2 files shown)


scrapers/JavLibrary_python.py
```diff
@@ -396,7 +396,7 @@ def jav_search_by_name(html, xpath):
             "title": jav_title[count],
             "url":
             f"https://www.javlibrary.com/en/{jav_url[count].replace('./', '')}",
-            "image": f"https:{jav_image[count]}"
+            "image": jav_image[count]"
```
bnkai, line 399: Was it always like that or did they change something in the layout?
The comment in line 391 shows a sample image url without the schema (https:).
Would it be safer to use something like 
```python
            "image": re.sub("^//","https://",jav_image[count])
```
or is it not necessary ?

Belleyy: Yea would be better, i think the layout has changed or when I made this feature I was blind.

PR #1061, opened 2022-07-18
add scraper for pinklabel.tv

scruffynerf:
Adds new scraper for pinklabel.tv, which hosts many 'indie' adult movies found nowhere else

Provides scene and movie scrapers


halorrr (approved):
LGTM, tested it with various examples and they all worked as expected for me.

PR #1063, opened 2022-07-18
Updated StasyQ works for Scenes and Performer scraping

scruffynerf:
Site changes, so I updated this to make it work...


halorrr (approved):
Tested this with various scenes and performers and it worked

PR #1069, opened 2022-07-22
Switch to a python scraper for ManyVids

bnkai:
- A python scraper is used to fetch meta directly from json endpoints. All scene scraper functionality should remain the same with the old scraper
- A new sub folder is now used as this scraper has more than 1 files. (According to the recent discussion in discord. This was the main reason for the scraper as well)

A few things that need some testing/verification
- [x] The py_common import works for me in linux and should work in all other os as well. Needs some verification though (thanks @Belleyy )

<del>TODO</del>DONE
- [x] A regex replace / mapping of the gender values for the performer scraper
- [x] Get the rest of the performer meta that is not available via the json api endpoint using the performers about page and plain xpath selectors
- [x] Add a requirements/readme file since we now have a separate directory

EDIT
The python scraper should now have the same capabilities as the CDP it replaces


scrapers/ManyVids/ManyVids.py
```diff
@@ -0,0 +1,332 @@
...
+    date = model_meta.get('dob')
+    if date:
+        date = re.sub(r"T.*", "", date)
+        scrape['birthdate'] = date
+
+    description = model_meta.get('description')
+    scrape['details'] = clean_text(description)
+
+    scrape['gender'] = model_meta.get('identification')
+    if scrape['gender'] and scrape['gender'] == "Trans":
```
Belleyy, line 177: don't need the `if scrape['gender']`, only  `scrape['gender'] == "Trans"` is needed

scrapers/ManyVids/ManyVids.py
```diff
@@ -0,0 +1,332 @@
...
+        # top scene tags for performer
+        tags = page_tree.xpath('//li[@class="mv-top-tags__item"]//text()')
+        scrape['tags'] = [{"name": x.strip()} for x in tags if x.strip() != ""]
+        
+        ethnicity = page_tree.xpath("//span[@class='mv-about__container__details__list-label'][contains(text(), 'Ethnicity')]/following-sibling::span/text()")
+        if ethnicity:
+            scrape['ethnicity'] = map_ethnicity(ethnicity[0])
+        
+        country = page_tree.xpath("//span[@class='mv-about__container__details__list-label'][contains(text(), 'Nationality')]/following-sibling::span/img/@alt")
+        if  country:
```
Belleyy, line 199:
```suggestion
        if country:
```

scrapers/ManyVids/ManyVids.py
```diff
@@ -0,0 +1,332 @@
...
+        tattoos = page_tree.xpath("//span[@class='mv-about__container__details__list-label'][contains(text(), 'Tattoos')]/following-sibling::span/text()")
+        if tattoos:
+            scrape["tattoos"] = tattoos[0]
+
+        piercings = page_tree.xpath("//span[@class='mv-about__container__details__list-label'][contains(text(), 'Piercings')]/following-sibling::span/text()")
+        if piercings:
+            scrape["piercings"] = piercings[0]
+        
+        measurements = page_tree.xpath("//span[@class='mv-about__container__details__list-label'][contains(text(), 'Measurements')]/following-sibling::span/text()")
+        if piercings:
```
Belleyy, line 219:
```suggestion
        if measurements:
```

Belleyy:
Tested on some URL/Performer, seems ok

PR #1071, opened 2022-07-24
Add search to Clips4sale

JackDawson94:
Add search scenes by name to Clips4sale.

Haven't touched the existing scene scraper.

### DIFF
--- a/scrapers/Clips4Sale.yml
+++ b/scrapers/Clips4Sale.yml
@@ -10,7 +10,30 @@ sceneByURL:
       - clips4sale.com/search/clip/
     scraper: c4sClipsSceneScraper
 
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://www.clips4sale.com/clips/search/{}
+  scraper: c4sSearch
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
+  scraper: c4sSceneScraper
+
 xPathScrapers:
+  c4sSearch:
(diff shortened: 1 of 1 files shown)


thismanyboyfriends (changes requested):
I've tried this out myself and I've found that whilst the search works well, and I get a list of results (often correct/very accurate), if I try to actually scrape with them I get an unsupported protocol error.

Specific error message is:

`error while fragment scraping with scraper Clips4Sale: get (search text) unsupported protocol sceme ""`

A shame as I really want a clips4sale name scraper, there's a lot of content on there and manually searching URLs is tedious.

scrapers/Clips4Sale.yml
```diff
@@ -10,7 +10,30 @@ sceneByURL:
...
+
 xPathScrapers:
+  c4sSearch:
+    common:
+      $card: //figure[contains(@class, "splide__slide")]
+    scene:
+      Title:
+        selector: $card/@aria-label
+      URL:
+        selector: $card//a[contains(@class, "search-clip__title")]
```
thismanyboyfriends, line 30: Found the error. This xpath resolves to the inner text, not the URL. You want to append `/@href` on the end.

JackDawson94: Thanks, pushed a fix :)

scruffynerf:
works for me!

thismanyboyfriends (approved):

PR #1072, opened 2022-07-24
Femdom Empire scene search by name

thismanyboyfriends:
Can now search by name using Femdom Empire's site search.

Sadly their site search feature is a bit ropey, I find it's very inclusive and if any of the words are commonly used, it'll grab everything even if you type the exact name of what you're looking for. You usually get better results if you type in a single (preferably unusual) word.

### DIFF
--- a/scrapers/FemdomEmpire.yml
+++ b/scrapers/FemdomEmpire.yml
@@ -1,10 +1,18 @@
-name: "FemdomEmpire"
+name: Femdom Empire
 sceneByURL:
   - action: scrapeXPath
     url:
       - femdomempire.com/tour/trailers/
       - feminized.com/tour/trailers/
     scraper: sceneScraper
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://femdomempire.com/tour/search.php?query={}
+  scraper: sceneSearch
+sceneByQueryFragment:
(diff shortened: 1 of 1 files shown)


scruffynerf:
looks good... works for me.

scrapers/FemdomEmpire.yml
```diff
@@ -60,6 +68,14 @@ xPathScrapers:
...
-
+  sceneSearch:
+    common:
+      $itemThumb: //div[@class='item-thumb']
+    scene:
+      Title: $itemThumb/a/@title
+      URL: $itemThumb/a/@href
+      Image: $itemThumb/a/img/@src
+driver:
+  useCDP: true
```
bnkai, line 79: Adding CDP is not needed and it makes the whole scraper a lot slower
You just need to get the correct selectors from page source eg
```yaml
      Image:
        selector: $itemThumb/a/img/@*[name()='src0' or name()='src0_1x']
        postProcess:
          - replace:
              - regex: ^/
                with: https://femdomempire.com/
```
seems to work for me

feederbox826:
Superceded by https://github.com/stashapp/CommunityScrapers/pull/1992 with suggested Image lookup

PR #1073, opened 2022-07-25
ClubDom and SubbyHubby name scrapers

thismanyboyfriends:
Adds name search scraping for ClubDom.com and SubbyHubby.com native site search.

Previously the ClubDom and SubbyHubby scrapers were in one - which makes sense given they duplicate a lot and are part of the same network. However they have different name search queries, so in order to search either by name, they have to be split out into different scrapers.

### DIFF
--- a/scrapers/ClubDom.yml
+++ b/scrapers/ClubDom.yml
@@ -1,16 +1,16 @@
-name: SubbyHubby
+name: ClubDom
 sceneByURL:
   - action: scrapeXPath
     url:
-      - subbyhubby.com/tour
+      - clubdom.com/tour
     scraper: sceneScraper
   - action: scrapeXPath
     url:
-      - subbyhubby.com/vod
+      - clubdom.com/vod
     scraper: vodScraper
 sceneByName:
   action: scrapeXPath
-  queryURL: https://www.subbyhubby.com/vod/search.php?query={}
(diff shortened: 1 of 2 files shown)


scrapers/ClubDom.yml
```diff
@@ -0,0 +1,81 @@
...
+        selector: $update-details/a[2]/text()
+      URL:
+        selector: $update-details/a[2]/@href
+      Image:
+        selector: $update-details/a[1]/img/@src
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https:"
+# Last Updated July 20, 2021
```
Belleyy, line 81:
```suggestion
# Last Updated July 20, 2022
```

scrapers/SubbyHubby.yml
```diff
@@ -61,12 +63,20 @@ xPathScrapers:
...
+        selector: $update-details/a[2]/text()
+      URL:
+        selector: $update-details/a[2]/@href
+      Image:
+        selector: $update-details/a[1]/img/@src
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https:"
 # Last Updated July 20, 2021
```
Belleyy, line 82:
```suggestion
# Last Updated July 20, 2022
```

PR #1084, opened 2022-08-05
Move EvilAngel, Wicked, Diabolic to Algolia

bnkai:
Moves EvilAngel to an Algolia scraper (reported in #123 )
Cleans up a bit the Algolia code and adds initial support for movie scraping (not fully tested on other sites). 
Moved Wicked as well to test the movie support
EDIT
Moved Diabolic to Algolia
Scrape movie name/url when scene scraping via Algolia


stg-annon:
Should probably bump Last Updated in [GammaEntertainment.yml#L294](https://github.com/stashapp/CommunityScrapers/blob/5b0cde1e0c5462507303f701738a81ce1fd72d43/scrapers/GammaEntertainment.yml#L294)

aussiehuddo:
suggest moving tsfactor.com from gammaentertainment.yml to new evilangel algolia as it is part of their network

PR #1086, opened 2022-08-07
tagger and scene url changes Theporndb.yml

MrX292:
tagger and url scene scraper get Performer url, gender, birthdate, ethnicity, height, measurements, tattoos, piercings and image
image url works in log but it does not scrape it


MrX292:
it shows `gender_types.Female` in tagger, but it only seams to be in tagger. after i save it shows `Female` on the profile

bnkai:
`image url works in log but it does not scrape it`
Due to the way the scraping works you can only get a scene's/performer's image when you are scraping for a scene/performer.
Getting the performer image when scene scraping is not possible afaik. Since the image part won't work it can be removed.
The gender issue is known and purely cosmetic, it works fine when saving.

PR #1093, opened 2022-08-11
Correctly handle missing scene/movie dates for IAFD scraper

selfhost-alt:
When a scene in IAFD has no release date listed, the field will have the value `No Data`. When that field is scraped, the scraper tries to parse the string as a date value and then convert it to a Stash date format. That conversion fails when the value is `No Data`, but then the scraper just leaves the `date` field as `No Data` which Stash then rejects when you try to save the scene.

This change tweaks the logic, so if parsing of the scene's release date fails, we set the scene/movie date back to `None` so that Stash doesn't try to update it to an invalid value.


selfhost-alt:
Derp. Thank you for the suggestions

PR #1095, opened 2022-08-13
Subbyhubby and ClubDom search added

scruffynerf:
Added SubbyHubby searching, and then realized that to allow ClubDom searching required splitting the yml into 2.


bnkai:
already a PR at #1073 ? will cross check to see differences

PR #1097, opened 2022-08-13
Hotmovies - small tweak

scruffynerf:
- Adds more info during search (Details and Date) to help decide the right item
- Reuse the existing URL selector (minor, but since I'm adding something else)


scruffynerf:
> yea the url for movies is useless since you need a url to use it

I've been finding movies with no url as a result of other tools.  It's always safer to return a url, if nothing else than to trim bad provided urls down to good URLs.

halorrr (approved):

PR #1100, opened 2022-08-14
Virtualtaboo Fix studio name not beeing scraped

Foucoubou:


scrapers/VirtualTaboo.yml
```diff
@@ -26,6 +26,9 @@ xPathScrapers:
       Performers:
         Name: $info//div[starts-with(@class,"info")]/a/text()
       Image: &imageSel //meta[@property="og:image"]/@content
+      Studio:
+        Name:
+          fixed: "Virtual Taboo"
```
halorrr, line 31: If we didn't want to us a fixed name `//meta[@name="author"]/@content` is available, but it doesn't have a space. I'm good with this path though.

halorrr (approved):

PR #1102, opened 2022-08-14
Update virtualrealporn network scraper

Foucoubou:
Fixed title naming  -  strip out VR Porn Video and vice versa from the other sites fro mthe network
+ added one more site VirtualRealAmateur to scraper

### DIFF
--- a/scrapers/VirtualRealPorn.yml
+++ b/scrapers/VirtualRealPorn.yml
@@ -2,13 +2,12 @@ name: "VirtualRealPorn"
 sceneByURL:
   - action: scrapeXPath
     url: &urlSel
+      - virtualrealamateurporn.com
       - virtualrealgay.com
       - virtualrealjapan.com
       - virtualrealpassion.com
       - virtualrealporn.com
       - virtualrealtrans.com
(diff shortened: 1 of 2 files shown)


Foucoubou:
Thanks for review and tweaks - will use knowledge next time.

scrapers/VirtualRealPorn.yml
```diff
@@ -15,7 +17,22 @@ movieByURL:
...
+              - regex:  VR Trans Porn video$
+                with: ""
+              - regex:  VR Gay Porn video$
+                with: ""
+              - regex:  VR Japanese video$
+                with: ""
+              - regex:  VR Female POV Porn video$
+                with: ""
+              - regex:  VR Amateur Porn video$
+                with: ""
```
halorrr, line 35:
```suggestion
          - replace:
              - regex:  VR [\s\w]+? video$
                with: ""
```

Lets simplify this down to a single regex statement. This is also more future proof if they add more sites.

If we wanted to be more precise still we could use something like `VR (?:Trans|Gay|Japanese|Female POV|Amateur)?\s?(?:Porn\s)?video$`

PR #1105, opened 2022-08-15
New Elite babes scraper

HijackHornet:
Performer by URL and name.

### DIFF
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -46,3 +46,4 @@ xPathScrapers:
         selector: $infoPiece[text() = 'Hair Color:']/../text()
       Details:
         selector: //div/p[@class='read-more-wrap']
+# Last Updated August 15, 2022
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -12,7 +12,7 @@ performerByURL:
 xPathScrapers:
   performerSearch:
     performer:
-      Name: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/span/text()
+      Name: //ul//a[contains(@href,"model/")]//text()
       URL:
         selector: //ul//a[contains(@href,"model/")]/@href
 
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -42,6 +42,7 @@ xPathScrapers:
         selector: $infoPiece[text() = 'First Seen:']/../text()
       Aliases:
         selector: //div/p[@class='read-more']/strong
+        concat: ", "
       HairColor:
         selector: $infoPiece[text() = 'Hair Color:']/../text()
       Details:
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -29,7 +29,7 @@ xPathScrapers:
         selector: $infoPiece[text() = 'Height:']/../text()
         postProcess:
           - replace:
-              - regex: .^(\d+).*
+              - regex: ^(\d+).*
                 with: $1
       Image: //div/figure/img/@src
       Country:
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -1,4 +1,3 @@
-#author Hijack Hornet
 name: EliteBabes
 performerByName:
   action: scrapeXPath
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -14,7 +14,7 @@ xPathScrapers:
     performer:
       Name: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/span/text()
       URL:
-        selector: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/@href
+        selector: //ul//a[contains(@href,"model/")]/@href
 
   performerScraper:
     common:
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -29,7 +29,7 @@ xPathScrapers:
         selector: $infoPiece[text() = 'Height:']/../text()
         postProcess:
           - replace:
-              - regex: .*\((\d+)cm\)
+              - regex: .^(\d+).*
                 with: $1
       Image: //div/figure/img/@src
       Country:
--- a/scrapers/EliteBabes.yml
+++ b/scrapers/EliteBabes.yml
@@ -13,16 +13,14 @@ xPathScrapers:
   performerSearch:
     performer:
       Name: //ul//a[contains(@href,"model/")]//text()
-      URL:
-        selector: //ul//a[contains(@href,"model/")]/@href
+      URL: //ul//a[contains(@href,"model/")]/@href
 
   performerScraper:
     common:
       $infoPiece: //div/ul[@class='list-aside hidden']/li/span
     performer:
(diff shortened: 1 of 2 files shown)


scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,49 @@
+#author Hijack Hornet
```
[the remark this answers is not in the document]
HijackHornet: Done

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,49 @@
...
+        selector: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/@href
+
+  performerScraper:
+    common:
+      $infoPiece: //div/ul[@class='list-aside hidden']/li/span
+    performer:
+      Name: //div/h2[@class='fn']/text()
+      Birthdate:
+        selector: $infoPiece[text() = 'Born:']/../text()
+      Gender: "Female"
```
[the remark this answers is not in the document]
HijackHornet: Thanks for the fix :)

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,50 @@
...
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - www.elitebabes.com/model/
+    scraper: performerScraper
+
+xPathScrapers:
+  performerSearch:
+    performer:
+      Name: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/span/text()
```
halorrr, line 15:
```suggestion
      Name: //article[@id='content']/ul[1]//span/text()
```

Slightly different strategy here that makes this much more compact. I select the first `ul` node, since all the other `ul` nodes on the page share the same class name anyway, from there we no longer need the contains and can go straight down to the first span node containing text

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,50 @@
...
+      URL:
+        selector: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/@href
```
halorrr, line 17:
```suggestion
        selector: //article[@id='content']/ul[1]//a/@href
```

Similar strategy to the above here

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,50 @@
...
+      Name: //div/h2[@class='fn']/text()
+      Birthdate:
+        selector: $infoPiece[text() = 'Born:']/../text()
+      Gender:
+        fixed: "Female"
+      Height:
+        selector: $infoPiece[text() = 'Height:']/../text()
+        postProcess:
+          - replace:
+              - regex: .*\((\d+)cm\)
```
halorrr, line 32:
```suggestion
              - regex: .^(\d+).*
```

There is a wide variety of height formats on the site: Found examples of all of these:
```
152 cm
175 cm, 5 ft 9 in cm
172
```
https://www.elitebabes.com/model/airu-kaname/
https://www.elitebabes.com/model/alisa-i/
https://www.elitebabes.com/model/mila-azul/

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,50 @@
...
+      URL:
+        selector: //article[@id='content']/ul[@class='gallery-a b d']/li/a[contains(@href,"model/")]/@href
+
+  performerScraper:
+    common:
+      $infoPiece: //div/ul[@class='list-aside hidden']/li/span
+    performer:
+      Name: //div/h2[@class='fn']/text()
+      Birthdate:
+        selector: $infoPiece[text() = 'Born:']/../text()
```
halorrr, line 25: Going to need some postProcess here to handle the birth date. Unfortunately they are not consistent

https://www.elitebabes.com/model/mango-a/ has `1989-12-11`
https://www.elitebabes.com/model/mila-azul/ has `1997`
https://www.elitebabes.com/model/alisa-i/ has `March 1997`

@bnkai I may be behind in the times, do we have a method of handling multiple possibilities of birth dates? Or is that only something a script could handle?

scrapers/EliteBabes.yml
HijackHornet: I'm not sure this is correct. In the case of a search that results in no performer found but only gallery, this would pull the gallery instead...

scrapers/EliteBabes.yml
HijackHornet: Same

scrapers/EliteBabes.yml
HijackHornet: Thanks

scrapers/EliteBabes.yml
halorrr: Good point, this one works instead:
```suggestion
      Name: //article[@id='content']/*[following-sibling::h2]//span/text()
```
Selects all the elements who's following sibling is an h2 under content. That way galleries will always be filtered out.

scrapers/EliteBabes.yml
halorrr:
```suggestion
        selector: //article[@id='content']/*[following-sibling::h2]//a/@href
```

Same new suggestion from above

scrapers/EliteBabes.yml
```diff
@@ -0,0 +1,50 @@
...
+      Gender:
+        fixed: "Female"
+      Height:
+        selector: $infoPiece[text() = 'Height:']/../text()
+        postProcess:
+          - replace:
+              - regex: .^(\d+).*
```
halorrr, line 32:
```suggestion
              - regex: ^(\d+).*
```
@HijackHornet sorry I made a typo here when I entered the suggestion in my previous comment and left that leading period on

scrapers/EliteBabes.yml
bnkai: In a couple of scrapers i think we use a series of `parseDate`. If a format doesnt match parseDate will print an error and return the original value and then continue to the next so no problem with using multiples.
The main issue from the referenced urls @halorrr is the lack of full date. Do we want to regex replace missing months/days with 01 or not? (The same issue is present in some movie scrapers). It is different having a birthdate in 1997, the birthdate actually being 1997-01-01 and the birthdate being regex replaced to 1997-01-01. 
Leaving just what is available means that the user has to decide and make a manual edit (maybe adding a comment/tag somewhere  that the data is not accurate) so that he can save.
Replacing, filling the date means that the user doesnt need to do anything but also doesnt know that the date was prefilled and thus not accurate.

scrapers/EliteBabes.yml
HijackHornet: As mentioned on discord, i disagree with this edits. It make the scraper more brittle for no reason.

scrapers/EliteBabes.yml
halorrr:
```suggestion
        selector: //ul//a[contains(@href,"model/")]/@href
```

Super short version that leaves the query in and strips out the unneeded parts of the path

scrapers/EliteBabes.yml
halorrr:
```suggestion
      Name: //ul//a[contains(@href,"model/")]//text()
```

Super short version that leaves the query in and strips out the unneeded parts of the path

scrapers/EliteBabes.yml
halorrr: Good point, I'm thinking it is safe to parse the full date format, and let anything else fall through to the user to adjust

halorrr (approved):

PR #1108, opened 2022-08-19
theNude scraper fixes

estellaarrieta:


bnkai:
Seems ok
A note on `partial` performer birth dates.
Prefilling with 01 - 01 as the date month where those are missing is probably ok, is it better though if we note somewhere that the birth date is partial?
For example adding to the details section ?
```yaml
      Details:
        selector: //li/span[@class="list-quest"][contains(text(),'Born')]/../text()
        postProcess:
          - replace:
              - regex: ^
                with: "Actual scraped Birthdate:"
```

PR #1109, opened 2022-08-19
LetsDoeIt scraper fixes

estellaarrieta:
*details xpath fix
+regex to remove Studio name at the end of a description
*image parse xpath fix

### DIFF
--- a/scrapers/LetsDoeIt.yml
+++ b/scrapers/LetsDoeIt.yml
@@ -43,7 +43,7 @@ xPathScrapers:
       Tags:
         Name: $details//div[@class="col"][4]//a/text()|$details//div[@class="col"][6]//a/text()
       Performers:
-        Name: $actors//span[position()>1]/a//text() # position >1 to ignore studio link
+        Name: $actors//span/a[contains(@href,"/models/")]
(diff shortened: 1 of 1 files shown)


scrapers/LetsDoeIt.yml
```diff
@@ -33,16 +33,18 @@ xPathScrapers:
...
           - replace:
               - regex: "&#039;"
                 with: "'"
+              - regex: (.*\.?\!?)(?:\s-\s\w.*-.*)$ # remove Studio name at the end of a description
+                with: $1
       Tags:
         Name: $details//div[@class="col"][4]//a/text()|$details//div[@class="col"][6]//a/text()
       Performers:
-        Name: $actors//span/a//text()
+        Name: $actors//span[position()>1]/a//text() # position >1 to ignore studio link
```
bnkai, line 46:
```yaml
        Name: $actors//span/a[contains(@href,"/models/")]
```
also works and is probably cleaner to use?

estellaarrieta: thanks, it is a better one, i reaplced it

PR #1114, opened 2022-08-22
Docker option to validate

Wasylq:
Dockerfile + description in README on how to use

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:current-alpine
+
+RUN mkdir -p /data
+
+COPY . /data
+WORKDIR /data
+
+RUN cd ./validator && yarn install --frozen-lockfile
+ENTRYPOINT node ./validate.js --ci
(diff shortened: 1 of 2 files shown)


Wasylq:
Is there something missing from it? Can I do something to get this merged?

bnkai:
> Is there something missing from it? Can I do something to get this merged?

It took me some time to actually test your PR
A few things that i noticed
- with your Dockerfile you would need to build the container every time you update a scraper as the scrapers are copied and checked inside the container
- the `COPY .` part when building the container copies all the root folder which in some cases can be big (for example if a user uses his stash folder and not a separate clean development folder)

I would create a  Dockerfile inside the validator folder like the below
```Dockerfile
FROM node:current-alpine

RUN mkdir -p /validator

COPY *.* /validator/
WORKDIR /validator
RUN cd /validator && yarn install --frozen-lockfile
```
Then we would only need to build the container when the validator code changes
```bash
sudo docker build -t stash_scrapper_validate validator/
```
And run the validator whenever we want to check the scrapers
```bash
sudo docker run --rm --mount type=bind,source="$(pwd)"/scrapers,target=/scrapers,consistency=delegated stash_scrapper_validate node /validator/index.js
```

feederbox826:
why even bother with a dockerfile? :)

```
❯ docker run -v .:/app node:alpine /bin/sh -c "cd /app/validator && yarn install && node index.js --ci"           
yarn install v1.22.22
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.25s.
Validation passed!
```

PR #1118, opened 2022-08-29
Update SexLikeReal.yml

Foucoubou:
Fix following two issues:
Description not beeing scraped
The name convention (The formerly scraped name - had <i>long name</i> which was not corresponding to correct titles (eg. titles checkable from other studios main pages. The new way scrapes the information from the image, where is " VR porn" additional in the end, but this is strippable.

Would be grateful for feedback & review - tested locally - should cover up the splitted description.

### DIFF
--- a/scrapers/SexLikeReal.yml
+++ b/scrapers/SexLikeReal.yml
@@ -7,12 +7,14 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: &titleDef
-        selector: //div[@id="tabs-photos"]//a/img/@alt
+      Title: 
+        selector: //script[@type="text/javascript"][contains(.,"videoData:")]/text()
         postProcess:
             - replace:
-                - regex:  " VR porn$"
-                  with: ""
+                - regex:  '.+videoData:\s{[^{]+title":"([^"]+)",.+'
+                  with: $1
+                - regex: '\\u2019'
+                  with: "’"
       Date: //time/@datetime
       Details: //div[@data-qa="scene-about-tab-text"]
       Tags:
--- a/scrapers/SexLikeReal.yml
+++ b/scrapers/SexLikeReal.yml
@@ -14,9 +14,7 @@ xPathScrapers:
                 - regex:  " VR porn$"
                   with: ""
       Date: //time/@datetime
-      Details: &detailsSel
-        selector: //div[@data-qa="scene-about-tab-text"]//text()
-        concat: " "
+      Details: //div[@data-qa="scene-about-tab-text"]
       Tags:
         Name: //meta[@property="video:tag"]/@content
       Performers:
--- a/scrapers/SexLikeReal.yml
+++ b/scrapers/SexLikeReal.yml
@@ -22,6 +22,17 @@ xPathScrapers:
       Performers:
         Name: //meta[@property="video:actor"]/@content
       Studio:
-        Name: //a[contains(@href,"/studios/")]/div[last()]/text()
+        Name: 
+          selector: //a[contains(@href,"/studios/")]/div[last()]/text()
+          postProcess:
+              - map:
+                  LustReality: "Lust Reality"
+                  DDFNetworkVR: "DDF Network VR"
+                  POVcentralVR: "POV Central"
+                  RealHotVR: "Real Hot VR"
+                  WankitnowVR: "Wank It Now VR"
+                  VirtualXPorn: "Virtual X Porn"
+                  SinsVR: "XSinsVR"
+                  LethalHardcoreVR: "Lethal Hardcore VR"
       Image: //div[contains(@class,"splash-screen")]/img/@src
 # Last Updated August 29, 2022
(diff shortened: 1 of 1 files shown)


scrapers/SexLikeReal.yml
```diff
@@ -7,15 +7,21 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: //h1/text()
+      Title: &titleDef
+        selector: //div[@id="tabs-photos"]//a/img/@alt
+        postProcess:
+            - replace:
+                - regex:  " VR porn$"
+                  with: ""
```
bnkai, line 15: The selector used matches more that one tabs and for some scenes it doesnt match anything eg `https://www.sexlikereal.com/scenes/dean-of-admissionary-8736`
I would just use the existing selector and use pp to remove the part after the `-` , i think that has the same end result but it matches for every scene i tested
```yaml
      Title: 
        selector: //h1
        postProcess:
            - replace:
                - regex:  '\s*-.+$'
                  with: ""
```
Can you paste some examples with urls and `proper` titles?

scrapers/SexLikeReal.yml
```diff
@@ -7,15 +7,21 @@ sceneByURL:
...
       Date: //time/@datetime
-      Details: //div[@id="modalSceneAbout"]//comment()[contains(.,'description')]/following-sibling::div
+      Details: &detailsSel
+        selector: //div[@data-qa="scene-about-tab-text"]//text()
+        concat: " "
```
bnkai, line 19:
```suggestion
      Details: //div[@data-qa="scene-about-tab-text"]
```
Should be enough. Stash gets the innertext of elements when needed

scrapers/SexLikeReal.yml
Foucoubou: @bnkai Initially I have been thinking to use the "-" as stripper - but it does not work in some cases.
I have primarily put my eye on SLR only scenes (eg. SLR Originals and studios, which are exclusively there)
For WankzVR there is an official site (often with different descriptions and release dates).

I have checked the data for the scene you mentioned and found that the title is also hidden in <script type="text/javascript">
   ....................
                        videoData: {"price":9.99,"id":8736,"title":"**Dean of Admissionary**",
...............

But I am not sure if it is possible to scrape from those fields (maybe yes)

The case will not work here:
https://www.sexlikereal.com/scenes/up-close-with-agatha-vega-27252 (Up Close with Agatha Vega)
https://www.sexlikereal.com/scenes/meana-wolf-pillow-fort-16389 (Meana Wolf - Pillow Fort)
https://www.sexlikereal.com/scenes/sin-fully-sexy-27568 (Sin-fully Sexy)

So I am afraid that the solution can't be that trivial :)

bnkai: @Foucoubou yeah sraping from the script is no problem
Can you try the below for the title?
```yaml
      Title: 
        selector: //script[@type="text/javascript"][contains(.,"videoData:")]/text()
        postProcess:
            - replace:
                - regex:  '.+videoData:\s{[^{]+title":"([^"]+)",.+'
                  with: $1
```
It works of for all urls mentioned above, i could use some more testing though (could break if the title includes a `"`)

Foucoubou: @bnkai will try intense testing in the evening and let you know

Foucoubou: @bnkai 
Sorry for delay...
Works on 99,9%
Cases to test more:
https://www.sexlikereal.com/scenes/whats-wet-in-granada-27758
(What\u2019s Wet in Granada)

https://www.sexlikereal.com/scenes/virtual-girlfriendvirtual-girlfriendfuck-this-beautiful-teen-jennifer-virtual-girlfriend-14816
(virtual girlfriendvirtual girlfriendFuck This Beautiful Teen Jennifer - Virtual Girlfriend)

except this ones it seems to work well.

One more thing to solve...
During testing I have found, we will need to have a map to convert the studio names in order to have the same names in db (and to match them)

bnkai: a good testing takes a long time :-)

The first issue is easy and i was hoping you could find such a case as well
we just need to add an extra regex to replace the special char like
```yaml
      Title: 
        selector: //script[@type="text/javascript"][contains(.,"videoData:")]/text()
        postProcess:
            - replace:
                - regex:  '.+videoData:\s{[^{]+title":"([^"]+)",.+'
                  with: $1
                - regex: '\\u2019'
                  with: "’"
```
The second case is the fault of the site, the scraper gets the title they gave so either leave it like that or add a mapping for just this case ?

When you say `in order to have the same names in db (and to match them)` what do you mean exactly? You mean to have the same names as stashdb? If yes then its ok to add an extra map

Foucoubou: Hi @bnkai 
Thanks for comment and patience :-)

Lets leave second case as is - the scraper probably never would be perfect :) - but if it works in 99,99999% i think it is good enough ;) If it is substudio issue we can do some fixes in the future.

Concerning the studio names
eg. Lust Reality is currently scraped as LustReality etc...
I will try to make the map of the known cases to me and add that into the commit :)

Foucoubou:
@bnkai  ready for review :)

PR #1119, opened 2022-08-29
Create TakeVan

Foucoubou:
Add scraper for page which was currently non-scrapable using URL button.
The scraper cannot parse Performers, because the performers on studio page are mentioned just by first names.
In case of performer search indexxx come handy.


scrapers/TakeVan.yml
```diff
@@ -0,0 +1,28 @@
...
+    scene:
+      Title: //h1[@class="videoTitle"]
+      Date: &dateAttr
+        selector: //div[@id="videoDate"]
+        postProcess:
+          - replace:
+              - regex:  "^Added "
+                with: ""
+          - parseDate: January 2, 2006
+      Details: &detailsSel //div[@class="videoDescription"]/text()
```
bnkai, line 19: This gets only a partial description (the first text only)
switch to 
```suggestion
      Details: //div[@class="videoDescription"]
```
to get all the description

scrapers/TakeVan.yml
```diff
@@ -0,0 +1,28 @@
...
+      Tags:
+        Name: //div[@class="videoTags"]/a/text()
+      #Performers:
+      #  Name: $info//div[starts-with(@class,"info")]/a/text()
```
bnkai, line 23:
```yaml
      Performers:
        Name:
          selector: //div[@class="videoFeaturedModels"]/text()
          split: ", "
```
Adds the performers when mentioned (single named performers are not that useful but some users may want to add them ?)

Foucoubou: Added :)

PR #1074, opened 2022-07-25
initial implementation of rb_common

halorrr:
This is my initial implementation of the rb_common shared interfaces similar to what we have available for python scrapers. I started with a 1:1 of the python design and the refactored from there to what I feel is a better long term design. Both the config and graphql have a base class so that different child classes (for example on configs I now have a separate one for Stash and StashDB, where previously there was just one for stash) can have a common API. This also worked will since `stash_graphql` is now a child class, and the base query method is in the base class so that other common api interface (maybe a stashdb one in the future) can share that method.

The graphql interface has a third party dependancy on the `faraday` gem which I have included in my PR that adds Ruby to stash https://github.com/stashapp/stash/pull/2707 (that should be included in the 0.17 release)

Within the stash graphql interface I've moved any of the written out graphql queries to private methods on the class that are class by public methods like:
```Ruby
    def configuration
      query(configuration_query)["configuration"]
    end
```

Found this made the available methods in the API much easier to read. In a future PR I will work on adding a shared interface for comparing and adding scene markers during a scrape.

I included REAME files at every level so that there is clear documentation on how everything is currently set up and how to use it.

PR #1121, opened 2022-08-30
add Thumbnail Frame Matcher scraper

mmenanno:
This scraper uses Ruby and is dependant on https://github.com/stashapp/CommunityScrapers/pull/1074

View the README included with this PR [here](https://github.com/stashapp/CommunityScrapers/blob/31ffbdf40ccf9b5bc776636303239b6c75a2d039/scrapers/thumbnail_frame_matcher/README.md) for more details on what this does.

I've tested it with a lot of scenes now, some real world examples of it's use:
- https://stashdb.org/edits/6c18cb7d-4976-4423-aeb8-17301c75e464
- https://stashdb.org/edits/679e9d7e-9384-46ff-af68-eb2c4acaac6f
- https://stashdb.org/edits/875954b1-ecee-4419-8a89-4530a96202de
- https://stashdb.org/edits/a3d02496-876c-4903-bc86-1a9071193fb6
- https://stashdb.org/edits/6d6d8b92-27af-499d-9250-0d7681b1e67d

Of note I made the minimum match percentage it could drop to 70 as in all my tests (and I threw some pretty bad files at it), the majority of scenes matches with the first attempt at 95, but the lowest that I got an accurate match for a scene was at 70. Anything below 70 was typically too off the mark to be a good replacement.


bnkai:
Only went through the code really fast, from what i understand it needs 
- ffmpeg/ffprobe to be in the path not where stash is run from ?
- it also requires `wget` to be installed that isnt mentioned in the README ? ( a little trickier for windows users  )

halorrr:
> * ffmpeg/ffprobe to be in the path not where stash is run from ?

Hmmm I've only tested it in the stash docker container and it worked without any extra config, so I think it pinks up from the stash build

> * it also requires `wget` to be installed that isnt mentioned in the README ? ( a little trickier for windows users  )

I didn't think about the windows limitation on `wget`, let me rethink that part to see if I can be done differently

halorrr:
Circling back on the `wget`. I think I can drop the need for it, but it would be best to do so after the Files Refactor. Cause dropping it involves adding logic to get the figure out the path to the screenshot file (since the FindScene query only gives me the web path), which looking at my stash currently it looks like I can get via the `oshash` but I believe this will change after the refactor. So for now I'll add a note to the Readme and iterate on this scraper after the refactor.

scrapers/thumbnail_frame_matcher/thumbnail_frame_matcher.rb
```diff
@@ -0,0 +1,168 @@
...
+
+require "open3"
+require "json"
+require 'base64'
+require 'securerandom'
+require_relative "../rb_common/rb_common"
+
+class ThumbnailFrameMatcher
+  PTS_TIME_REGEX = /pts_time:([\d\.]+)/
+  TEMPORARY_DIRECTORY = "/generated/tmp/"
```
bnkai, line 11: For users not running the docker container this directory needs to be configured

bnkai:
Seems to work ok for me in linux (not docker, had to adjust the temp directory mentioned above).

For the wget dependency can't you use plain ruby code to download the file instead ? (If windows users need to install the faraday 
gem anyway another gem wouldnt be an issue compared to getting wget and putting in the correct path ?)
Is it tested/supposed to work for windows/macos ? If not just mention that in the README

halorrr:
> Seems to work ok for me in linux (not docker, had to adjust the temp directory mentioned above).

Weird I use docker and it worked right out the box for me. I just realized though that I forgot the Generated path user user-configurable so I should be pulling it from the config with:
```GraphQL
query {
  configuration {
    general {
      generatedPath
    }
  }
}
```

I'll tweak it to use that when searching for the tmp location instead. Maybe your docker instance had a custom path?

> For the wget dependency can't you use plain ruby code to download the file instead ? (If windows users need to install the faraday
gem anyway another gem wouldnt be an issue compared to getting wget and putting in the correct path ?)

That is fair, I was avoiding adding other gem dependancies, but I also don't think I gave this enough thought. I'll take another look.

> Is it tested/supposed to work for windows/macos ? If not just mention that in the README

Only tested on docker using unraid. I'll put that in the readme

halorrr:
@bnkai I've successfully removed the `wget` dependancy. The generated path is also now pulled from the user config, can you let me know if you still have to do special setup when running this on your docker instance?

I also bundled into this a small tweak to the graphql call to stash for a scene to also pull the new `files` property.

bnkai:
@halorrr i didn't make myself clear sorry...
I tried your scraper running stash directly (debian distro) not from a container. It should work fine from a container (if the user opted for the recommended docker installation). It only needed the config change because i didnt use docker.

bnkai:
@halorrr is it meant to only work with the files-refactor build and later ? It errors out with a `31`schema version.
You can either mention that as well in the README or add an extra check (using the systemStatus query <https://github.com/stashapp/stash/wiki/API#get-system-status> ) to get the api version (atm api and db schema version are the same) and use a different function to get the fingerprint

bnkai:
The error is mentioned above
```bash
 [Scrape / Thumbnail Frame Matcher] [GraphQL] Query failed: 422 - {"errors":[{"message":"Cannot query field \"files\" on type \"Scene\". Did you mean \"file\" or \"title\"?","locations":[{"line":30,"column":15}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}},{"message":"Cannot query field \"files\" on type \"Scene\". Did you mean \"file\" or \"title\"?","locations":[{"line":30,"column":15}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}}],"data":null} 
ERRO[2022-09-25 18:30:02] [Scrape / Thumbnail Frame Matcher] /home/username/.stash/scrapers/rb_common/graphql/stash.rb:33:in `get_scene': undefined method `[]' for nil:NilClass (NoMethodError) 
ERRO[2022-09-25 18:30:02] [Scrape / Thumbnail Frame Matcher]    from thumbnail_frame_matcher.rb:170:in `<main>' 
ERRO[2022-09-25 18:30:02] could not unmarshal json from script output: EOF 
```
It is just matter of using a separate grapqhql query per version

PR #1128, opened 2022-09-11
JimWeathersArchives xPath scene/performer scraper

DogmaDragon:
Requested on Discord

### DIFF
--- a/scrapers/JimWeathersArchives.yml
+++ b/scrapers/JimWeathersArchives.yml
@@ -1,0 +1,54 @@
+name: "Jim Weathers Archives"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - jimweathersarchives.com
+    scraper: sceneScraper
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - jimweathersarchives.com
+    scraper: performerScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: 
+        selector: //div[@class='title_bar']/span/text()
+      URL: 
+        selector: //div[@class='rearrangeTab options_block active']/a[@class='option_button']/@href
+      Details:
+        selector: //span[@class='update_description']
+      Date:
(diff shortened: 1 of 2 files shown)


scrapers/JimWeathersArchives.yml
```diff
@@ -0,0 +1,54 @@
...
+          - parseDate: 01/02/2006
+      Image:
+        selector: //meta[@name="twitter:image"]/@content
+        postProcess:
+          - replace:
+              - regex: ^(.+)\-.+
+                with: $1-4x.jpg
+      Studio:
+        Name:
+          fixed: BondageCafe
```
bnkai, line 33: Is  the studio actually `BondageCafe` ?

scrapers/JimWeathersArchives.yml
```diff
@@ -0,0 +1,54 @@
...
+        URL: 
+          selector: //div[@class='gallery_info']/span[@class='update_models']/a/@href         
+  performerScraper:
+    performer:
+      Name:
+        selector: //span[@class='title_bar_hilite']/text()
+      Gender:
+        fixed: female     
+      Image:
+        selector: //img[@id='set-target-13']/@src0_3x
```
bnkai, line 49: the `set-target-xx` is not always 13 (eg `https://www.jimweathersarchives.com/store/models/DianaKnight.html`, `https://www.jimweathersarchives.com/store/models/KendraJames.html`), i suggest switching to something like 
```suggestion
        selector: //img[contains(@class,'model_bio_thumb')]/@src0_3x
```
instead

scrapers/JimWeathersArchives.yml
DogmaDragon: Based on the person who requested the scraper, yes. 
https://discord.com/channels/559159668438728723/651418567475986432/1014170965019729981

scrapers/JimWeathersArchives.yml
DogmaDragon: Noticed that and fixed it on my previous version, but it got lost during some other testing. Thanks for catching it.

PR #1129, opened 2022-09-11
BrandNewAmateurs xPath scene scraper

DogmaDragon:
Minimum information is available on the site. 

Requested here https://github.com/stashapp/CommunityScrapers/issues/73#issuecomment-1175113261

### DIFF
--- a/scrapers/BrandNewAmateurs.yml
+++ b/scrapers/BrandNewAmateurs.yml
@@ -1,0 +1,22 @@
+name: "Brand New Amateurs"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - brandnewamateurs.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: 
+        selector: //meta[@property="og:title"]/@content
+      Details:
+        selector: //div[@class='videoDetails clear']/p
+      Image:
+        selector: //meta[@property="og:image"]/@content
+      Studio:
+        Name:
+          fixed: Brand New Amateurs
+      Tags:
+        Name:
+          selector: //div[@class='featuring clear']/ul/li/a    
+# Last Updated September 11, 2022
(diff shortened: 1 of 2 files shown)


scrapers/BrandNewAmateurs.yml
```diff
@@ -0,0 +1,22 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - brandnewamateurs.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: 
+        selector: //meta[@property="og:title"]/@content
```
bnkai, line 11: In some scenes the content of the title contains a sort of `composite` title. Eg `https://brandnewamateurs.com/trailers/Sadie-Holmes-Pops-By.html` or `https://brandnewamateurs.com/trailers/Ashley-Wolf-BJ-InstructionAshley-Wolf-BJ-Instruction.html`
Adding a pp like
```yaml
        postProcess:
          - replace:
              - regex: ";;.+$"
                with:
``` 
clears the extra part at least for the above sample urls i found. 
For some scenes like `https://brandnewamateurs.com/trailers/Dixie-Normous-Returns.html` is it ok to remove the second part (after the `;;`) ?

DogmaDragon: I think the video title is the second part. since it matches what `//div[@class='videoDetails clear']/h3` shows for titles. 
E.g. `https://brandnewamateurs.com/trailers/Dixie-Normous-Returns.html`

Not using the h3 as a selector because only newer ones have it, older scenes hide it so that's why I went with meta. 

But I don't know what's best in this case. The first part provides a performer, while the second part looks more like a title for me.

bnkai: The fact that the h3 was missing in some older scenes made me think that the title was the first part :-(... I think we should just leave the `content` untouched and let users decide how to edit the title...

PR #1136, opened 2022-09-21
Scrape filename to scene title

o-fl0w:
Get scene filename from stash instance to set as scene title, i.e. sync external changes to filenames back to stash as scene titles.

edit:
Optionally remove extension from title:
set `REMOVE_EXT = True` in `Filename.py`

- [x] uses graphql client from py_common
- [x] uses config from py_common


scrapers/Filename.py
```diff
@@ -0,0 +1,31 @@
+import json
+import os
+import sys
+
+from py_common import graphql
+from py_common import log
```
bnkai, line 6: we use something like this to be more user friendly
```suggestion
try:
    from py_common import graphql
    from py_common import log
except ModuleNotFoundError:
    print(
        "You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)",
        file=sys.stderr)
    sys.exit()
```

bnkai:
Looks good...
Maybe add an option to remove the file extension as well ? 
eg

```py
import json
import os
import sys

from py_common import graphql
from py_common import log

REMOVE_EXT = False # remove file extension from title
.
.
.
.
    if REMOVE_EXT:
        filename = os.path.splitext(filename)[0]
    if scene_title != filename:
        log.info(f"Scene {scene_id}: Title differs from filename: '{scene_title}' => '{filename}'")
        return {"title": filename}
.
.
.
```

PR #1137, opened 2022-09-21
ThePornDB.yml add queryURLReplace

MrX292:
added queryURLReplace to performerByURL and sceneByURL to enable scrapping by using https://metadataapi.net/scenes/ and https://metadataapi.net/performers/


scruffynerf (approved):
looks good...
works with old (uuid) and new urls (slugs)

We need to remove the obsolete PornDB.yml scraper too.

PR #1145, opened 2022-09-27
[ThePornDB/Performer] Add details, birthdate, country, haircolor and weight

TgSeed:
Adds `Details`, `Birthdate`, `Country`, `HairColor` and `Weight` to `ThePornDB`'s performer scraper
Weird that these items has been missing.


bnkai:
> Weird that these items has been missing.

Yeah i dont remember why... Probably i didnt check the performer scraper that much since i use rpdb mainly for scenes ?
Anyway  i removed the extra performer stuff from the scene scrapers as when scraping scenes with multiple performers there can be issues if not all attributes are present

PR #1156, opened 2022-10-12
Adding scraper for transmodeldatabase.com

kozobot:
Supports performerByName and performerByUrl.  

tramsmodeldatabase only supports a mixed performer and scene search.  The performer search filter is based on the tease-model class.  Only the first page of results is processed.


bnkai:
Only known issue is with partial birth dates. When year is not available a non valid date is returned eg `30 November`. Imo it is fine as it is and the user should decide what to do with it. (for example adding a tag or adding an extra text in the details field)

Merging this.

PR #1158, opened 2022-10-14
Fix Reality Kings scraping

hphpanon:
RealityKings hasn't been scrape-able in a bit. After doing some testing, it appears to Just Work:tm: with the MindGeek scraper.

### DIFF
--- a/scrapers/MindGeek.yml
+++ b/scrapers/MindGeek.yml
@@ -199,7 +199,7 @@ xPathScrapers:
 
   performerScraper:
     performer:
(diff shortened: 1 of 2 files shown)


scrapers/MindGeek.yml
```diff
@@ -58,6 +59,7 @@ performerByURL:
       - lookathernow.com/model/
       - mofos.com/model/
       - mofosnetwork.com/model/
+      - realitykings.com/scene/
```
bnkai, line 62: it should be 
```yaml
      - realitykings.com/model/
```
but double check that the layout is the same. I think it needs some minor adjustments to work correctly

PR #1160, opened 2022-10-28
Fix JavHD scraper

bnkai:


Tanarion40:
Hello.

This scrapper is not working. Not even showing in my list of scrappers

Maista6969:
> Hello.
> 
> This scrapper is not working. Not even showing in my list of scrappers.
> 
> Would it be possible to fix it ?

Just tested this and it works fine: it's an URL scraper and some of the patterns were outdated, but I've pushed an update so you can pull the newest version and it should work with both the new and old style URLs

Tanarion40:
Thank you.
Its working well :)

PR #1164, opened 2022-11-04
[AdultPrime] Add search

Flashy78:
The images on the search results page are lazy loaded, and only 4 at a time. In order to use CDP to click through them, it takes a long time, so it's commented out by default.

Also fixed the scene scraper to correctly find the date, skip the image if it's their placeholder and add the URL and details.

### DIFF
--- a/scrapers/AdultPrime.yml
+++ b/scrapers/AdultPrime.yml
@@ -62,8 +62,6 @@ xPathScrapers:
           - replace:
             - regex: .*=
               with: "https://adultprime.com/studios/video/"
-      Details:
-        selector: //p[contains(@class, 'ap-limited-description-text')]/text()
       Studio: 
         Name: //p[contains(@class,"update-info-line")]/b/a[contains(@href,"/studio/")]/text()
       Tags:
--- a/scrapers/AdultPrime.yml
+++ b/scrapers/AdultPrime.yml
@@ -1,26 +1,70 @@
+# The search by name does not return images by default. To get images you will need to uncomment the driver section at the end of this file.
+# Each click you uncomment will load another 4 images, however the search will be much slower, so it's disabled by default.
 name: "AdultPrime"
+sceneByName:
+  action: scrapeXPath
+  queryURL: "https://adultprime.com/studios/search?q={}"
+  scraper: sceneSearch
+
+sceneByQueryFragment:
+  action: scrapeXPath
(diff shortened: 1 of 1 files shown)


scrapers/AdultPrime.yml
```diff
@@ -1,31 +1,84 @@
...
+          - replace:
+              - regex: .+url\("(.+)"\).+
+                with: $1
+
   sceneScraper:
     scene:
       Title: //h2[contains(@class,"update-info-title")]/text()
       Date:
-        selector: //p[@class="update-info-line regular"][i[contains(@class,"fa-calendar")]]/b[1]/text()
+        selector: //p[@class="update-info-line regular"][1]/b[1]/text()
```
bnkai, line 48: were there cases that the original was not working ?

scrapers/AdultPrime.yml
```diff
@@ -1,31 +1,84 @@
...
           - replace:
               - regex: .+url\((.+)\).+
                 with: $1
-      Studio:
+      URL:
+        selector: //a[contains(@href, "/signup?galleryId")][1]/@href
+        postProcess:
+          - replace:
+            - regex: .*=
+              with: "https://adultprime.com/studios/video/"
```
bnkai, line 64: The id seems to be present in a seperate div as well
```yaml
      Code:
        selector: //div[@class="keys"]/@title
        postProcess:
          - replace:
            - regex: .+/(\d+)$
              with: $1
```
Not sure if its safer than the signup selector. Can you add the `Code` part though ( any of the selectors is fine)

scrapers/AdultPrime.yml
Flashy78: Was a long time ago, so not sure, but I'm guessing there was a scene with no date, and so it tried to load whatever the next field was as the date.

scrapers/AdultPrime.yml
Flashy78: Added code, although I don't see it in my stash anywhere to know if it worked.
Couldn't get the url correct with the keys selector, so left it as the signup link.

Flashy78: Looks like code is not supported in the CI/CD validation?

bnkai: Yeah sorry the validator is not updated yet.
Both studio code/ director fields are only in dev version

PR #1166, opened 2022-11-10
Add date to SpankBank

JackDawson94:
Update to Spankbang to include scrapping upload date, and fix on the Studio

### DIFF
--- a/scrapers/SpankBang.yml
+++ b/scrapers/SpankBang.yml
@@ -21,8 +21,7 @@ xPathScrapers:
         selector: //script[contains(text(),'uploadDate')]/text()
         postProcess:
           - replace:
-            - regex: .*uploadDate":\s"(\d{4}-\d{2}-\d{2}).*
-              with: $1
+              - regex: .*uploadDate":\s"(\d{4}-\d{2}-\d{2}).*
+                with: $1
           - parseDate: 2006-01-02
-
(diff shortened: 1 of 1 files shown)


scrapers/SpankBang.yml
```diff
@@ -11,11 +11,18 @@ xPathScrapers:
       Details: //meta[@name='description']/@content
       Image: //div[@class='play_cover']/img/@src
       Studio:
-        Name: //li[@class='us']/a[@class='ul'][contains(@href,'channel')]/text()
+        Name: //li[@class='us']/a[@class='ul']/text()
```
bnkai, line 14: The channel was there to make sure no  plain `users`  where added as studios.
We were considering studios only channels(<https://spankbang.com/channels>)
I dont use the scraper and i think sure most users would scrape from the original sources instead so i dont have an issue with this change if that is the intended usage?

JackDawson94: At least in my case it's the original creators posting there, so the "user" is the studio.
Not sure if this will create issues for others though.

JackDawson94: @bnkai  Spankbang is one of the homes of the PMV community, in this case it is the original source.
So getting as much data as possible from them will definitely be useful 😊

PR #1167, opened 2022-11-10
Adding PMV creator NoodleDude's website

JackDawson94:
New scraper for https://www.noodledude.io/

### DIFF
--- a/scrapers/Noodledude.yml
+++ b/scrapers/Noodledude.yml
@@ -11,15 +11,15 @@ xPathScrapers:
         selector: //meta[@property='og:title']/@content
         postProcess:
           - replace:
-            - regex: (.*) \| NoodleDude PMVs$
-              with: $1
+              - regex: (.*) \| NoodleDude PMVs$
+                with: $1
       Details:
         selector: //div[@class="description padding-medium"]/div[contains(@class, "rich-text-block-2") and not(contains(@class, "w-condition-invisible"))]//p
         concat: "\n\n"
       Image: //meta[@property='twitter:image']/@content
       Studio:
         Name:
-            fixed: NoodleDudePMV
+          fixed: NoodleDudePMV
       Date:
         selector: //script[contains(., 'ownerEmail')]/text()
         postProcess:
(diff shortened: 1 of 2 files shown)


scrapers/Noodledude.yml
```diff
@@ -0,0 +1,36 @@
...
+              - regex: .+(?:"created_at":\s")([^"]+).+
+                with: $1
+              - regex: (.+)T.+
+                with: $1
+          - parseDate: 2006-01-02
+      Tags:
+        Name:
+          selector: //div[@class='tags-wrapper']//div[contains(@class, 'tag-list') or @class='other-tag-items']//text()
+          concat: ","
+          split: ","
```
bnkai, line 34: The other tags end up having an extra space due to the split...
Either use the comma with the space (if the tags are consistent along all releases)
```yaml
          concat: ", "
          split: ", "
```
or do a postprocessing to remove extra spaces

scrapers/Noodledude.yml
```diff
@@ -0,0 +1,36 @@
...
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title:
+        selector: //meta[@property='og:title']/@content
+        postProcess:
+          - replace:
+            - regex: (.*) \| NoodleDude PMVs$
+              with: $1
+      Details: //meta[@name='description']/@content
```
bnkai, line 16: The description from the meta ends up having the  related tags text at  the end (and we already have the tags anyway)
Do we want to postprocess and remove them?
Do we care for the extra info as voice over and music?
Something like
```yaml
      Details:
        selector: //div[@class="description padding-medium"]/div[@class="rich-text-block-2 text-rich-text w-richtext"]//p
        concat: "\n\n"
```
seemed to work, didnt test with all pmvs though

scrapers/Noodledude.yml
JackDawson94: Used your suggestion, added postprocessing, seems to be working fine now :)

scrapers/Noodledude.yml
JackDawson94: Info about voice / music is useful to have for PMVs yeah.
Your query didn't work for all links, I had to rework it a bit but now it properly grabbed the description for all my test cases.

PR #1176, opened 2022-11-14
Update graphql.py

Tweeticoats:
the chapters_vtt field is no longer in the paths data for scenes so any call to get scenes now fails. We should remove this from the request as this is breaking any scrapers that use this python library to query stash such as the onlyfans scraper.


A-Random-Mf:
Thank you

PR #1182, opened 2022-11-25
Add scene scraper for (ecchi.)iwara.tv

xx790:
sceneByFragment relies on the default filename format from Iwara:
`<unix_timestamp>_<id>_<resolution>.<ext>`

Example file names:
`1543383579_G5mzkTeE5i7AJoe8_Source.mp4`
`1550374900_z028Qt4K7GIkyYbAD_540.mp4`

### DIFF
--- a/scrapers/Iwara.yml
+++ b/scrapers/Iwara.yml
@@ -1,0 +1,92 @@
+name: "Iwara"
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - iwara.tv
+    scraper: sceneScraper
+
+sceneByFragment:
+  action: scrapeXPath
+  scraper: sceneScraper
+  queryURL: https://ecchi.iwara.tv/videos/{filename} # also works for sfw videos
+  queryURLReplace:
+    filename: # expects the default filename format when saved from Iwara - <some digits>_<id>_<resolution>.<ext>
+      - regex: '.*_([0-9a-zA-Z]{12,})_.*'
+        with: $1
+      - regex: .*\.[^\.]+$ # if no id is found in the filename
+        with: # clear the filename so that it doesn't leak
+
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://ecchi.iwara.tv/search?query={}
+  scraper: sceneSearch
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
+  scraper: sceneScraper
+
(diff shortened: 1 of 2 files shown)


xx790:
I wish it were possible to set studio logo from scraper along with name and URL.
Would save some time in case of Iwara.
Was it proposed before? In which repo should I open an issue otherwise?

scrapers/Iwara.yml
```diff
@@ -0,0 +1,92 @@
...
+              - regex: '.+(\d{4}-\d{2}-\d{2}).+'
+                with: $1
+          - parseDate: '2006-01-02'
+  sceneScraper:
+    common:
+      $infoNode: //div[@class="node-info"]
+    scene:
+      URL: //link[@rel="canonical"][contains(@href,"http")]/@href
+      Title: $infoNode//h1/text()
+      Details: $infoNode/div[contains(@class,"field-name-body")]/div/div
```
bnkai, line 67: Using your previous selector with a concat retains the details formatting a bit better
```yaml
      Details:
        selector: $infoNode/div[contains(@class,"field-name-body")]//p
        concat: "\n"
```

bnkai:
> I wish it were possible to set studio logo from scraper along with name and URL. Can't see it in the docs.
Would save some time in case of Iwara.
Was it proposed before? In which repo should I open an issue otherwise?

AFAIK studio images are not scrapable and there is no standalone studio scraper support atm (and is probably low priority). You could use a script to call some graphql to update the studio info.

For the formatting can you go down to the `text()` or `<p>` level and do some concats as needed (that should cover most of the cases ). Stash tries to clean a bit the input after getting the innertext for most elements and it works well for most things, only the description/details part needs a bit of more `manual` pp in some cases.

xx790:
> do some concats as needed

Is it possible with bare yml and XPath 1.0 (unless higher version is actually supported)?
I would like to keep away from Python for as long as I can.

scrapers/Iwara.yml
xx790: Dang. This is what I was guessing about in another PR. `concat` should be better documented.

xx790: This should be solved.
`//p//*/text()` seems to be the maximum that can be done with this system to extract text from arbitrary html.
In the long term it might be beneficial to add something like a htmlToText or htmlToMarkdown function.

PR #1183, opened 2022-11-25
Add scene scraper for oreno3d.com

xx790:
oreno3d is a more polished frontend for ecchi.iwara.tv with some more detailed information.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -904,7 +904,7 @@ onlyteenblowjobs.com|GammaEntertainment.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 oopsie.com|Algolia_Adultime.yml|:heavy_check_mark:|:x:|:x:|:x:|Python|-
 openlife.com|GammaEntertainment.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 oraloverdose.com|PervCity.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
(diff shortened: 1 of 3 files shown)


xx790:
An issue that is not handled well currently:
`<br>` tags in the details are lost.
It would require either a Python scraper or a better HTML to text transformation within stash.

bnkai:
> There is potProcess/concat but it has no string parameter like split, and it is applied before replace so can't insert newlines either way even if I can capture all text() fragments.And XPath 1.0 also can't use concat function in the required way.

Thats why stash has its own post proccesing functions. The help section in stash  `? -> Metadata scraping -> Scraper Development` has more info and should be up to date (compared to other info in the README or wiki)

xx790:
> The help section in stash `? -> Metadata scraping -> Scraper Development` has more info and should be up to date (compared to other info in the README or wiki)

`concat` is still under-described.

xx790:
> In general when do you do a search via an xPath or JSON scraper it is better to only return result fields that you are sure they exist eg title/url and maybe image/date.

Iwara/oreno3d always have creator specified. They kinda built around it. And I think it is rather important there.

bnkai:
For the details you can use something like
```yaml
      Details:
        selector: $article//blockquote[@class="video-information-comment"]//text()
        concat: "\n\n"
```
or even with `concat: "\n"`
I am not sure which is closer to the missing `<br>`

xx790:
Just pushed the commit. One line break seems to be fine.

PR #1184, opened 2022-11-28
Add scrapping by official filename to a few scrappers

puc9:
Some studios have a filename descriptive enough that we can build the scrapping url from it and then use the existing scrape by url.
To use simply pick the scrapper either when identifying or from the dropdown in the scene editor

### DIFF
--- a/scrapers/VirtualRealPorn.yml
+++ b/scrapers/VirtualRealPorn.yml
@@ -37,6 +37,11 @@ sceneByFragment:
         with: "-"
   scraper: sceneScraper
 
+movieByURL:
+  - action: scrapeXPath
+    url: *urlSel
+    scraper: movieScraper
+
 xPathScrapers:
   sceneScraper:
     scene:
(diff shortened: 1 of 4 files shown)


scrapers/VirtualRealPorn.yml
```diff
@@ -49,19 +73,6 @@ xPathScrapers:
...
-        Synopsis: *detailsSel
-        URL: //meta[@property="og:url"]/@content
-  movieScraper:
-    movie:
-      Name: *movieName
-      Date: *dateAttr
-      Studio:
-        Name: *studioName
-      Synopsis: *detailsSel
-      FrontImage: *imageSel
```
bnkai, line 66: I think this was used for users that want to create a movie from the scene page.
That is have the scene also as movie.
Do we need to remove this?

puc9: Probably we do not _need_ to, but if left there then the movie will automatically be added when running identify since now it is possible to do so as it is no longer just an URL scrapper. Yes, the settings for identify can be tweaked to not import the movie but for some other scrappers we do actually want the movie info.

My assumption is that the users wanting the same info for both the scene and as a movie are in the minority, but I will leave it up to you, feel free to restore and/or comment it out if you think it's better.

bnkai: Should we comment out then lines 52-58 and leave the rest as is?
The movieScraper part will only used when scraping by URL and shouldnt interfere with the identify task

puc9: Let me play with it for a bit to see what could be done.

PR #1189, opened 2022-12-06
Fix GammaEntertainment movie cropped images

### LINKED ISSUE #1188: Front image of GammaEntertainment movies cropped
**Describe the bug**
Front image of GammaEntertainment movies is cropped at the top and bottom when scraping by URL

**To Reproduce**
1. Go to a GammeEntertainment movie (e.g. EvilAngel)
2. Scrape with URL
3. Scraped front cover image is cropped

**Screenshots**
image

I am using the latest version of community scrapers.

Possible fix:
Add ``?format=webp`` suffix for .jpg URL.

whadup93:
Closes #1188

``?format=webp`` missing in front image request. Without this parameter, image gets cropped at the top and bottom.

### DIFF
--- a/scrapers/Algolia.py
+++ b/scrapers/Algolia.py
@@ -464,9 +464,9 @@ def parse_movie_json(movie_json: dict) -> dict:
     scrape["date"] = movie_json[0].get(date_by_studio)
 
     scrape[
-        "front_image"] = f"https://transform.gammacdn.com/movies{movie[0].get('cover_path')}_front_400x625.jpg?format=webp"
+        "front_image"] = f"https://transform.gammacdn.com/movies{movie[0].get('cover_path')}_front_400x625.jpg?width=450&height=636"
     scrape[
-        "back_image"] = f"https://transform.gammacdn.com/movies{movie[0].get('cover_path')}_back_400x625.jpg"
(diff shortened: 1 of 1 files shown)


bnkai:
> Closes #1188
> 
> `?format=webp` missing in front image request. Without this parameter, image gets cropped at the top and bottom.

Can you provide a few url samples? And is the issue present in back images as well?

whadup93:
> > Closes #1188
> > `?format=webp` missing in front image request. Without this parameter, image gets cropped at the top and bottom.
> 
> Can you provide a few url samples? And is the issue present in back images as well?

Apparently this problem occurs only with older movies.

E.g. working URL ``https://transform.gammacdn.com/movies/29694/29694_oral_angels_front_400x625.jpg?width=450&height=636&format=webp``

You also get the full image with only on of the two extra parameters, so ``.jpg?width=450&height=636`` works as fine as ``.jpg?format=webp`` suffix.

Without my suggested fix only the ``.jpg`` is getting requested which results in the cropped front image (back is fine).
Bad URL with cropped image lookes like this ``https://transform.gammacdn.com/movies/29694/29694_oral_angels_front_400x625.jpg``

bnkai:
@whadup93 i just remembered that the same issue was mentioned in the older (non api) gamma scraper. 
```
              # if using the transport subdomain, parameters need to be passed
              # otherwise a cropped square image is returned by default
```
Can you use the width height instead of the webp? I think it is better for compatibility issues if we get the jpg image

PR #1192, opened 2022-12-07
Add Code selector for LegalPorno.yml

DogmaDragon:
Regex examples https://regex101.com/r/RHyMTe/2


bnkai:
If the CODE is always at the end of the title a single regex should be enough. 
Something like `.+?([A-Z]{2,3}\d+)$` ? (assuming code is 2 or 3 chars plus the numbers)
<https://regex101.com/r/k3bsN4/1>

PR #1193, opened 2022-12-07
Torrent scraper rewrite v1

EsEnZeT:
This v1 (pure python) rewrite of torrent.py scraper proposed as a temporary solution.

The final solution will be Cython based torrent reader which is far more faster.

Torrent files are being read from: /root/.stash/torrents

### DIFF
--- a/scrapers/torrent.py
+++ b/scrapers/torrent.py
@@ -1,66 +1,96 @@
-import os
 import sys
 import json
+from os import walk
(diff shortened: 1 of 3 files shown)


EsEnZeT:
Finished v2 -- https://github.com/stashapp/CommunityScrapers/pull/1198
Because of this I'm closing current PR.

PR #1196, opened 2022-12-11
Add Rule34Video xPath scene scraper

bnkai:
The search by name only returns the results from the first page


DogmaDragon:
Verified everything works.

PR #1198, opened 2022-12-12
Torrent scraper rewrite v2

EsEnZeT:
This is v2 (Cython) rewrite of torrent.py scraper proposed as a final solution.

Personal tests with over 10k .torrent files between v1 and v2 resulted in over 4x speedup after switch to Cython bencode reader. 

Torrent files are being read from: /root/.stash/torrents


JackDawson94:
@EsEnZeT  Had not seen your rework of Torrent scraper.
I made an upgrade as well here, also supporting searching for the right torrent file with `sceneByName `

Maybe worth merging that small part into your change and cancelling mine ?

EsEnZeT:
> @EsEnZeT Had not seen your rework of Torrent scraper. I made an upgrade as well here, also supporting searching for the right torrent file with `sceneByName ` #1206
> 
> Maybe worth merging that small part into your change and cancelling mine ?

That would be great. Maybe you can use my branch to implement that, then I could review and cherry-pick your work here if needed.

What's your disscord handle?

JackDawson94:
@EsEnZeT  Haven't found out how I can commit to the PR directly...
I submitted some changes to your repo as PR https://github.com/EsEnZeT/CommunityScrapers/pull/1

bnkai:
@EsEnZeT can you move this v2 scraper to a separate folder?
I think it is probably better to keep the original (or your v1) for compatibility issues

bnkai:
> @bnkai do you have anything in mind regarding potential issues? We are currently working on update to that version with @JackDawson94 on my repo. Original plan was to keep torrents folder where it was so I think there won't be any benefits from leaving old version.
> 


My main concern was <bencoder.pyx> ? Is it available for all platforms ?
The official docker container should be ok and installing in linux is no issue are windows/macos ok though?
If yes then ignore what i wrote. 

> PR link: [EsEnZeT#1](https://github.com/EsEnZeT/CommunityScrapers/pull/1)
Yeah it is probably better to merge that first and then make a new PR here? I'll have a look over there as well

EsEnZeT:
> > @bnkai do you have anything in mind regarding potential issues? We are currently working on update to that version with @JackDawson94 on my repo. Original plan was to keep torrents folder where it was so I think there won't be any benefits from leaving old version.
> > 
> 
> 
> My main concern was <bencoder.pyx> ? Is it available for all platforms ?
> The official docker container should be ok and installing in linux is no issue are windows/macos ok though?
> If yes then ignore what i wrote. 
> 
> > PR link: [EsEnZeT#1](https://github.com/EsEnZeT/CommunityScrapers/pull/1)
> 
> Yeah it is probably better to merge that first and then make a new PR here? I'll have a look over there as well

It should be fine for most of the platforms. Tested on Linux and macOS. Additionally the change addressed some time ago in Dockerfile adds bencoder.pyx to the app container.

@bnkai you can ignore this PR because it changes torrents directory. I will create new one later with @JackDawson94 changes and we can go from there.

scrapers/torrent.py
```diff
@@ -1,66 +1,139 @@
...
-    else:
-        query(title)
-    print(json.dumps(fragment))
-# Last Updated February 18, 2021
+    print(json.dumps(process_torrents(get_scene_data(fragment))))
+elif sys.argv[1] == "fragment":
+    filename = json.loads(sys.stdin.read()).get('url')
+    with open(filename, 'rb') as f:
+        torrent_data = bdecode(f.read())
+        print(json.dumps(get_torrent_metadata(torrent_data)))
```
bnkai, line 122: Is the url supposed to hold the actual filename?

scrapers/torrent.py
```diff
@@ -1,66 +1,139 @@
...
+    ratios = {}
+    for t in torrents:
+        clean_t = cleanup_name(t)
+        ratios[round(10000*(1-similarity_file_name(search, clean_t)))] = {'url': str(t.absolute()), 'title': clean_t}
+
+    # Order ratios
+    ratios_sorted = dict(sorted(ratios.items()))
+    # Only return the top 5 results
+    if len(ratios) > 5:
+        ratios = ratios_sorted[5:]
```
JackDawson94, line 135: Small error here, can't fix it myself anymore. This and the next line should be replaced with:
`if len(ratios) > 5:
        ratios_sorted = list(ratios_sorted.values())[0:5]
    else:
        ratios_sorted = ratios_sorted.values()
    
    print(json.dumps(list(ratios_sorted)))`

scrapers/torrent.py
JackDawson94: Required because after a user selects a scene in the `sceneByName` search, the filename is passed in the url variable.

I think I did some testing back then to use another field, but only url actually worked, but it's been a while ^^

bnkai:
@EsEnZeT sorry for the late reply. I wasn't sure if it is finished. Is this fix <https://github.com/stashapp/CommunityScrapers/pull/1198/files#r1073815909> applied? It should be ok after that?

bnkai:
> @bnkai no worries, you are right, after @JackDawson94 stopped responding I guess our best bet is testing this again locally and implementing what he wrote. I'm unfortunately bit busy lately, but when I'll find time I'll try to setup some local env and test as much as I can and report back!
I will be away from my pc for a couple of weeks so i will not be able to actually test. The code looks ok though so when you merge JackDawson94's PR and it tests ok with you let me know and i can merge

JackDawson94:
Unfortunately my fix is still waiting for a merge... https://github.com/EsEnZeT/CommunityScrapers/pull/2

PR #1201, opened 2022-12-14
Add scene scraper for javdb.com for chinese

renmu123:


feederbox826:
Closing this as it's identical to javdb, just with a cookie value of zh rather than en and a broken second site

PR #1202, opened 2022-12-15
Update DreamTranny.yml

nymeras:
Corrections to work with site update.

### DIFF
--- a/scrapers/DreamTranny.yml
+++ b/scrapers/DreamTranny.yml
@@ -7,15 +7,15 @@ sceneByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: //div[@class="updateInfo"]/h2/text()
+      Title: //div[@class="section-title"]/h4/text()
       Details:
-        selector: //div[@class="updateDescription"]/p/text()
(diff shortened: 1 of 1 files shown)


DogmaDragon:
Verified everything works.

nrg101:
I just tried this with a scene, https://dreamtranny.com/update/840 , and the image scraped is not the cover image of _this_ scene, it is the cover image for the one listed below in the More Scenes section

nrg101:
Here is what works for me to scrape the actual cover image:
```
      Image:
        selector: //video[contains(@class,'video-js')]/@poster
        postProcess:
          - replace:
            - regex: ^
              with: "https://dreamtranny.com"
```

nymeras:
> Here is what works for me to scrape the actual cover image:
> 
> ```
>       Image:
>         selector: //video[contains(@class,'video-js')]/@poster
>         postProcess:
>           - replace:
>             - regex: ^
>               with: "https://dreamtranny.com"
> ```

Nice catch, thank you. The original line worked on some scene pages, but not all. Updated.

PR #1204, opened 2022-12-15
Add studio code to Pornbox scrapper

ryosaeba75:


bnkai:
While the code part works the actual code we get is from the pb site and not from the original studios (if available).
I guess it makes sense for users that source material only from pb?
For example <https://pornbox.com/application/watch-page/234717>
I dont think pb urls are accepted in stashdb so it shouldnt be an issue ? @ryosaeba75 @DogmaDragon ?

DogmaDragon:
At the moment StashDB uses studio-provided unique code as per general consensus, but there are [ongoing discussions](https://github.com/stashapp/StashDB-Docs/issues/4) on how to deal with them in the future. 

[234717](https://pornbox.com/application/watch-page/234717) would be considered PornBox code, not studio provided-code.

I would probably hold off on using this selector until the guidelines are finalized either way. 

In the meantime, this would grab the studio code (same as used on [LegalPorno.yml](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/LegalPorno.yml)). 
```yaml
      Code: 
        selector: scene_name
        postProcess:
          - replace:
              - regex: .+?([A-Z]{2,3}\d+)$|(.+)
                with: $1
```

bnkai:
> I would probably hold off on using this selector until the guidelines are finalized either way.
Yeah thats probably better for now. 

> In the meantime, this would grab the studio code (same as used on [LegalPorno.yml](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/LegalPorno.yml)).
The selector works for LP sites could there be false matches for scenes from other networks though? Anyway it is probably better to hold this off for now as mentioned.

DogmaDragon:
Their naming scheme is pretty unique and they cross-post most of their stuff to either analvids.com or pissvids.com. 

But I'm good either, it's what I'm gonna use locally.

PR #1200, opened 2022-12-14
Add more Algolia (Being Trans 24/7, Burning Angel, Caught Fapping, DevilsFilm, DevilsTgirls, GenderXFilms, Muses, Pure Taboo, Transfixed, TransgressiveFilms, We Like Girls)

nrg101:
This moves the following sites to Algolia_Adultime:

- devilsfilm.com
- devilstgirls.com
- puretaboo.com

And adds the following sites to Algolia_Adultime:

- accidentalgangbang.com
- adulttimepilots.com
- beingtrans247.com
- burningangel.com
- caughtfapping.com
- forbiddenseductions.com
- muses.com
- officemsconduct.com
- transfixed.com
- transgressivefilms.com
- welikegirls.com

And moves the following sites to its own Algolia YAML

- genderxfilms.com

PR #1207, opened 2022-12-16
add pansexualx.com and transgressivexxx.com to Algolia_EvilAngel

nrg101:
This adds two sites to the existing Algolia_EvilAngel scraper:

- pansexualx.com
- transgressivexxx.com

Both are sub-studios of [EvilAngel](https://stashdb.org/studios/0f1d8e2c-627e-4d05-a52e-c821b26cab0c) at stashdb.org:

- [PansexualX](https://stashdb.org/studios/daee68aa-5261-4bac-af6d-e04998b78704)
- [TransgressiveXXX](https://stashdb.org/studios/45a72078-08f9-4707-8b91-02c27ebf8a99)

### Note on `galleryByURL`

You can use the corresponding scene video URL of a photo set to populate the details for a gallery. It provides all the info including performers, date, tags, title, details/description, and worth adding here even though it doesn't support the more intuitive `DOMAIN/en/photo/` URL pattern of a studio photo set page


nrg101:
CommunityScrapers-validate-improve-evilangel

bnkai:
>Note on galleryByURL

>You can use the corresponding scene video URL of a photo set to populate the details for a gallery. It provides all the info including performers, date, tags, title, details/description, and worth adding here even though it doesn't support the more intuitive DOMAIN/en/photo/ URL pattern of a studio photo set page

Sites that use the algolia api have a separate query for photosets. Is it worth it to implement a proper gallery support instead?
Some sites have browsable links of the photos while other only join links. Thats the only issue/problen i can think of ( getting the actual urls of the photosets )
For a couple of scenes/photosets i  browsed it seems most meta is indeed the same with some photosets having a description that the scenes lack.

nrg101:
> Sites that use the algolia api have a separate query for photosets.

I couldn't find a `galleryByUrl` example in any of the scrapers that contain "Algolia.py"

> Is it worth it to implement a proper gallery support instead? Some sites have browsable links of the photos while other only join links. Thats the only issue/problen i can think of ( getting the actual urls of the photosets ) For a couple of scenes/photosets i browsed it seems most meta is indeed the same with some photosets having a description that the scenes lack.

Ahh... yeah, I think the EvilAngel photo pages are good enough on consistency and populated data to work with just an xPath scraper for `galleryByUrl`... I'll look at adding that

bnkai:
>or were you suggesting add a photo/gallery query function to the Algolia.py script?

yes that was what i meant, sorry
can you try the Algolia.py version i pasted in #1200?

nrg101:
> can you try the Algolia.py version i pasted in #1200?

there is a publicly accessible https://www.transgressivexxx.com/en/photos with galleries listed so I could confirm the updated Algolia.py with gallery scraper works for that domain, and have modified this file in #1200 instead, for the updated script dependency

PR #1209, opened 2022-12-16
add brazilian-transsexuals.com and braziltgirls.xxx

nrg101:
This adds a new scrapers for the Grooby Brazilian sites:

- brazilian-transsexuals.com
- braziltgirls.xxx

These sites share a common page structure that is different to the existing other Grooby site scrapers


nrg101:
CommunityScrapers-validate-add-brazilian

PR #1210, opened 2022-12-16
map all Kink sub-studios

nrg101:
The existing scraper for kink.com parses the studio as `ts-seduction` for TS Seduction scenes. This is the correct studio, but needs to be mapped to the studio name as seen at stashdb: [TS Seduction](https://stashdb.org/studios/ad0ebf7c-a23e-4c62-a149-b1b99fa14b7c)

### DIFF
--- a/scrapers/Kink.yml
+++ b/scrapers/Kink.yml
@@ -55,6 +55,7 @@ xPathScrapers:
                   with: ""
             - map:
                 kink-classics: Kink Classics
+                ts-pussy-hunters: TS Pussy Hunters
                 ts-seduction: TS Seduction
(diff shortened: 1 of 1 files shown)


DogmaDragon:
This applies to most 2 or more words studios <https://www.kink.com/channels>, it's just easier to add them as aliases on local Stash than mapping everything in the scraper.

nrg101:
ha... ok, missed your comment and too late 😆 

I've mapped all the sub-studios listed at stashdb now... it might be worth merging this just to save people the bother of having to add aliases to this number of studios locally

nrg101:
CommunityScrapers-validate-improve-kink

PR #1214, opened 2022-12-19
add scene scraper for ladyboygold.com and tsraw.com

nrg101:
Add scene scraper that works on tour pages, e.g. https://ladyboygold.com/tour/show_video.php?galid=6635&section=1552&nats=MC4wLjYuNi4wLjAuMC4wLjA

### DIFF
--- a/scrapers/LadyboyGold.yml
+++ b/scrapers/LadyboyGold.yml
@@ -4,6 +4,11 @@ performerByURL:
     url:
       - ladyboygold.com/index.php
     scraper: performerScraper
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - ladyboygold.com/tour
+    scraper: sceneScraper
 xPathScrapers:
   performerScraper:
     common:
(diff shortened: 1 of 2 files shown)


scrapers/LadyboyGold.yml
```diff
@@ -4,6 +4,12 @@ performerByURL:
     url:
       - ladyboygold.com/index.php
     scraper: performerScraper
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - ladyboygold.com/tour
```
bnkai, line 10:
``` yaml
     - ladyboygold.com/tour/show_video.php`
```
is probably better to indicate the kind of urls we expect?

scrapers/LadyboyGold.yml
```diff
@@ -4,6 +4,12 @@ performerByURL:
...
+      - tsraw.com
```
bnkai, line 11: Wasnt able to test this.All links seem to be paywalled?

scrapers/LadyboyGold.yml
nrg101: i wasn't sure if `/tour/show_video.php` was the only URI path or not, so just left it as `/tour` to be more open-ended... I think it's fine as-is for the purpose of `sceneByURL` though?

DogmaDragon:
Tested with several URLs from <https://www.tsraw.com/index.php?section=1610> and <https://ladyboygold.com/tour/index.php?section=1555>, everything seems okay.

PR #1224, opened 2022-12-23
Use 1920x1080 for image resize in GammaEntertainment

nrg101:
This just alters the scene image URL to use 1920x1080 instead of the current, low-resolution 960x543.

All the scene cover images I could find were a 1920x1080 source, and I chose to maintain that resolution rather than resizing to the stashdb max of 1280x720, in case stashdb cover image max increases in future.

When submitting an edit/draft, a 1920x1080 is resized to 1280x720 anyway by stashdb.


scrapers/GammaEntertainment.yml
```diff
@@ -139,7 +139,7 @@ xPathScrapers:
               # if using the transport subdomain, parameters need to be passed
               # otherwise a cropped square image is returned by default
               - regex: (https:\/\/transform.+)
-                with: $1?width=960&height=543&enlarge=true
+                with: $1?width=1920&height=1080&enlarge=true
```
bnkai, line 142: I wasnt sure about the enlarge part, if it needs to be removed but all the sample images i found with transform urls seem to now be 1920x1080.
Merging as is and we can have another look later if needed

PR #1225, opened 2022-12-23
add devilsfilm algolia movie scraper

jessqic:
example url: https://www.devilsfilm.com/en/dvd/OnlyTrans-02/93581


nrg101:
I've already done this in #1200

PR #1226, opened 2022-12-24
Update MindGeek.yml (add Noirmale.com)

baloodusudouest:


DogmaDragon:
Please update the data on the last line `# Last Updated October 14, 2022`.

scrapers/MindGeek.yml
```diff
@@ -36,6 +37,7 @@ sceneByURL:
       - transangels.com/scene/
       - transsensual.com/scene
       - trueamateurs.com/scene/
+      - noirmale.com/scene/
```
hphpanon, line 40:
```suggestion
```

Per the automated PR check validation error:

```
URLs for type `scene` should be unique, already exists on /sceneByURL/0/url/15
```

I'm not clear what the value of having two separate lists of URLs for the same scraper type (or if there even is one), but you don't want the same URL listed twice.

nrg101: the two separate lists are for `sceneScraper` and `scriptScraper`, where presumably the domains work with one or the other

bnkai:
Added to scriptscraper section and merging for now.
The preferred scraper should be the new MindGeekApi one, will update the relevant PR(#1240) once the scraper is ok.

Issue #1218, opened 2022-12-21
ManyVids - getting 180p Previews instead


BlokeCDP:
Using:
https://www.manyvids.com/Video/2517066/Jessa-Rhodes---The-Hookup/
this returns a 180p preview image.

Looking at the page source the image in poster= of the MV-VideoPlayer class, this returns a larger preview thumb:
https://ods.manyvids.com/1003987502/o1esp5ndlb1kdp1o24grr197f1v3ub/screenshots/423e531d5050ba3efe9bd2fa6cddcd52.jpg

Is there a consensus that this image should be used over the 180p thumb?

nrg101:
I just noticed this today and manually scraped from that same place in the page to get the 360p image (which was identical to the 180p except resolution)

BlokeCDP:
Thanks nrg101. At least I'm not going mad 😄

bnkai:
Existing scraper uses the player api for the scene image and not the actual page so i didnt notice the image was wrong (too small)
You  can try changing line 120 in `ManyVids.py`
from
```python
    scrape['image'] = meta.get('thumbnail')
```
to
```python
    scrape['image'] = meta.get('screenshot')
```
I dont have videos from that site so i wasnt sure if the `screenshot` was available for all scenes.
Let me know if it works ok and i'll make a PR to prefer the screenshot over the thumbnail

BlokeCDP:
Can confirm that works. Thanks bnkai 👍

bnkai:
Related PR was merged, closing as completed

PR #1229, opened 2022-12-30
Use higher res scene cover for ManyVids

bnkai:
Fixes issue reported in #1218


DogmaDragon:
Verified everything works.

PR #1236, opened 2023-01-06
Fix XConfessions scene scraper (performers, tags and details)

ThatDeltaGuy:


DogmaDragon:
Verified everything works.

scrapers/XConfessions.yml
```diff
@@ -24,7 +24,7 @@ xPathScrapers:
                 with: $1
           - parseDate: 2006-01-02
       Details: &details
-        selector: //div[@class="my-2 tablet:my-0 w-full tablet:w-3/5"]/*[local-name()="h2" or local-name()="span"]
+        selector: //div[@class="flex flex-col items-start w-full my-2 tablet:my-0 laptop:w-8/12 gap-y-2"]/*[local-name()="h2" or local-name()="span"]
```
DogmaDragon, line 27: I would use something like this to include the second part of the description, but maybe that's not something people want. 
```yaml
$details: //div[@class="flex flex-col items-start w-full my-2 tablet:my-0 laptop:w-8/12 gap-y-2"]

selector: $details/h2/text()|$details/div/*
```
They sometimes use `<div>`, and others times `<p>` for it.

nrg101: The Details div class matcher looks very specific and looks like it is client/device/responsive dependent, both the old and the new version.

Seeing as this change is to fix the class attribute no longer matching, I think it would be better to use a looser matcher with `contains`, e.g.

```
//div[@class[contains(.,'items-start')]]
```

or whatever the uniquely identifying class(es) is/are for the desired div

bnkai: From what i can see something like 
```yaml
      Details: &details
        selector: //div[contains(@class, "flex flex-col items-start") and contains(@class, "tablet:") and contains(@class, "laptop:")]/*[local-name()="h2" or local-name()="span" or local-name()="div"]
        concat: "\n\n"
```
should work (not sure if we can use an even looser match) and yes it is probably better to use a `contains` based selector
For the second part of the description as @DogmaDragon mentioned i think they switched to divs so spans shouldnt be needed anymore, leaving it there is probably safer though

JackDawson94 (commented):
Tested - this works !

Maista6969:
Closing this as superseded by #1758

PR #1244, opened 2023-01-19
Add scraper for R18.dev

DoctorD1501:
This adds a scraper for https://r18.dev. R18.dev is a mirror of all the data that R18.com had, but with R18.com going down this site will be needed for future updates. I think they pull in new data too from DMM too, so it should be useful going forward. The R18 family of sites are especially good for translated titles/descriptions compared to the other JAV websites.

I've left some comments on some of the future enhancements that it can maybe support if the R18.dev website refactors their API a bit as well as some potential areas of cleanup during that same refactor.

### DIFF
--- a/scrapers/R18.dev.yml
+++ b/scrapers/R18.dev.yml
@@ -1,0 +1,426 @@
+name: R18.dev (JAV)
+
+sceneByFragment:
+  action: scrapeJson
+  queryURL: https://r18.dev/videos/vod/movies/detail/-/dvd_id={filename}/json
+  queryURLReplace:
+    filename:
+    # gets just the JAV ID out of the filename. This also removes the file extension which is pretty nice.
+    # You can have your filename be something like "Something Something ABC123 Something Something.mp4" and it will scrape as ABC123.
+      - regex: (.*[^a-zA-Z0-9])*([a-zA-Z-]+\d+)(.+)
+        with: $2
+  scraper: sceneSearchIndirect
+  
+sceneByURL:
+  - action: scrapeJson
+    url:
+      - r18.dev/videos/vod/movies/detail/-/i
+    scraper: sceneScraper
+    queryURL: "https://r18.dev/videos/vod/movies/detail/-/combined={url}/json"
+    queryURLReplace:
+      url:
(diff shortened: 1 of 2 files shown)


scrapers/R18.dev.yml
```diff
@@ -0,0 +1,428 @@
...
+    # Title not on this page, so we need to fetch another page to get it
+      Title:
+        selector: content_id
+        postProcess:
+          - replace:
+            - regex: (.+)
+              with: https://r18.dev/videos/vod/movies/detail/-/combined=$1/json
+          - subScraper:
+              selector: dvd_id       
+      # Code not on this page, so we need to fetch another page to get it. It's the same as title. Surely there must be a way to only query once?
```
bnkai, line 384: At the moment there is no way to keep meta from a query. Another query is required, like you do.

scrapers/R18.dev.yml
```diff
@@ -0,0 +1,428 @@
...
+        Aliases:
+          selector: actresses.#.name_kanji
+      Image:
+        selector: jacket_full_url
+      Director:
+        selector: directors.#.name_romaji
+      Details:
+        selector: title_en
+        # All this uncensoring stuff happens automatically on the HTML version of the page due to some JS on the frontend there.
+        # I put in a request to the site dev to add this to the backend json api as well so we don't have to do it everytime ourselves plus keep a ever growing list of this stuff
```
bnkai, line 65: Sounds good. The issue was present in javlibrary and the original r18 site so we had to do that as well in the javlibrary_python scraper (you can have a look there for more word replacements if you want). I will merge the following as they are for now

LogicSense1:
There is an issue when setting cover image. The reason may be caused by certs

In WebUI,

sceneUpdate: input: sceneUpdate http error 404

In server,

Could not set image using URL https://pics.dmm.co.jp/mono/movie/adult/ssis663/ssis663pl.jpg: Get "https://pics.dmm.co.jp/mono/movie/adult/ssis663/ssis663pl.jpg": x509: certificate has expired or is not yet valid: current time 2023-05-03T00:51:37Z is after 2015-10-09T03:52:02Z

PR #1245, opened 2023-01-20
Add performer scraper for Babestation TV

vonschmooze:
Fairly basic performer scraper for Babestation.tv

### DIFF
--- a/scrapers/BabestationTV.yml
+++ b/scrapers/BabestationTV.yml
@@ -1,0 +1,44 @@
+name: "BabestationTV"
+
+performerByName:
+  action: scrapeXPath
+  queryURL: https://www.babestation.tv/girls/page/1?search={}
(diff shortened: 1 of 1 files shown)


scrapers/BabestationTV.yml
```diff
@@ -0,0 +1,44 @@
...
+                with: https://www.babestation.tv
+
+  performerScraper:
+    performer:
+      Name: //*[@id="app"]//h1/text()
+      Image: //*[@id="app"]//div[@class="w-full relative"]/img/@src
+      Gender:
+        fixed: "Female"
+      Country:
+        fixed: "United Kingdom"
```
bnkai, line 40: Those dont seem to be fixed.
Having a look at the tags we can see that not all performers are actually from UK and that there are some non female as well (tag trans) ?

PR #1248, opened 2023-01-22
Add scraper for Anime-DB

timo95:
Add scraper for [Anime-DB](https://adultanime.dbsearch.net/). It is a japanese database of hentai anime and eroge. Each entry is a DVD/BD/etc. release, which in most cases means one or more entries per episode. They have only very basic tags, but most entries have the official cover image, initial release date and a studio code.

Search only works with japanese and the results are also in japanese.


scrapers/Anime-DB.yml
```diff
@@ -0,0 +1,125 @@
...
+        postProcess:
+          - replace:
+              - regex: "^"
+                with: "https:"
+      Image:
+        selector: //section[@class="item-box"]/div[@class="item-img"]/a/img/@data-src
+        postProcess:
+          - replace:
+              - regex: "^"
+                with: "https:"
```
bnkai, line 45: do the images in the search work for you?
even when trying to access them directly i get a 403 error

timo95: No. Without CDP it worked sometimes but loading very late. Probably something to do with the lazy loading.

scrapers/Anime-DB.yml
```diff
@@ -0,0 +1,125 @@
...
+                with: "-"
+              - regex: "\u65E5"
+                with:
+              - regex: -(\d)-
+                with: -0$1-
+              - regex: -(\d)$
+                with: -0$1
+
+driver:
+  useCDP: true  # needed for the age confirmation redirect - cookies only work temporarily
```
bnkai, line 123: Are you sure about this?
Did you try with both cookies?
```yaml
driver:
    cookies:
      - CookieURL: "https://adultanime.dbsearch.net/"
        Cookies:
          - Name: "adultanime_dbsearch_net[bc]"
            Domain: "adultanime.dbsearch.net"
            Value: "true"
            Path: "/"
          - Name: "adultanime_dbsearch_net[id]"
            Domain: "adultanime.dbsearch.net"
            Value: 0123456789
            Path: "/"
```
For the second value use the actual Cookie value from your browser not the one (0123456789) above

Also how where you  running CDP ? Remotely via address or localy by path? The first works i am not sure if the second would.

timo95: Yes i did. It worked until the second one expired (1 week). Then i had to update the value.

Via a separate docker image aka remotely.

scrapers/Anime-DB.yml
timo95: Maybe adding sleep to CDP would allow it to finish loading the images.

timo95:
I replaced Maker with Label for the Studio field.

scrapers/Anime-DB.yml
bnkai: This might not work for users that use CDP locally since the cookies may be reset between scrapes. Will merge as is and make a new PR if needed

PR #1251, opened 2023-01-24
Update Hypnotube.yml to pull in main video thumbnail

zzazzcdts:
First time submitting a pull request, so apologies if I've made a hash of it.

I've made a change to the Hypnotube.yml scraper so that it pulls in the main thumbnail for the video (as seen on the front page/video list view). This thumbnail seems to be the best representation of videos from the site.

Tested on my own instance of Stash and validated via the validation script, which comes back clean.


bnkai:
The selector you use now was what was proposed in the original (https://github.com/stashapp/CommunityScrapers/pull/1050) PR. It didnt work then for all scenes, it seems that it works now though (they probably changed something in the site again).

PR #1254, opened 2023-01-28
The Score Group: Add scoreland2 and modify title XPath

hphpanon:
As far as the XPath goes, the structure has changed (most critically h2 -> h1), but also the section's ID value is not consistent across sites. On Pornmegaload it's the videos_page-page that is there already. On scoregroup2 it's mixed_page-page. If there's a way to do an endsWith check, that would likely get us some desirable specificity, but at the very least I can confirm the XPath in this PR works on both sites mentioned earlier.


scrapers/TheScoreGroup.yml
```diff
@@ -18,7 +19,7 @@ galleryByURL:
 xPathScrapers:
   sceneScraper:
     scene:
-      Title: //section[@id="videos_page-page"]/div[@class="row"]/div/h2/text()
+      Title: //main/div/section[1]/div[@class="row"]/div/h1/text()
```
bnkai, line 22: there is no ends with isnt the contains enough though?
```suggestion
      Title: //section[contains(@id,"_page-page")]/div[@class="row"]/div/h1/text()
```

hphpanon:
Had a minor mishap with pushing the date update. Sorry about the extra notifications around the changes but it should now include your improvements to the XPath in addition to a further-updated modified date in the yaml.

PR #1255, opened 2023-01-29
HelixStudios.yml - Scene Scraper - increase image size, add director, add studio code

Valmox:
Scene Image: use regex to switch the pulled image to a larger version.

Scene Director: pulled from the director class.

Scene Studio Code: extracted from the title image URL (Helix name all media files with the studio code), using regex to remove everything else in the URL.

### DIFF
--- a/scrapers/HelixStudios.yml
+++ b/scrapers/HelixStudios.yml
@@ -56,15 +56,7 @@ xPathScrapers:
       Date:
         selector: //div[@class="info-items"]/span[@class="info-item date"]/text()
         postProcess:
-          - replace:
-              - regex: "th|st|nd|rd"
-                with:
-              - regex: '(\d+)\sdays\sago'
-                with: $1
-              - regex: '^([a-zA-Z]+)\s(\d+)$'
-                with: "$1 $2, 2021"
-          - subtractDays: true
-          - parseDate: Jan 2, 2006
+          - parseDate: January 2, 2006
       Details:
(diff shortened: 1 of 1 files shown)


bnkai:
Regarding the studio code.
Do you mean that it is in the form of `hx123_scene123` because thats how the original files are named?
I have no media files from that site so the `id` in the URL looked like a more obvious studio code?

They seem to have changed the scene date format as well, can you verify that
```yaml
      Date:
        selector: //div[@class="info-items"]/span[@class="info-item date"]/text()
        postProcess:
          - parseDate: January 2, 2006
```
works for you?

ps update the `# Last Updated` date as well

Valmox:
Cheers bnkai,

**Studio Code:** They've changed the URL codes during previous site updates. However, the hx123_scene123 type always remains static. They use it on both the media files for the tour and membership sections, for example:
Banner at the top of the scene tour page: https://cdn.helixstudios.com/media/stills_ws/hx135_scene20_1600.jpg
Gallery image on the tour page: https://cdn.helixstudios.com/img/250w/media/stills/hx135_scene20_001.jpg
Video in the tour player: https://media.helixstudios.com/scenes/hx135_scene20/hx135_scene20_preview_hls/hx135_scene20_preview.m3u8

That code is the only consistent unique ID that the studio never changes.

**Date:** Sorry, I didn't notice that. Thanks.

Update Date: Will do, apologies.

Cheers

PR #1257, opened 2023-01-29
Add scraper for CockyBoys

Jacintocuenca:


scrapers/CockyBoys.yml
```diff
@@ -0,0 +1,43 @@
...
+      $performer: //div[@class="movieModels"]/span/a[@class="name gothamy"]
+    scene:
+      Title: //meta[@property="og:title"]/@content
+      Date:
+        selector: //strong[contains(text(), "Released:")]/following-sibling::text()
+        postProcess:
+          - parseDate: 01/02/2006
+      Details:
+        selector: //div[@class="movieDesc"]/h2/following-sibling::text()
+        concat: "\n\n"
```
bnkai, line 25: This will need some more tweaking
Any scene that has performer links in the description (eg `https://cockyboys.com/scenes/olivier-robert-raphael-louis.html?type=vids` ) has an issue with this
From a quick look this
```yaml
      Details:
        selector: //div[@class="movieDesc"]
        postProcess:
          - replace:
              - regex: ^Description\s+
                with:
```
gets the whole description just not that well formatted :-(
Not sure if there is an easy way to get all the description with the formatting as close as possible to the original

Maista6969:
I apologize for stepping on your toes with this one, but I have integrated your changes into [CockyBoys.yml](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/CockyBoys.yml): the details are still not correctly scraped due to the issue bnkai pointed out, but I also do not think we can capture the proper formatting using an XPath scraper

PR #1258, opened 2023-01-31
Update GoddessSnow.com scraper to have search by name functionality. …

DoctorD1501:
…Use /scenes URL instead of /updates URL as it has more accurate and complete information.

Specifically, the /scenes URL has the correct release date (things are released a year later on /updates) and always the full description instead of a truncated one like on /updates.

### DIFF
--- a/scrapers/GoddessSnow.yml
+++ b/scrapers/GoddessSnow.yml
@@ -8,10 +8,25 @@ sceneByName:
 
 # We do not have the /updates URL here because it has the wrong release date by a year (scenes get released a year early on /scenes)
 # And also the descriotion is truncated often on /updates
+# /scenes also has two versions, one that ends in "_vids.html" and one that ends in ".html" 
+# We want to make sure we get the _vids.html version as that is the one with the images
 sceneByURL:
   - action: scrapeXPath
     url:
+      - goddesssnow.com/updates/
       - goddesssnow.com/vod/scenes
+    queryURL: "{url}"
+    queryURLReplace:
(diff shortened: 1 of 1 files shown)


bnkai:
Are all scenes from /updates in the vod section and are they the same all the time (1<->1 )?
If so does it make sense instead of removing the `/updates` to just url replace so that we scrape the vod url instead?
Eg 
```yaml
# We do not have the /updates URL here because it has the wrong release date by a year (scenes get released a year early on /scenes)
# And also the descriotion is truncated often on /updates
sceneByURL:
  - action: scrapeXPath
    url:
      - goddesssnow.com/updates/
      - goddesssnow.com/vod/scenes
    queryURL: "{url}"
    queryURLReplace:
      url:
        - regex: /updates/
          with: "/vod/scenes/"
    scraper: sceneScraper

```

DoctorD1501:
I think /scenes has more than /updates does since /updates gets stuff a year later, so something like this may work. The proposed code doesn't work exactly (we want the _vids.html version of the page), but I can probably rework it in this manner to do what you are suggestion. I'll add a new commit later today to do so.

DoctorD1501:
@bnkai - OK, I've updated things now. This will handle /updates, /scenes that do not end in _vid.html and /scenes that do end in _vids.html.

We need it to be /scenes that end in _vids.html in order to get the pictures. There is confusingly a version of scenes without that "_vids" that do not have the pictures! Quite the minefield of bad pages on this site!

PR #1263, opened 2023-02-03
Fix Clips4Sale images on scene scraper and search

DoctorD1501:
Minor edits to get images working again on this scraper. Left the `clips4sale.com/search/clip` stuff alone since I honestly have no idea what this is for or how to use it, so I couldn't test it to make updates. Can make a further commit to fix that if anyone lets me know how to use it. :)

### DIFF
--- a/scrapers/Clips4Sale.yml
+++ b/scrapers/Clips4Sale.yml
@@ -29,11 +29,12 @@ xPathScrapers:
       URL:
         selector: $card//a[contains(@class, "search-clip__title")]/@href
       Image:
-        selector: $card//img[@draggable="false"]/@data-src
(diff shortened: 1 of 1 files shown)


DoctorD1501:
Did a few force pushes to my branch to fix the failing test on the "Last Updated" part at the end of the file. Apparently you need a zero in front of '3' in February 03. :)

bnkai:
> Left the clips4sale.com/search/clip stuff alone since I honestly have no idea what this is for or how to use it, so I couldn't test it to make updates

When the scraper was merged some of the clips4sale studios were available with a different layout as well (using `/search/clip` slugs) and thus the extra scraper for that part. Not sure if they are still available so leaving as is for now.

PR #1271, opened 2023-02-07
Create scraper Zishy.yml

quantumandwoody:
Create a scraper for the site www.zishy.com


Zishy.yml
```diff
@@ -0,0 +1,53 @@
...
+          - replace:
+            - regex: added on (\w{3}) (\d{2}). (\d{4})
+              with: $3-$1-$2
+          - parseDate: 2006-Jan-02
+      Image:
+        selector: //video[@poster]/@poster
+        postProcess:
+          - replace:
+            - regex: /(.*)
+              with: https://zishy.com/$1
```
bnkai, line 36: Do you have any urls available where images match this?
I tried the galleries with video part but the few i tried had the image of the video in the `Bonus HD Video Clip` part
eg
`https://www.zishy.com/albums/2077-ariel-gossimer-too-much-awesome-2`
To get the images from those something like
```yaml
      Image:
        selector: //div[@class="player"]//@style[contains(.,"background")]
        postProcess:
          - replace:
            - regex: .+url\((/uploads/.+)\);.+
              with: https://zishy.com/$1
```
works for me

quantumandwoody: Thank you for that. Your change works much better and grabs the actual poster frame of the video from the page. I have made the adjustments.

PR #1274, opened 2023-02-15
Add ThisVid xPath scene scraper

imagineimaginingthings:
Gets title, image, tags and description. Date is not scraped as it's only publicly displayed relatively, and very vaguely at that (x years ago, x months ago etc.)


bnkai:
Looks ok
Do we want the id as a studio code? eg 
```yaml
      Code:
        selector: //meta[@property="og:video:url"]/@content
        postProcess:
          - replace:
              - regex: .+/(\d+)/?$
                with: $1
```
?

imagineimaginingthings:
Hmm, we could. Is that a generally accepted use case for the studio code field now then? I did notice [this discussion](https://github.com/stashapp/StashDB-Docs/issues/4) being linked in another recent pull request talking about the same thing, but haven't seen if that was generally agreed upon yet.

I suppose having extra metadata available is never a bad thing, so I'm happy to add that in if we think this'll be fine in the long term.

Perhaps worth mentioning that with ThisVid the video id is arguably not particularly useful as it's not used in the URL slug. It is however used in the embeddable player URL `thisvid.com/embed/[videoId]` but this isn't as good for scraping as it lacks the description box. URL slugs are not updated if the video uploader renames the video, so in theory, they should always work.
That's not to say we shouldn't add it, just pointing it out.

bnkai:
Merging as is. If needed scraper can be updated in a future PR.

PR #1275, opened 2023-02-15
Improve Xvideos scraper

Stephan972:
- Added sceneByFragment support
- Added scene Date field scraping
- Added scene Url field scraping
- Added more domains
- Fixed scene Studio scraping
- Capitalized scraper name

### DIFF
--- a/scrapers/Xvideos.yml
+++ b/scrapers/Xvideos.yml
@@ -4,7 +4,7 @@ sceneByURL:
     url:
       - xvideos.com
       - xvideos2.com
-      - xvideos.es
+      #- xvideos.es
       - xvideos.red
     scraper: sceneScraper
 
--- a/scrapers/Xvideos.yml
+++ b/scrapers/Xvideos.yml
@@ -1,10 +1,29 @@
-name: xvideos
+name: Xvideos
 sceneByURL:
   - action: scrapeXPath
     url:
       - xvideos.com
+      - xvideos2.com
+      - xvideos.es
       - xvideos.red
     scraper: sceneScraper
+
+sceneByFragment:
+  action: scrapeXPath
+  scraper: sceneScraper
+  queryURL: https://www.xvideos.com/video{filename}/x
+  queryURLReplace:
+    filename: 
+      # expects an id in square brackets before extension, as saved by yt-dlp by default
+      - regex: '.*\[([0-9a-zA-Z]{4,})\]\.[^\.]+'
+        with: $1
+      # or expects an id
+      - regex: '(?i)^.+-(\d+)\..+$'
+        with: $1
+      # if no id is found in the filename
+      - regex: .*\.[^\.]+$
+        with: # clear the filename so that it doesn't leak
+
 xPathScrapers:
   sceneScraper:
     scene:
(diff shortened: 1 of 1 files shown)


scrapers/Xvideos.yml
```diff
@@ -1,10 +1,29 @@
-name: xvideos
+name: Xvideos
 sceneByURL:
   - action: scrapeXPath
     url:
       - xvideos.com
+      - xvideos2.com
+      - xvideos.es
```
DogmaDragon, line 7: I would prefer to exclude this from the scraper. It's machine translated metadata, not original releases, which would create confusion if submited to the stash-box instances.

Stephan972: If a user uses one of those domains, he would be left off by Stash. 

- xvideos2.com
This domain redirects to xvideos.com. See: https://wheregoes.com/trace/2023732209/
yt-dlp also supports this domain.

- xvideos.es
_"create confusion if submited to the stash-box instances."_
stash-box may implement some mecanism (tag aliases, rejection of text written in an unknown language etc) allowing only metada in a specific language.

I would exclude xvideos.es only. Do you agree ?

DogmaDragon: I only meant `xvideos.es`, not sure why it selected both. Sorry for the confusion.

stash-box is just the software so not sure it will implement any language restrictions. But canonical community instance StashDB allows non-English content via unique entry as long as it has an official link. And I think this would create confusion as they are not official releases, but just AI translated metadata of the English release. 

From Stash perspective I can see the value in them and managing them is up to individual user. Maybe instead of removing it, you can just comment out so it's not used by default.

Stephan972: Comment done.

scrapers/Xvideos.yml
```diff
@@ -14,11 +33,20 @@ xPathScrapers:
       Performers:
         Name: //li[@class="model"]/a/span[1]/text()
       Studio:
-        Name: //a[@class="btn btn-default label main uploader-tag hover-name"]/span[@class="name"]/text()
+        Name: //li[@class="main-uploader"]/a[starts-with(@href,'/channels')]/span[@class="name"]/text()
```
bnkai, line 36: this breaks scenes where the studio is an amateur eg `https://www.xvideos.com/video75110099/real_amateur_milf_gets_passionate_pounding_in_her_lingerie`
do we get any false matches if we relax the selector a bit  ? eg 
```yaml
        Name: //li[@class="main-uploader"]/a/span[@class="name"]/text()
```

Stephan972: >do we get any false matches if we relax the selector a bit ? (...)

Yes, if original poster is any random person having just a *profile*, he/she gets considered as a real studio.
This is why I added precision with *channels* only in the xpath.

Beside the amateur channel you provided (@bnkai), they are other channels types :
- channels
https://www.xvideos.com/video74652617/english_tutor_camilla_seduces_her_pupil_john_luna_with_her_big_natural_tits
- amateur-channels
https://www.xvideos.com/video75110099/real_amateur_milf_gets_passionate_pounding_in_her_lingerie
- model-channels 
https://www.xvideos.com/video75149889/29183767/0/all-natural_jessica_starling_gets_passionate_sex_from_johnny_sins
- pornstar-channels
https://www.xvideos.com/video74936872/mia_khalifa_-_trying_anal_it_s_never_going_to_happen_with_english_subtitles_
- ??-channels
...

Xvideos FAQ indicates there can be amateurs, porn stars,webcam models and erotic models account statuses.
*Model : model accounts can be amateurs, porn stars, webcam models, and erotic models. They indicate that you are the main content for your own account* (https://info.xvideos.com/account-status)
I wasn't able to find any other types of channels given this information.

So we are left with the following options so far :
1. Accept anything (original option) with the drawback of entering random person as real studio
2. Accept everything except random person (ie *profiles*) with the drawback of leaving the door opened for accepting unexpected "thing" as real studio
3. Be flexible by accepting anything matching `(?:.*-)?channels` with the same drawback as in (2.)
4. Be restrictive by accepting only the above listed channels types leaving the opportunity to add later newly discovered "thing" 
5. Any other option...

Stephan972: Amateur channels are now supported. Addionnaly, model and pornstart channels are also supported.

Boomslang29 (commented):
> -

PR #1276, opened 2023-02-18
Added DoTheWife.com scene scraper

l1xander:
Added dothewife.com scene scrapper for
    - Title
    - Performers
    - Scene cover
    - Tags
    - Details

### DIFF
--- a/scrapers/DoTheWife.yml
+++ b/scrapers/DoTheWife.yml
@@ -23,14 +23,11 @@ xPathScrapers:
         Name:
           fixed: Do The Wife
       Image:
-        selector: //img[@class="video-banner"]/@src|//div[@class="vjs-poster"]/@style
+        selector: //img[@class="video-banner"]/@src|//video/@poster
         postProcess:
           - replace:
               - regex: (?:.+)(\/usermedia\/.+\.jpg)(?:.+)
                 with: $1
               - regex: "^/usermedia/"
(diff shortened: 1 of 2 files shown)


scrapers/DoTheWife.yml
```diff
@@ -0,0 +1,36 @@
...
+        selector: //img[@class="video-banner"]/@src|//div[@class="vjs-poster"]/@style
+        postProcess:
+          - replace:
+              - regex: (?:.+)(\/usermedia\/.+\.jpg)(?:.+)
+                with: $1
+              - regex: "^/usermedia/"
+                with: "https://dothewife.com/usermedia/"
+driver:
+  useCDP: true
+  sleep: 4
```
bnkai, line 35: Is this even needed?
If it is for the image part, we can get it directly from the page source (`//video/poster`) eg the below seems to work without any cdp
```yaml
      Image:
        selector: //img[@class="video-banner"]/@src|//video/@poster
        postProcess:
          - replace:
              - regex: "^/usermedia/"
                with: "https://dothewife.com/usermedia/"
```

l1xander: Thank you for pointing this out. I have updated the PR with the above change.

PS: I had copied this from another scraper and adjusted it for dothewife.com. The CDP stuff were leftover from the original yml file.

l1xander:
@bnkai I have updated the PR.

PR #1281, opened 2023-02-27
Improve ThisVid scraper

imagineimaginingthings:
* Now uses video id for Studio Code (as mentioned in the [original PR](https://github.com/stashapp/CommunityScrapers/pull/1274))
* Adds performerByURL scraper for fetching name, photo, country, birthdate and gender from a member page if available.

I'm well aware that the majority of ThisVid members simply upload videos they've found on other platforms, but for any actual content creators on there I'd figured we should be able to scrape them.


JackDawson94 (approved):
Tested - works as intended

PR #1291, opened 2023-03-16
Revision of AEBN scraper

SirCumAlot1988:
Hi,

I propose this as an alternative/replacement to the existing AEBN scraper.

It has the same functionality as the existing AEBN scraper with one major additional feature:

It is now possible to scrape the metadata of a specific movie scene. This is helpful if you have a movie saved as split scenes (i.e. one file per movie scene). To scrape a single scene instead of the complete movie, do one of the following:

1. Enter the movie URL and add a separator + scene nr to the URL. Default separators are plus, comma and full stop but you can define your own in the header of the .py file. Example: If you want to scrape scene 2 of Kendras Obsession, enter https://straight.aebn.com/straight/movies/218523/kendras-obsession+2

2. Enter the name of the movie in the search field at the top of the AEBN site. You will now get a list of scenes. If you hover over the scenes, you will see the movie title and scene nr. Search for the scene you want and click on the 3 points ("Scene Information"). A popup window will appear. Now copy the link at the top of the popup window (in this example the link "Kendras Obsession, scene 2") and hand it over to the scraper.

I personally prefer option 1, but it is up to you. 

This way the scraper will load **only** the metadata (actos, tags, etc.) of this specific scene. The scraper randomly selects one of the scene thumbnails as cover image. If the complete movie is scraped, it uses the movie cover.

Despite that I added some minor improvements:

- Movie performers are now scraped correctly. If you have a movie with a long list of performers (e.g. Orgy Masters 6),  some performers will be loaded via ajax calls. Hence, the old scraper did not scrape all performers. This is now handled properly.
- The performer scraper now scrapes aliases, tattoos and piercings, which are "hidden" in the performer biography (see Anna Bell Peaks as an example for tattoos/piercings and Angelika Grays for aliases)

Finally, as an "experimental" feature, I added the option to invoke the performer scraper from the scene scraper. If you set the option "scrape_performer_details" in the header of the .py file to true, the scene scraper will scrape the details for each performer of the scene. So, if you do not have this performer yet and create it from the scene scraping window the performer detail fields will be populated already. Similarly with the option "scrape_performer_images" enabled, you will have the performer image available without the need to rescrape that performer. Of course you can combine both options. However, admittedly this slows down the scraper. So, I am not sure if it is a reasonable option and I am open for discussions on that.

Cheers!

### DIFF
--- a/scrapers/AEBN.yml
+++ b/scrapers/AEBN.yml
@@ -1,11 +1,11 @@
-name: AEBN_dev
+name: AEBN
 sceneByURL:
   - url:
       - aebn.com
     action: script
     script:
       - python3
-      - AEBN_dev_v02.py
+      - AEBN.py
       - scene
 movieByURL:
   - url:
@@ -13,7 +13,7 @@ movieByURL:
     action: script
     script:
       - python3
-      - AEBN_dev_v02.py
+      - AEBN.py
       - movie
 performerByURL:
   - url:
@@ -21,6 +21,6 @@ performerByURL:
     action: script
     script:
       - python3
-      - AEBN_dev_v02.py
+      - AEBN.py
       - performer
 # Last Updated March 16, 2023


scruffynerf:
thank you for this code.  Been doing a bunch of AEBN scraping and didn't realize it wasn't get all the performers.
I actually managed to massage the yml version  to grab the full list (among other tweaks I've made to it), and the essence of this is good, but it's using too many one-offs, and would benefit from a big rewrite using StashAPI.
On my radar to do do that rewrite (because as much as I love yml, I'm now pushing it to a limit... subscraping every performer works, but...

SirCumAlot1988:
Yeah, I also did not realize in the beginning that not all performers are scraped. Actually I came across it when I implemented this scraper.

However, in my eyes the main advantage of my scraper is that it can scrape a single scene of a movie. That's really helpful if you have movies as split scenes. Scraping the metadata of the whole movie doesn't make too much sense in this situation.

AEBN is really good at providing metadata for each scene separately. I realized too late however that hotmovies is even better in this regard. So, maybe I will implement something similar for hotmovies in the future.

feederbox826:
Committing to sitting down and going through this today, hopefully going to get it updated

Update: not entirely abandoned, looking at code parity that might still be possible with yml and if it would be possible to convert scene search into a yml or only split off scene search into py, and also doing performer lookup with sub-scrapes

feederbox826:
probably end up having to just close as stale since a lot has changed and we have real groups support

PR #1297, opened 2023-03-22
Add BoyFun scraper

MortonBridges:
Add new scraper for boyfun.com and updated scrapers list accordingly.

### DIFF
--- a/scrapers/BoyFun.yml
+++ b/scrapers/BoyFun.yml
@@ -6,25 +6,25 @@ sceneByURL:
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
+    common:
+      $perf: //span[@class="models"]/span[@class="content"]/a
     scene:
-      Title: 
-        selector: //span[@class="title"]
+      Title: //span[@class="title"]
       Image: //video/@poster
       Date:
         selector: //span[@class="date"]/span[@class="content"]
         postProcess:
           - replace:
-              - regex: (st|[nr]d|th) 
-                with: "" 
+              - regex: (st|[nr]d|th)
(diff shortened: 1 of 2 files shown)


scrapers/BoyFun.yml
```diff
@@ -0,0 +1,31 @@
...
+      Date:
+        selector: //span[@class="date"]/span[@class="content"]
+        postProcess:
+          - replace:
+              - regex: (st|[nr]d|th) 
+                with: "" 
+          - parseDate: Jan 2, 2006
+      #Details: //div[@class="heading"]/following-sibling::p[1]
+      Details: 
+        selector: //div[@class="heading"]/following-sibling::text()
```
DogmaDragon, line 22: `//div[@class='content-information-description']/p/text()` also works. Any reason to not go with the more precise selector?

DogmaDragon:
Tested, everything works.

scrapers/BoyFun.yml
MortonBridges: > `//div[@class='content-information-description']/p/text()` also works. Any reason to not go with the more precise selector?

No specific reason, except that I probably missed the obvious.

scrapers/BoyFun.yml
```diff
@@ -0,0 +1,31 @@
+name: BoyFun
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - www.boyfun.com/video
```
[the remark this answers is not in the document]
MortonBridges: Thanks.  I've removed the www from the url.  Also removed an errant, unnecessary commented line.

PR #1300, opened 2023-03-24
Update Spizoo.yml

randoumuser:
Added:
* Mr Lucky Raw
* Mr Lucky Vip


JackDawson94 (changes requested):
Tested the two new sites

It seems they don't follow the same template as the others in this scraper.
Performers are not parsed correctly on both
Image is not parsed on mrluckyraw

Maista6969:
Closing due to inactivity

PR #1301, opened 2023-03-26
JavLibrary_python.yml, Better formatting keywords before sending request to source site

AndyHoang:
javlibrary python is sending  whole filename as  scene title, therefore Identifying task usually have zero result just because media file naming is bad

use https://github.com/jvlflame/Javinizer/blob/master/src/Javinizer/jvSettings.json#L45 from javinizer seem to be better for guessing jav code

Some of testcase:

```
[Mosaic Removed Uncensored]MEYD-799 愛妻交換 後輩の妻と俺の妻を交換して中出ししまくった4日間の記録。 森沢かな 南ま
```
become `MEYD-799`

`hhd800.com@JUQ-208.mp4`

become `JUQ-208`


feederbox826:
Not familiar with JAVLibrary and the code has strayed a lot from the original base. Will close as it's been stale for years

PR #1306, opened 2023-03-29
Convert DoTheWife scraper to SmutPuppet

MortonBridges:
The existing DoTheWife scraper works equally well for other sites in the SmutPuppet network, including SmutPuppet itself.  

URLs have been added to include other sites in the network. Modifications have been made to generalize the image URLs and to gather the appropriate studio name for each site.

This has been validated with a random scene from each site, but the sites are a bit odd, particularly in how they display images.  There are two known methods for referencing a scene's poster image, which means there might be unknown methods that haven't been encountered yet.

A Performer by URL scraper was added.  There's not much info available (name, nationality, date of birth and image).  A mapping is included for some "non-country" text the sites use.  They do not have a consistent way of presenting the nationality.  Most times it's the country name, other times it's the nationality (Russia vs Russian, for example).  Sometimes it's in ALL CAPS, other times it's Camel Case.   It's extremely likely that additional mappings will be necessary.

Scrapers list has been updated for all the URLs covered by this scraper.


Maista6969 (changes requested):
Would you mind giving this another look since #1406 was merged?

MortonBridges:
No longer necessary after implementation of 1406.  Closing as unneeded.

PR #1307, opened 2023-03-29
Fix RealJamVR scraper

Emilo2:
- Fixed to work with the new layout
- Added studio code from the trailer URL
- Added gallery scraping


JackDawson94 (changes requested):
Testing, made an error in my previous comment, resuming testing

Maista6969:
This could use another look after the merge of #1387 so we can get galleries from this scraper 👍

Maista6969 (changes requested):
Will need to pull in the changes from master and resolve the conflicts arising from #1387

Emilo2:
Merged from master.

#1387 failed to fetch multiple performers and this should fix it.
Also the additional tags should be properly fetched now.

Maista6969 (approved):

PR #1310, opened 2023-04-01
Add white label Let's Doe It domain

devnsfw:
This adds the domain for a white-label version of Let's Doe It. I know there's at least another white label site, but I can't remember the URL, if I remember it I'll submit a new PR

### DIFF
--- a/scrapers/LetsDoeIt.yml
+++ b/scrapers/LetsDoeIt.yml
@@ -3,14 +3,15 @@ sceneByURL:
   - action: scrapeXPath
     url:
       - amateureuro.com
-      - dirtycosplay.com      
+      - dirtycosplay.com
+      - doe-tv.com
       - doegirls.com
       - forbondage.com
       - letsdoeit.com
       - mamacitaz.com
+      - mrskin-letsdoeit.com
       - transbella.com
-      - vipsexvault.com
-      - doe-tv.com
+      - vipsexvault.com      
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
--- a/scrapers/LetsDoeIt.yml
+++ b/scrapers/LetsDoeIt.yml
@@ -10,6 +10,7 @@ sceneByURL:
       - mamacitaz.com
       - transbella.com
       - vipsexvault.com
+      - doe-tv.com
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:


DogmaDragon:
Added site scrapes full metadata while official site only partial. I think the official site selectors should be fixed instead.

Also the site offers different date. Might be a re-release, don't know.

Tested on 
<https://letsdoeit.com/watch/73111/playing-with-myself.en.html> - date returned `2021-04-23`
<https://doe-tv.com/watch/73111/playing-with-myself.en.html> - date returned `2023-03-30`

devnsfw:
Since this site is several months (and one channel) behind the main one, I think it's useful to refer to the date used by the secondary site, if that's where the user is downloading their videos from. If not, they would be using the URL from the main site as well.
In this way, the scenes will appear on the home page "recently released" section, which would be correct for those users who doenload from the secondary site, but which would not happen if referring to the main URL

DogmaDragon:
I'm looking from the [StashDB (canonical stash-box instance) perspetive](https://guidelines.stashdb.org/docs/scenes/scene-dates/#preferred-scene-dates). They prefer original date and anything submitetd from this site would be incorrect. 

It's not up to me if this gets added, I just said my opinion.

bnkai:
@DogmaDragon regarding the specific scraper i will have a look at fixing the selectors for the `main` site. Like a couple of other scrapers this is mostly a `convenient` scraper, for those who download from that site and want to automate things. The date is obviously wrong but i am not sure how we should handle that... I think it is probably easier/better to `blacklist` any entries to stashdb with this domain (considering that all meta is available from source/official directly)?

DogmaDragon:
@devnsfw argument is that secondary site date is preffered by him, if that's the source he used to get the scene, even if it's years off it's official release. 

I agree about blacklisting it on StashDB side since it offers no value for StashDB. The whole purpose of the site is to recirculate old scenes as new.

devnsfw:
Yeah, since the scraper doesn't use member access, I could go and fetch the data from the official site. I just need to replace the domain in the URL, the path is the same. But if I do it while I'm browsing, it's just more convenient to use the while label site.
It's not a mayor annoyance to go to the official site, though, so feel free to drop the PR.
I don't know if scrapers can do this (just getting started with them), but I am thinking that maybe the URL could be recognized for the purposes of enabling the scrape button, but then the URL could be transformed by replacing the domain for the main one. This way, it would fetch the official date for each scene (the same had occurred to me for other sites, but changing the members URL for the public URL *iff* they have the same path or one can be automatically transformed in the other)

bnkai:
@DogmaDragon @devnsfw sorry for the late reply
According to what @devnsfw mentioned it seems we can do a URL replace for `doe-tv.com` and then use the official site data
I tried the below (fixes performer/tag/studio for letsdoeit  and does a simple URL Replace)
It seems to work ( minimal testing with scenes from doe-tv and letsdoeit )
```yaml
name: "LetsDoeIt"
sceneByURL:
  - action: scrapeXPath
    url:
      - amateureuro.com
      - dirtycosplay.com
      - doegirls.com
      - doe-tv.com
      - forbondage.com
      - letsdoeit.com
      - mamacitaz.com
      - transbella.com
      - vipsexvault.com
    queryURL: "{url}"
    queryURLReplace:
      url:
        - regex: doe-tv\.com
          with: letsdoeit.com

    scraper: sceneScraper
xPathScrapers:
  sceneScraper:
    common:
      $actors: //div[@class="actors" or @class="-mvd-grid-actors"]
      $details: //div[@class="row sides-xs"]
      $letsdoeit: //div[@class="-mvd-grid-more"]//span/a[@class="-mvd-list-url"]
    scene:
      Title:
        selector: //title/text()
        postProcess:
          - replace:
              - regex: (.+)(?:\s\|.+)
                with: $1
              - regex: (\sExclusive\W+(\w+\W+)?Porn\W+Video)
                with: ""
      Date:
        selector: //meta[@itemprop="uploadDate"]/@content
        postProcess:
          - replace:
              - regex: (\d{4})-(\d{2})-(\d{2}).+
                with: $1-$2-$3
          - parseDate: 2006-01-02
      Details:
        selector: //meta[@itemprop="description"]/@content
        postProcess:
          - replace:
              - regex: "&#039;"
                with: "'"
              - regex: (.*\.?\!?)(?:\s-\s\w.*-.*)$ # remove Studio name at the end of a description
                with: $1
      Tags:
        Name: $details//div[@class="col"][4]//a/text()|$details//div[@class="col"][6]//a/text()|$letsdoeit
      Performers:
        Name: $actors//span/a[contains(@href,"/models/")]
      Studio:
        Name: $actors//a//text()
      Image: //source[@type="image/webp"]/@srcset
# Last Updated May 08, 2023
```
- Using the above all meta will be scraped from official site
- URL is not scraped so users that use the site for convenience will have their urls untouched, with proper meta though
- `doe-tv.com` urls should still be blocked/blacklisted by stashdb as all scenes are available from `official` site anyway and we can't be sure how the meta was scraped

PR #1312, opened 2023-04-06
Add JavDB_zh scraper

8ar10der:
Add a new scraper for gathering Chinese info from javdb.com


feederbox826:
Closing for the same reason as #1201 as it is just a set cookie

8ar10der:
Hi @feederbox826 ,

No, it is not easy work to let the script work by just changing the cookies. The element's key names in the Chinese web are totally different, they all will be changed to the Chinese.

If you compare with [javdb.yml](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/javdb.yml), you will find the answer.

feederbox826:
OK, will try to resolve the differences between the two

feederbox826:
Resolved and merged, I wasn't able to verify all the metadata, please report back

PR #1316, opened 2023-04-16
Add J&M Elite scraper

MortonBridges:
New URL scene and movie scraper for Jacquie et Michel Elite (https://www.jacquieetmichelelite.com/en/), with studio added to scraper list.


JackDawson94 (approved):
Tested - works as intended

PR #1323, opened 2023-04-24
Faphouse: Changed the details selector

ryosaeba75:


scrapers/FapHouse.yml
```diff
@@ -13,11 +13,11 @@ xPathScrapers:
         postProcess:
           - parseDate: "02.01.2006"
       Details:
-        selector: //div[contains(@class,"video-info-details__description")]/span
+        selector: //div[contains(@class,"video-info-details")]/div[@class="fh-text-with-collapse-multiline"]/p[@class="fh-text-with-collapse-multiline__text"]
         concat: " "
```
bnkai, line 17: From a quick look it seems that  now`//div[contains(@class,"video-info-details")]//p` should be enough.
Did you encounter any scenes with multiple paragraphs (we could leave the concat probably with a "\n" to be safe)?

ryosaeba75: I just check some scenes and I only encountered single paragraph, I'll update the selector.

JackDawson94 (approved):
Tested - works as intended ! Haven't found any multi-paragraph scenes either

PR #1327, opened 2023-04-25
Update Wankitnow.yml

BlokeCDP:
Changed year to 23 from 22.


BlokeCDP:
Corrected scraper so when no year is scraped it returns 2023 instead of 2022.

Created scraper based off existing scrapers with massive help from stg-annon.

JackDawson94 (changes requested):
DownblouseJerk.yml is in the wrong folder, it should be moved to /scrapers/  like all the others :)

Tested both - working as intended

PR #1328, opened 2023-04-28
Add Support for parasited.com

thefathacker:
Child Site of Hentaied.yml using same site Template. Added URL to existing Hentaied.yml to enable site Support


bnkai:
@thefathacker It seems that `details` need to be fixed as they are currently broken?
Something like 
```yaml
      Details:
        selector: //div[@class="fullss"]/p
        concat: "\n\n"
```
seems to work ok for all sites

thefathacker:
Thank you @bnkai, Replaced Details as requested

JackDawson94 (commented):
Tested - works as intended

JackDawson94 (approved):

PR #1332, opened 2023-05-02
update HobyBuchanon.yml

HeavyDinosaur:
Update scraper and change from xpath to json scraper


JackDawson94 (changes requested):
The json url does not use a static ID, the scraper is not functional.
Either you need to find a way to find the right ID when making the API call, or use the XPath scraper

PR #1333, opened 2023-05-03
Updating the xbvr scraper

Tweeticoats:
This version queries the xbvr api for matched scenes when XBVR_HOST is set to the ip and port of the instance. If XBVR_HOST=None it will query xbvr.db as before

### DIFF
--- a/scrapers/xbvrdb.py
+++ b/scrapers/xbvrdb.py
@@ -1,7 +1,20 @@
 import json
 import sys
 import sqlite3
-from os import path
+import os
+import requests
+
+try:
+    import py_common.graphql as graphql
+    import py_common.log as log
+except ModuleNotFoundError:
+    print("You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)", file=sys.stderr)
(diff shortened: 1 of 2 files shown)


scrapers/xbvrdb.py
```diff
@@ -48,26 +65,76 @@ def find_scene_id(title):
...
         if id is not None:
             return id[0]
     else:
         return id[0]
     return None
 
-if not path.exists("xbvr.db"):
+def query_api(filename):
+    request_config={"dlState":"available","cardSize":"1","lists":[],"isAvailable":True,"isAccessible":True,"isHidden":False,"isWatched":None,"releaseMonth":"","cast":[],"sites":[],"tags":[],"cuepoint":[],"attributes":[],"volume":0,"sort":"release_desc","offset":0,"limit":1}
+    response = requests.post(XBVR_HOST+'/api/scene/list', json=request_config)
```
bnkai, line 78: `requests` throws a lot of exceptions on dns, timeout, general connection errors. This shouldnt happen on same lan based hosts but doesnt hurt to avoid that with a try? Same for L85

PR #1334, opened 2023-05-06
New XPath scene scraper for apovstory.com

plz12345:
I wrote this one because the MissaX scraper doesn't cover it, and this studio doesn't format the scene pages the same way as missax.com .


JackDawson94 (changes requested):
Tested - works as intended

Can you please add the "Last updated on" line at the end as requested by the validator ?

plz12345:
@JackDawson94 sorry for the huge delay.  Comment added, and I did re-test it just to be sure it still works.

Maista6969:
The validator is very strict, the comment needs to be like this with a space after the `#` 😅
```
# Last updated August 24, 2023
```

plz12345:
One more try...

PR #1341, opened 2023-05-24
Added XPath scene and performer scraper for PornHex

imagineimaginingthings:


scrapers/PornHex.yml
```diff
@@ -0,0 +1,49 @@
...
+          - feetToCm: true
+      Measurements: $pornstar//span[contains(text(),"Measurements -")]/following::span/text()
+      URL: $pornstar//span[contains(text(),"Official site -")]/following::a/@href
+      Image:
+        selector: //section[@class="pornstar-header"]//div[@class="thumbnail-col-2 col-2"]/img[not(contains(@src,"default_portrait"))]/@src
+        postProcess:
+          - replace:
+              - regex: ^
+                with: "https://pornhex.com"
+
```
DogmaDragon, line 48: Unneccesary.

DogmaDragon:
Tested:
- Scenes work
  - Doesn't include studio, but they are focused around user uploaded content.
- Performers work

JackDawson94 (commented):
Tested
- Scenes: works as intended
- Performers: Height is not properly parsed

For the performer bug, here is a test url: https://pornhex.com/pornstars/abella-danger

On the page the performer Height is ```1.63 m```
When scraped, this is returned as ```191```

This is probably due to the "feetToCm" conversion in the code. I can't see any measurements in feet on the site (even when I switch languages).

imagineimaginingthings:
Ah, interesting, I had the opposite situation. I happened to test it on https://pornhex.com/pornstars/adriana-chechik whose height is in feet and inches and assumed all were like this. I do wonder if that field is simply a string and any data can be added - which would be very inconvenient.

Definitely need better parsing on height field then. I'm a little busy at the moment but will work on fixing this over the coming week so that both metric and imperial heights are correctly parsed. Thanks for pointing that out.

imagineimaginingthings:
Ok, their height field seems to accept any kind of string (some have extraneous whitespace), and after clicking through a lot of performers on their Pornstars page, have noticed that it's pretty much a 50/50 split between those in feet and inches and those in cm.

The regex for both of these is simple enough, the problem I am having is the `feetToCm` conversion. While it correctly does the conversion if a performer's height is in feet and inches, it ***also*** does the conversion on a metric figure. So using the example of Abella Danger, with my updated yml her height of 1.63m gets parsed as 163ft... converted to 4968cm. Yikes!

```
      Height:
        selector: $pornstar//span[contains(text(),"Height -")]/following::span/text()
        postProcess:
          - replace:
              - regex: ^(\d+)\.(\d+)\s?m
                with: $1$2
          - replace:
              - regex: ^((\d+)\'(\d+)\")
                with: $1
          - feetToCm: true
```

The first regex captures the value in metres and removes the decimal for a cm value, works fine on its own. The second regex captures a feet and inches value then converts it, also works fine on its own.
But the issue is I don't seem to be able to have both working in tandem.
Is there no way to only perform the feetToCm conversion on one of the regex matches, rather than both?
I'm not sure if it's technically impossible or if I'm getting the order of things, the indentation, or some other syntax wrong here.

Maista6969:
Going to merge this because there's no real solution to their data being in two different formats: we will just have to accept that users will have to fix their heights when the conversion fails

PR #1343, opened 2023-05-25
Add performer scraping to Kink scraper;

silveredgold:
I've tested as much as I can, and it seems to pass validation. That being said, XPath is not exactly my forte so always open to improvements.


DogmaDragon:
Tested:
- Performers
  - Search works
  - By URL works
    - Some issues with details parsing when it contains links (e.g. https://www.kink.com/model/51511/Mona-Wales)

silveredgold:
Thanks for testing, I'll see if I can clean up the results with the links, but that might be pushing my regex skills. 

Do you know of any other scrapers that might have solved a similar problem I could use for reference?

DogmaDragon:
`text()` strips all the HTML so it leaves empty spots where `<a></a>` was used. Workaround would be to don't use `text()` and try to concat different `<p></p>` or use some regex magic to remove irrelevant parts. It's a hard task since Kink is not consitent in their descriptions. 

Most of the cases where this happens are probably moved to Python as it's a lot easier to solve there. Can't think of any examples of the top of my head. But Even if unsolved, I think it can be merged as is or just removing `text()` and losing some formatting.

silveredgold:
Thanks for the tip. I messed around with using regex on the non-`text()` results and I think it's a little better. On that test case, it now completely strips the link-only "indent" list text, but results in much more sensible paragraph breaks.

Python is really not my strong suit, so if it needs porting to that, that's probably beyond me.

PR #1345, opened 2023-05-25
Improve algolia studio detection

nrg101:
The algolia.py parsing currently struggles to get the correct studio name for every single supported studio.

These changes adds some logic to use the Algolia API JSON to determine the correct studio.

Some additional domains are added as they are already sub-studios in stashdb.


JackDawson94 (commented):
Tested
- scene search
- scene by URL
- movie by URL

Overall it works, but there is an issue when importing Movies. The studio name that is returned is different from the one returned for the scenes.
Example:
Movie: https://www.devilstgirls.com/en/dvd/Dont-Tell-My-Wife-I-Buttfucked-Her-Trans-Friend/101236

Scene:
https://devilstgirls.com/en/video/devilstgirls/Dont-Tell-My-Wife-I-Buttfucked-Her-Trans-Friend---Gracie-Jane--James-Blow/219149

nrg101:
Nice spot. I've made `parse_movie_json(movie_json: dict)` call the function I added, `determine_studio_name_from_json`, which I'd already used for scenes and galleries. So movie scraping uses the same logic to get the studio name as scene scraping.

JackDawson94 (approved):
Tested again - works nicely !

PR #1349, opened 2023-05-30
Add search to Plushies.tv

Flashy78:


scrapers/Plushies.yml
```diff
@@ -22,12 +49,12 @@ xPathScrapers:
       Details:
         selector: $ub/span[@class="latest_update_description"]
       Image:
-        selector: //img[@class="stdimage large_update_thumb left thumbs"]/@src
+        selector: //img[@class="stdimage small_update_thumb thumbs"]/@src
```
DogmaDragon, line 52: I'm not sure how I feel about this change. But I noticed that some of the large ones are just blank and the small ones are used in the browse page. What's your reason behind it?

DogmaDragon:
Tested:
- Scenes work

scrapers/Plushies.yml
Flashy78: The large one is often blank, and it is just a screenshot from the start of the video. I'm guessing it's often blank because the intro in many of their videos has several black screen transitions. The small is an actual custom cover image and is always available.

JackDawson94 (commented):
Looks good !

JackDawson94 (approved):

PR #1351, opened 2023-06-01
feet: create JavLibrary_CN.yml

armnjc:
before the scraper feature supports i18n, provide a Chinese version of JavLibrary scraper


feederbox826:
Superceded by #1999

PR #1355, opened 2023-06-10
Added xPath sceneScraper for Xrares

imagineimaginingthings:
A very simple one as most the info is contained in the web page's `<meta>` tags.
All upload dates are in the format of "x days ago".
Using video ID for studio code.


JackDawson94:
I'm not familiar with the site could you please provide a couple links to run tests ?

JackDawson94 (approved):
Tested - works as intended.

Maybe you could also add a Studio field with the user name ? Not sure if that is relevant on this site ?

imagineimaginingthings:
> Maybe you could also add a Studio field with the user name ? Not sure if that is relevant on this site ?

I didn't think so, to be honest, but happy to add it if it's desired?
I only have a couple of videos ripped from this site in my collection but from what I can tell from a quick browse, the vast majority of content here appears to be uploaded by random users and pretty much all of it amateur. I don't see any evidence of verified pornstars or the like, or even people uploading their own content.

JackDawson94:
Makes sense to not have the Studio, indeed.

Good to go for me !

PR #1359, opened 2023-06-13
xpath, improvement: add tags/categories scraping to LetsDoeIt.yml

nrg101:
This adds another xpath matcher for Tags so that it works on transbella.com

E.g. https://transbella.com/watch/5536/gorgeous-latina-tranny-gets-pleased-by-two-cocks-in-wild-italian-threesome.en.html


JackDawson94 (commented):
Confirm this is working !

JackDawson94 (approved):

PR #1361, opened 2023-06-17
IFeelMyself: fix styled text in description missing spaces

Maista6969:
Scraping scenes where the description contains any tags (most notably `<i>`, `<b>`, `<strong>`) currently results in string concatenation without any spaces.

For example [this scene](https://ifeelmyself.com/public/main.php?page=flash_player&out=bkg&media_id=21833&artist_id=f15831) has the HTML `No one really <i>likes</i> doing chores [...]` which was turned into "No one reallylikesdoing chores [...]"

Telling `get_text` to use a space as the separator fixes this issue and produces "No one really likes doing chores [...]"


JackDawson94:
Tested - works as intended

PR #1365, opened 2023-06-19
Add scene scraping for MetalBondage.com

silveredgold:


scrapers/MetalBondage.yml
```diff
@@ -0,0 +1,39 @@
...
+    scraper: sceneScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //title/text()
+      Date:
+        selector: //div[@class="metabar"]/text()
+        postProcess:
+          - parseDate: January 2, 2006
```
Maista6969, line 15: This doesn't seem to work with any of the scenes I've tried, here's my suggested fix
```
Date:
  selector: //div[@class="metabar"]//text()
  postProcess:
    - replace:
        - regex: '(.*\d{1,2}, \d{4}).*'
          with: $1
    - parseDate: January 2, 2006
```

scrapers/MetalBondage.yml
```diff
@@ -0,0 +1,39 @@
...
+        Name:
+          selector: //div[@class="tags"]//a[@rel="category tag"]/text()
+      Image: //div[@class="textcontent"]//img/@src
+      Studio:
+        Name:
+          fixed: "Metal Bondage"
+      URL: //link[@rel="canonical"]/@href
+
+driver:
+  useCDP: true
```
Maista6969, line 38: As far as I can tell there's no need for a CDP here, the scraper works fine for me without it

Maista6969 (changes requested):
Everything looks good except the date parsing and arguably the CDP setting, both of which may be due to changes to the site layout since this PR was created

scrapers/MetalBondage.yml
silveredgold: Good catch, thanks! I tried a few more scenes and they did not work, but did with your solution which I'm now using.

scrapers/MetalBondage.yml
silveredgold: I thought I had it fail once without CDP but I can't reproduce now, so you might be right.

Maista6969 (approved):

PR #1366, opened 2023-06-19
Add Blackfoxbound UK Scene support

silveredgold:


scrapers/BlackFoxBound.yml
```diff
@@ -0,0 +1,31 @@
...
+      Details:
+        selector: //meta[@property="og:description"]/@content
+      Performers:
+        Name:
+          selector: //span[@title="Models"]//a/text()
+      Tags:
+        Name:
+          selector: //span[@title="Categories"]//a/text()
+      #Image: //div[@class="video-container"]//video/@poster
+      Image: //meta[@property="og:image"]/@content
```
Maista6969, line 25: You can have both of these if you OR them together, in case some scenes have video posters and some have opengraph image tags
`//div[@class="video-container"]//video/@poster|//meta[@property="og:image"]/@content`

Poking around at a few of these scenes it looks like `//div[contains(@id, "video")]/@poster` would also work and it seems to grab higher resolution images (unfortunately in many cases they're just upscaled)

Maista6969 (approved):
Works as expected

PR #1367, opened 2023-06-19
Add support for BeltBound.com

silveredgold:


scrapers/BeltBound.yml
```diff
@@ -0,0 +1,35 @@
...
+          - parseDate: January 2, 2006
+      Details:
+        selector: '//div[@class="post"]/p[2]'
+      Performers:
+        Name: '//a[@rel="tag"]'
+      Image: '//div[@class="post"]/a/img/@src'
+      Studio:
+        Name:
+          fixed: "BeltBound"
+
```
JackDawson94, line 34: This extra newline will get blocked by the verifier script

JackDawson94 (changes requested):
Tested - works as intended

Can you please also add it to Scrapers List ?

Also, have you run the verifier locally ? Just to be safe please run it once, otherwise when it is run on Github it will block your PR :)

PR #1370, opened 2023-06-23
xpath, improvement: add another xpath for image for Hustler

nrg101:
This adds another xpath selector for the existing Hustler.yaml scraper so that it can get the cover image for scene pages, e.g.

- https://www.hustler.com/model/charity-bangs/movie/20151116/black_cock_justice_pt__2
- https://www.hustler.com/model/madelyn-monroe/movie/20151123/black_cock_justice_pt__2

I left the existing selector in case there are different styles/layouts for scene pages where the existing selector will continue to work.

### DIFF
--- a/scrapers/Hustler.yml
+++ b/scrapers/Hustler.yml
@@ -39,19 +39,21 @@ sceneByURL:
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
+    common:
+      $content: //div[@class="panel-content"]
     scene:
-      Title: //h3/a/text()
+      Title: //h3/a/text()|$content//h3[contains(@class, "headline")]/a/text()
       URL: //link[@rel='canonical']/@href
       Performers:
         Name: 
-          selector: //span[@class="attr-key" and contains(text(),"Cast")]/following-sibling::span[@class="attr-value"]/a[not(text()="Hustler Models") and not(text()="Barely Legal Models")]/text()
+          selector: //span[@class="attr-key" and contains(text(),"Cast")]/following-sibling::span[@class="attr-value"]/a[not(text()="Hustler Models") and not(text()="Barely Legal Models")]/text()|$content//span[contains(text(), "Cast")]//following-sibling::span/a/text()
       Date:
-        selector: //span[@class="attr-key" and contains(text(),"Released")]/following-sibling::span[@class="attr-value"]/text()
+        selector: //span[@class="attr-key" and contains(text(),"Released")]/following-sibling::span[@class="attr-value"]/text()|$content//span[contains(text(), "Released")]//following-sibling::span/text()
         postProcess:
           - parseDate: Jan 02, 2006
-      Details: //meta[@property="og:description"]/@content|//div[@class="description"]/p
+      Details: //p[following-sibling::a[@class="clickable"]]|//meta[@property="og:description"]/@content|//div[@class="description"]/p
       Image:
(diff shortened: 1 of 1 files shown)


JackDawson94 (commented):
This works for the image as intended.

However most of the selectors are not working, when scanning with the test links above., Performers, Date, Details are not parsed.

nrg101:
I've tweaked the xpath selectors for Title, Date, Performers, and Details

|URL|Title|Performers|Date|Details|
|----|-------------|-----|--------|
|https://www.hustler.com/model/sahara-skye/movie/20230629/CAUGHT_MY_BUSTY_NEIGHBOR_MASTURBATING_3| ✔️ | ✔️  | ✔️ | ✔️ |
|https://www.hustler.com/model/charity-bangs/movie/20151116/black_cock_justice_pt__2| ✔️ | ✔️ | ✔️ | ❌ |
|https://www.hustler.com/model/madelyn-monroe/movie/20151123/black_cock_justice_pt__2| ✔️ | ✔️ | ✔️ | ❌ |
|https://www.hustler.com/model/mona-azar/movie/20230601/STACKED_MILFS| ✔️ | ✔️ | ✔️ | ✔️ |

I think older scenes with descriptions must have something about the pages that means Details can't be scraped in the same way.

In any case, the images work and I've now fixed all of the issues for new scenes and all but one (Details) of the issues for older scenes.

JackDawson94 (approved):
Indeed, it looks like for these old urls the only way to get the Details would be to have a subscraper call the JSON api... Would probably need to migrate to Python or Ruby for this. So I think this is the best we can do with the current XPath scraper

scrapers/Hustler.yml
```diff
@@ -39,18 +39,25 @@ sceneByURL:
...
-          selector: //span[@class="attr-key" and contains(text(),"Cast")]/following-sibling::span[@class="attr-value"]/a[not(text()="Hustler Models") and not(text()="Barely Legal Models")]/text()
+          selector: //span[@class="attr-key" and contains(text(),"Cast")]/following-sibling::span[@class="attr-value"]/a[not(text()="Hustler Models") and not(text()="Barely Legal Models")]/text()|$content//span[contains(text(), "Cast")]//following-sibling::span/a/text()
       Date:
-        selector: //span[@class="attr-key" and contains(text(),"Released")]/following-sibling::span[@class="attr-value"]/text()
+        selector: //span[@class="attr-key" and contains(text(),"Released")]/following-sibling::span[@class="attr-value"]/text()|$content//span[contains(text(), "Released")]//following-sibling::span/text()
         postProcess:
           - parseDate: Jan 02, 2006
-      Details: //meta[@property="og:description"]/@content|//div[@class="description"]/p
-      Image: //div[@class="img-container"]/img/@src
+      Details: //p[following-sibling::a[@class="clickable"]]|//meta[@property="og:description"]/@content|//div[@class="description"]/p
```
MrX292, line 54:
```suggestion
      Details: //div[@class="panel-content"]/div/div/text()|//meta[@property="og:description"]/@content|//div[@class="description"]/p
```

MrX292, line 54:
```suggestion
      Details: //div[@class="panel-content"]/div/div/text()|//meta[@property="og:description"]/@content|//div[@class="description"]/p
```
gets part of the details for older scenes

nrg101: hmm, when I try this with the new scenes, it also gets the first part (before clicking Read More) rather than the full description... will see if I can come up with a selector combo that gets the full description for all scenes

PR #1372, opened 2023-06-23
Add scene scraping for underhentai.net

escargotbuffed:
#                                                HOW TO SET UP                                                 #
Store this file in the ~/stash/scrapers/UnderHentai.yml (If the scrapers directory is not there it needs to be created).
#                                                  HOW TO USE                                                  #
#                                                   SCENES:                                                    #
The scene Scraper by Fragment is the best option in case the file name is the name of the anime.
Scenes that were not found can easily be found by the name scraper.
Don't put the episode number otherwise it won't find it.
It is also possible to scrape individually with the anime URL.
The scraper doesn't recognize the episode number, I recommend changing it manually at the end.
#                                              THAT'S IT, ENJOY!                                               #


JackDawson94 (approved):
Tested - works as intended

PR #1373, opened 2023-06-23
Add support for anidb.net

escargotbuffed:
#                                                HOW TO SET UP                                                 #
Store this file in the ~/stash/scrapers/AniDB.yml (If the scrapers directory is not there it needs to be created).
#                                                 SET COOKIES:                                                 #
This step is only necessary to access 18+ content.
Access the anidb.net website > login > right button > inspect > find cookies storage
Copy the "Value" of "adbsess" and "adbuin" and replace in the cookies category of this document.
If your account is new, you need to access any NSFW anime and confirm that you want to see 18+ content.
Do not change the order of the columns, as it can make it stop working.
I recommend creating a new account just for this.
#                                               SET USER AGENT:                                                #
Go to your Stash > Settings > Metadata Providers > Scraping > Scraper User Agent
Use the User Agent of your choice.
I'm currently using: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0
#                                                  HOW TO USE                                                  #
#                                                   SCENES:                                                    #
The scene Scraper by Fragment is the best option in case the file name is the name of the anime.
Scenes that were not found can easily be found by the name scraper.
Scene Name
It is also possible to scrape individually with the anime URL.
URL: https://anidb.net/anime/4563
Scene URL
The scraper doesn't recognize the episode number, I recommend changing it manually at the end.
#                                                 PERFORMERS:                                                  #
Performers need to be individually scraped by name or URL.
Performer Name
URL: https://anidb.net/character/308
Performer URL
I recommend creating them by scraping scenes and then searching individually by name.
#                                              THAT'S IT, ENJOY!                                               #

### DIFF
--- a/scrapers/AniDB.yml
+++ b/scrapers/AniDB.yml
@@ -128,7 +128,7 @@ driver:
         # Replace value field
         - Name: "adbsess"
           Domain: "anidb.net"
-          Value: "" # Enter the value of the 'adbuin' here
+          Value: "" # Enter the value of the 'adbsess' here
           Path: "/"
         - Name: "adbuin"
           Domain: "anidb.net"
--- a/scrapers/AniDB.yml
+++ b/scrapers/AniDB.yml
@@ -1,0 +1,137 @@
+name: AniDB
+################################################################################################################
+#                                                HOW TO SET UP                                                 #
+#                              Store this file in the ~/stash/scrapers/AniDB.yml                               #
+#                       (If the scrapers directory is not there it needs to be created)                        #
+#                                                 SET COOKIES:                                                 #
+#             Access the anidb.net website > login > right button > inspect > find cookies storage             #
+#       Copy the "Value" of "adbsess" and "adbuin" and replace in the cookies category of this document        #
+#    If your account is new, you need to access any NSFW anime and confirm that you want to see 18+ content    #
+#                    Do not change the order of the columns, as it can make it stop working                    #
+#                               I recommend creating a new account just for this                               #
+#                                               SET USER AGENT:                                                #
+#               Go to your Stash > Settings > Metadata Providers > Scraping > Scraper User Agent               #
+#                                      Use the User Agent of your choice                                       #
+#     I'm currently using: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0      #
+################################################################################################################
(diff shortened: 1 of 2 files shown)


JackDawson94:
Could you please add a description of the objective of this PR ? And maybe some urls we can use to test that it's working as intended ?

escargotbuffed:
> Could you please add a description of the objective of this PR ? And maybe some urls we can use to test that it's working as intended ?

Of course, I hadn't added it as it was inside the yml file as a comment.

JackDawson94 (commented):
Tested
- Performer by Name & Url - works as intended
- Scenes, the scanner works but I'm not sure about the logic.

From what I can tell, right now the "scenes" scanner goes to the master page for an anime.
However in most cases I assume if an anime has several episodes, this doesn't make sense, you would end up with the same data for each episode.

This is only my opinion and I'm not sure how the files you have match, but I would assume that
Scene = Episode on AniDB
Movie = Anime on AniDB

scrapers/AniDB.yml
```diff
@@ -0,0 +1,137 @@
...
+      
+driver:
+  cookies:
+    - CookieURL: "https://anidb.net/"
+      Cookies:
+        # Access adult content requires a anidb account
+        # Replace value field
+        - Name: "adbsess"
+          Domain: "anidb.net"
+          Value: "" # Enter the value of the 'adbuin' here
```
JackDawson94, line 131: I assume the comment here should be the value of 'adbsess' ?

escargotbuffed: yes, my mistake

escargotbuffed:
> Tested
> 
>     * Performer by Name & Url - works as intended
> 
>     * Scenes, the scanner works but I'm not sure about the logic.
> 
> 
> From what I can tell, right now the "scenes" scanner goes to the master page for an anime. However in most cases I assume if an anime has several episodes, this doesn't make sense, you would end up with the same data for each episode.
> 
> This is only my opinion and I'm not sure how the files you have match, but I would assume that Scene = Episode on AniDB Movie = Anime on AniDB

You're right, the scene scraper is working like this, as aniDB has almost no information about the episodes. All information is grouped on the anime page. The options are to leave the scene with almost no information, or not have the movie category. As hentai in general have an average of 1 or 2 episodes and I usually only look for scenes by tags, I preferred for the first option.

PR #1375, opened 2023-06-25
Update BangBros scraper for new site layout

reynn:


DogmaDragon:
Tested:
- Scenes work

JackDawson94 (approved):
Tested - works as intended.
Moving it to MindGeekAPI is a good idea but out of scope of this PR imo

PR #1376, opened 2023-06-27
Update Nubiles.yml to scrape avaliable resolution height as tags

estellaarrieta:


DogmaDragon:
I don't think that is a useful addition to majority of users.

JackDawson94 (commented):
I'm not opposed to the idea, but might be good to change the names to add a "p" at the end ? To fit the usual naming convention for resolutions ?


In terms of the value it brings, I don't see any, but I also don't see any harm from having it ?
If people are not interested in the tags, they will simply not grab these tags in the scrape. Nothing is forcing the tag to be applied.

What do you think @dogmadragon ?

DogmaDragon:
It would introduce the issue with submitting to StashDB (or other stash-box isntances depending on their guidelines) where it would submit all the resolution tags if the user has them matched, while StashDB only uses it as a reference for the highest available quality. 

I think having it locally for users who want it is enough and offers nothing of substance to the rest. If it were to be merged, it might be better idea to comment it out by deafult and let users who explicity want it to enable it.

Maista6969:
Closing this for the reasons stated by DogmaDragon, but also because they're not really tags from the studio

PR #1378, opened 2023-07-01
Update FetishPros scraping for current site layout

silveredgold:
The existing scraper doesn't seem to work on any scene URL I've tried, so I've updated it to match the current site layout.


JackDawson94 (approved):
Tested - works as intended

PR #1379, opened 2023-07-02
MenAtPlay: Fix line breaks in details

evandeaubl:
Seems that MenAtPlay has two different formats for the description block the scraper uses for details: one `<p></p>` per paragraph, and one big `<p></p>` with paragraphs split by `<br><br>`. The current scraper only worked for the former; all paragraph breaks get dropped with the latter. This change to the scraper makes it preserve paragraph breaks with both.


JackDawson94 (approved):
Tested - works as intended

PR #1380, opened 2023-07-03
Fix MenAtPlay validation failure

evandeaubl:
Oof, that will teach me not to run validation when "I'm sure it's all correct." 😞 This time, change passes validation. Sorry about that!


MrX292:
no problem i also did not see that

PR #1382, opened 2023-07-07
xpath, fix, improvement: transerotica.com - fixes the Date and Studio Name

nrg101:
The Date scraping does not currently work, as it selects the duration text. This change fixes the selector to grab the commented date text, and then extract the MM/DD/YYYY date.

The fixed studio name should be TransErotica, to match that on stashdb, so I've done that too.


Maista6969:
Tested on multiple scenes, works great! Looks like a lot of the scenes on StashDB could use a rescraping now that we can get proper dates :)

PR #1384, opened 2023-07-09
Improve WowNetworkVenus

WillyOrtrun:
- added filename schema parsing for files downloaded via the normal download buttons (still works with filenames, that do not fit the schema)
- results sorted by similarity taking performers into account if they could be determined from the filename, instead of the useless default sorting of the scraped search results
- fix no result issue, if there is a number in the scene name (removes numbers from search scraping request)
- added sceneByFragment/Scene Identify capabilities

It is unlikely, that I will find the time to implement changes the maintaners may want, It works for my usecases so basically I am leaving this here for anyone that wants to use it, or for anybody that has the time to upstream it. I may miss notifications, as this is obvsiously a throwaway account.

Potential Improvements (listing this here in case somebody else takes over):
- [x] Bring into naming conventions (artists -> performers)
- [ ] Change param[1] naming, currently in line with torrent.py, but quite confusing
- [ ] Get rid of interleave_results(), the sorting makes this unnecessary, but still calling the function
- [ ] Check if the new sys.argv[1] logic covers all usecases of the previous  if NAME / if URL logic
- [ ] remove debug log messages (?)

### DIFF
--- a/scrapers/WowNetworkVenus/WowNetworkVenus.py
+++ b/scrapers/WowNetworkVenus/WowNetworkVenus.py
@@ -5,6 +5,9 @@ import re
 import sys
 import urllib.parse
 from itertools import chain, zip_longest
+from difflib import SequenceMatcher
+import array as arr
+
 
 # to import from a parent directory we need to add that directory to the system path
 csd = os.path.dirname(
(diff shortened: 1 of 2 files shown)


feederbox826:
No scenes to test against but LGTM. 
Thanks!

PR #1385, opened 2023-07-11
move DogfartNetwork to the Algolia series of scrapers

reynn:
Tested against about 20 scenes in my stash instance and all working great.


scrapers/Algolia_DogfartNetwork.yml
```diff
@@ -0,0 +1,39 @@
+name: "DogfartNetwork"
+sceneByURL:
+  - action: script
+    url:
+      - dogfartnetwork.com/en/video
```
Maista6969, line 5: Could we add more substudios here while we're at it? 😃

Maista6969 (commented):
Looks good, will merge tomorrow unless you want to add support for more studios to `sceneByURL`

reynn:
@Maista6969 okay I think this is at a good place, scraper list has been updated as well. Noticed there was a separate scraper for `watchingmydaughtergoblack.com` since that is part of dogfart I removed the old scraper (wouldn't work now anyway).

PR #1392, opened 2023-07-19
an updated tokyohot scraper

awalkabout:
basic fragment scraper for stashdb scene tagger. Should work with N, N multipart, K, KB series. Should retrieve performer info.


Maista6969 (approved):
Tested this on the few Tokyo-Hot scenes I have, works great! I extended it to grab the english tags as well
```python
def get_tags(self):
    potential_tags = self.soup.find("div", {"class": "infowrapper"}).find_all("a", href=lambda h: "type" in h)
    return [a.text for a in potential_tags if a.text.isascii()]
```

Maista6969 (approved):

PR #1393, opened 2023-07-19
Create SmutPuppet scraper

echo6ix:


echo6ix:
Closing for same reason as other one. I think I branched this from the wrong version.

PR #1394, opened 2023-07-19
Improved/updated WeAreHairy scraped

echo6ix:
- Fixed broken xpath selectors
- Improved some xpath selectors to be less compex (less susceptible to breaking on site updates)
- Scraped studio name now matches StashDB
- Added galleryByURL scraper
- Added performerByURL scraper
  - Disabled performer image scrape by default since they are low res. Users can comment it out if they want to enable it


echo6ix:
I think I did something wrong.

PR #1395, opened 2023-07-19
New scraper: Futanarica

JackDawson94:
New scraper for Futanarica.com

Description is not perfect, didn't find a way to get the full text consistently, each page manages the description with a different path...
Image has the "Play button" watermark but that is the official image 🤦


nrg101:
all the scenes I spot checked have an affect3dstore.com link, e.g.

https://futanarica.com/my-personal-futa-trainer-plank-orgasm/
has link to https://affect3dstore.com/my-personal-futa-trainer-plank-orgasm.html

where the cover image is the same, but without the play button

you could maybe do a subscraper on the affect3dstore.com button to get the cover image

PR #1399, opened 2023-07-23
Add Redgifs Parser

WillyOrtrun:
- Redgifs videos do not have titles, so any search functions of this scraper are not real searches/only work with exact gif IDs or filenames
- It uses [Temporary tokens](https://github.com/Redgifs/api/wiki/Temporary-tokens) for the API, that do not require credentials. The user could run into rate limiting (maybe when identifying many scenes at once)
- Ignores gif description, because it is useless on 99,9% of gifs (mostly advertisement links)

### DIFF
--- a/scrapers/Redgifs.py
+++ b/scrapers/Redgifs.py
@@ -149,7 +149,10 @@ class Redgifs:
         )
 
 def parseFilename(filename):
-    return filename.split(".")[0]
+    id = filename.replace("redgifs_", "") #remove possible filename prefix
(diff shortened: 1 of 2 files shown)


Maista6969:
This works great, I'll merge it once you add a "Last Updated" comment to the bottom of `RedGifs.yml` to appease the validator

WillyOrtrun:
> This works great, I'll merge it once you add a "Last Updated" comment to the bottom of `RedGifs.yml` to appease the validator

done

PR #1400, opened 2023-07-24
fix/improvement, xpath: use subsite (sub-studio) at realitykings.com

nrg101:
Some sub-studio domains now redirect to the network site homepage, e.g.

https://www.momsbangteens.com/scene/4411883/
is now
https://www.realitykings.com/scene/4411883/do-not-fuck-around-with-my-stepmom

on the latter, current page, the (sub) studio is show in the "Subsite" section of the page. This change adds the text of the substudio link to the existing xpath union/alternatives

### DIFF
--- a/scrapers/MindGeek.yml
+++ b/scrapers/MindGeek.yml
@@ -174,7 +174,7 @@ xPathScrapers:
                 with: $1
       Studio:
         Name:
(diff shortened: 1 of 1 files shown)


nrg101:
another example:

https://www.milfhunter.com/scene/10316/
is now
https://www.realitykings.com/scene/10316/jerk-for-jane

PR #1402, opened 2023-07-27
Improved WeArehairy scraper

echo6ix:
- Added gallerybyURL scraping
- Updated and improved sceneByURL scraping
- Renamed yaml file using camel case


scrapers/WeAreHairy.yml
```diff
@@ -0,0 +1,68 @@
...
+        selector: //p[@id="height_both"]/text()
+        postProcess:
+          - replace:
+            - regex: .+(/d/d/d)cm
+              with: $1
+      Weight:
+        selector: //p[@id="weight_both"]/text()
+        postProcess:
+          - replace:
+            - regex: .+(/d/d)kg
```
Maista6969, line 66: `/d` isn't a valid meta-sequence, the backend must handle it for you 😅

This only finds two digits, so it will fail for any bigger models that weigh more than 100kg: this might never occur on the site, I don't know them well enough, but the regex could be more robust: `.*(\d{2,}) kg` will capture any kilogram weight of 2 digits or more

scrapers/WeAreHairy.yml
```diff
@@ -0,0 +1,68 @@
...
+      Gender:
+        fixed: Female
+# Uncomment to scrape performer image
+#      Image:
+#        selector: //div[@class="lhs"]/img/@src
+      Height:
+        selector: //p[@id="height_both"]/text()
+        postProcess:
+          - replace:
+            - regex: .+(/d/d/d)cm
```
Maista6969, line 60: `/d` is not a valid meta-sequence, the backend must handle it for you 😅

I suggest using `.*(\d{3}) cm` here

SCRAPERS-LIST.md
```diff
@@ -1417,7 +1417,7 @@ watchreal.com|Hustler.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
 watchyoujerk.com|WatchYouJerk.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 waybig.com|Waybig.yml|:heavy_check_mark:|:x:|:x:|:heavy_check_mark:|-|-
 wcpclub.com|AdultEmpireCash.yml|:heavy_check_mark:|:x:|:heavy_check_mark:|:x:|-|-
-wearehairy.com|wearehairy.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
+wearehairy.com|wearehairy.yml|:heavy_check_mark:|:heavy_check_mark:|:x:|:x:|-|-
```
Maista6969, line 1420: Needs a checkmark for the fourth column as well since you've added Performers to this scraper

You can also change the entry in the second column to indicate that you've changed the casing of the file to `WeAreHairy.yml`

Maista6969 (changes requested):
Tested with several scenes, galleries and models: works great, only nitpicks in the scraper

Needs to fully update `SCRAPERS-LIST.md` before merging

Maista6969 (approved):

PR #1403, opened 2023-07-27
Created AuntJudysXXX scraper

echo6ix:


scrapers/AuntJudysXXX.yml
```diff
@@ -0,0 +1,24 @@
...
+  - action: scrapeXPath
+    url: 
+      - auntjudysxxx.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Studio:
+        Name:
+          fixed: Aunt Judy's XXX
```
Maista6969, line 12: I suggest replacing this to support Aunt Judy's as well:
```
Studio:
  Name:
    selector: //base/@href
    postProcess:
      - replace:
          - regex: ^https?://(?:www\.)?(.+)\.com/.*$
            with: $1
      - map:
          auntjudys: Aunt Judy's
          auntjudysxxx: Aunt Judy's XXX
```

Maista6969 (approved):
Noticed that this site has the same structure as `auntjudys.com` so I suggest that you include this as well 😄

echo6ix:
> Noticed that this site has the same structure as `auntjudys.com` so I suggest that you include this as well 😄

Good idea.

- Now scrapes auntjudys.com using @Maista6969's suggestions
- Renamed yaml to `AuntJudys.yml` (becausee AuntJudyXXX is derivative of AuntJudys)
- Updated scrapers list to reflect the above changes

Read for review.

PR #1407, opened 2023-07-27
Updating/improving The Score Group scraper

echo6ix:
Additions:
- ~~Sanitizes the URL by removing any referral codes~~
- Updates canonical URL without tracking parameters
- Gets studio code

Improvements:
- Scrapes the primary scene image from the studio URL
- Other minor xpath tweaks


scrapers/TheScoreGroup.yml
```diff
@@ -30,20 +33,44 @@ xPathScrapers:
...
-      Details:
-        selector: //div[@class="row"]/div/div[@class="p-desc"]/text()
-        concat: "\n"
+      Details: &details
+        selector: //div[@class="p-desc p-3"]//text()
+        postProcess:
+          - replace:
+              - regex: Read More »
+                with:
+        concat: "\n\n"
```
Maista6969, line 42: This leads to some funny extra paragraphs for any descriptions that use HTML tags for things like italicized text, see [this XL Girls scene](https://www.xlgirls.com/bbw-videos/Juliana-Rodriguez/72936/) for instance

It's better than the old scraper, though, because that didn't return anything at all for this scene

scrapers/TheScoreGroup.yml
```diff
@@ -30,20 +33,44 @@ xPathScrapers:
...
+#            - replace:
+#                - regex: _lg
+#                  with: _x_800
+      URL: &urlAttr
+        selector: $url
+        # Sanitize the URL
+        postProcess:
+          - replace:
+              - regex: (https?:\/\/(?:www\.)?[0-9A-za-z-]+\.com\/[0-9A-Za-z-_]+\/[0-9A-Za-z-_]+\/\d+)\/?.*
+                with: $1
```
Maista6969, line 67: The post-processing should not be necessary here: the canonical URL by definition does not contain any tracking parameters so we can use it as it's written

scrapers/TheScoreGroup.yml
```diff
@@ -30,20 +33,44 @@ xPathScrapers:
...
+      Code: &codeAttr
+        selector: $url
+        postProcess:
+          - replace:
+              - regex: https?:\/\/(?:www\.)?[0-9A-za-z-]+\.com\/[0-9A-Za-z-_]+\/[0-9A-Za-z-_]+\/(\d+)\/?.*
```
Maista6969, line 72: Can be simplified to `(\d+)\/?$` 😃

Maista6969 (approved):
Looks good for most of Score Group, let me know if you don't want to bother with the nitpick I had about descriptions and I can merge this

scrapers/TheScoreGroup.yml
echo6ix: Manually obtaining a URL from their site always contains the tracking parameters in my experience. Users could sanitize it themselves but I'm betting they're not going to bother. This is why I'm sanitizing it in the scraper.

Maista6969: I believe all of their anchor tags have tracking parameters, but none of the pages I've checked have had tracking parameters in the `<link rel="canonical">` tag: I only bothered to point this out because they also consider the trailing slash to be canonical and your reconstruction omits that 😅

echo6ix: Oh, snap. I understand what you're communicating now. Sanitizing the URL is uncessary because I am already obtaining the clean version from `//link[@rel="canonical"]/@href`. 🙃😂 Major brainfail there.

I think the sanitization bit was included because as I updated my very old scraper I forgot to omit that vestigal section and just pieced it together with the new one. 💩

scrapers/TheScoreGroup.yml
echo6ix: Yes, I caught that too and could not think of a way to address it. That issue is a tricky one.

echo6ix: Figured out a way to fix the "extra paragraphs". All other issues addressed. Scraper should be ready to review.

PR #1409, opened 2023-07-27
fix: move brazilian-transsexuals.com and tgirlsfuck.com to other Grooby scraper

nrg101:
The brazilian-transsexuals.com and tgirlsfuck.com sites have changed their HTML layout code which now uses the same CSS classes as in the majority of other Grooby sites.

This change just moves the domains over. The braziltgirls.xxx site still uses the old layout style.


Maista6969:
Tested on scenes from the two affected networks and it works great! I hope to set up some kind of tooling that will let me test these network-level scrapers more efficiently, because I haven't had the capacity to test every network in this file but I trust that users will get loud if something breaks 😅

PR #1410, opened 2023-07-27
Updated scraper list for Tug Pass sites

echo6ix:


Maista6969:
Thanks for catching this, I was a bit too quick on the trigger with #1408

PR #1411, opened 2023-07-27
Fix name formatting for LoveHerFeet.com

Maista6969:
Also rename the YAML file to LoveHerFilms.yml to match the network name.

Essentially recreating #1374


DogmaDragon:
Thanks for following up. I messed up my branches so I closed https://github.com/stashapp/CommunityScrapers/pull/1374 and forgot to follow up.

PR #1415, opened 2023-07-31
Update Bang scraper

WithoutPants:
Fixes a bunch of outdated fields in the Bang scraper. 

The performer twitter field was removed as I can't see any performers with this field populated. The movie and studio fields come out uncapitalised.


Maista6969 (approved):
Works with everything I've thrown at it! Would you mind updating the comment at the bottom of the scraper?

PR #1423, opened 2023-08-04
Made screne scrapper actually scrape scenes and added functionaly to …

freeagent1384:
…pull performer image

### DIFF
--- a/scrapers/AdultEmpire.yml
+++ b/scrapers/AdultEmpire.yml
@@ -1,22 +1,36 @@
 name: AdultEmpire
 
+performerByName:
+  action: scrapeXPath
+  queryURL: https://www.adultempire.com/performer/search?q={}
+  scraper: performerSearch
+
+performerByURL:
+  - action: scrapeXPath
+    url: 
+      - www.adultempire.com
+    scraper: performerScraper
+
 movieByURL:
   - action: scrapeXPath
     url:
       - adultdvdempire.com
       - adultempire.com
     scraper: movieScraper
+
 sceneByURL:
   - action: scrapeXPath
     url:
       - adultdvdempire.com
       - adultempire.com
     scraper: sceneScraper
+
 sceneByName:
   action: scrapeXPath
   # if needed replace `dvd` with `vod`
   queryURL: https://www.adultdvdempire.com/dvd/search?q={}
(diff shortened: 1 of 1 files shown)


freeagent1384:
Later today I'll see if I can't differentiate between the scene and the full length movie and create scene scrapers that will work for both so people can have their cake and eat it too and you don't have to worry to much about breaking peoples workflow.

freeagent1384:
I'm very close to being able to pull an image for the scene scraper but I don't know of a way to use the URL of the page being scraped as a variable.  I would then need to parse the URL for the URI as a token used in the xPath in the subScraper.  Is this possible or do you have a better suggestion?

Maista6969:
> I'm very close to being able to pull an image for the scene scraper but I don't know of a way to use the URL of the page being scraped as a variable. I would then need to parse the URL for the URI as a token used in the xPath in the subScraper. Is this possible or do you have a better suggestion?

Stash doesn't give us the URL we're currently scraping, but we can usually find it in the page! 😄 Looking at adultempire.com they do the very common thing of having the canonical URL of a page in the head element: `/html/head/link[@rel="canonical"]/@href`

freeagent1384:
> /html/head/link[@rel="canonical"]/@href

That is true in the actual movie page but unfortunately not true in the scene page.  I searched the html of https://www.adultempire.com/clip/1619049/scene-1-streaming-scene.html for example and no where in the page does clip/1619049/scene-1-streaming-scene.html exist.  I was thinking I might be able to reconstruct the URI, I can grab the ID for example but "scene-1" isn't in the page.  I can use the title for this piece but would require postProcessing to add the dashes.

I suspect this will have to be done in Python to get the image and I haven't written in Python in years.  Unless you have a suggestion on how to discover that URI from the scene page I'm afraid I'm going to have to table the image extraction.

freeagent1384:
I"m really close to having everything working the way you requested but I have one issue I'm hoping you might have some advice on.  

In order to get the image from the subScraper I need to provided a Scene ID as a part of the xPath but it would appear I cannot provide a "Common" variable as a part of the xPath. You can see in my last commit that fails to build line 163 "selector: (//a[@scene_id="$sceneId"])[1]/@href" Do you have any creative way of making this happen?

PR #1427, opened 2023-08-04
Added scene by name and URL also movie by URL for FreeonesCommunity

freeagent1384:


Maista6969 (changes requested):
Please update `SCRAPERS-LIST.md` with these new capabilities :)

Maista6969 (approved):

PR #1431, opened 2023-08-04
Search by scene name added but for some reason the image is not being…

freeagent1384:
… picked up.  From what I can tell the xPath is good but no image appears in my UI.

### DIFF
--- a/scrapers/VRBangers.yml
+++ b/scrapers/VRBangers.yml
@@ -2,6 +2,16 @@
 
 name: "VRBangers"
 
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://vrbangers.com/search/?s={}
+  scraper: sceneSearch
+
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
+  scraper: sceneScraper
+
 sceneByURL:
   - action: scrapeXPath
     url:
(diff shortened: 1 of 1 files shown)


DogmaDragon:
It uses JS for LazyLoad so you probably need to use [sub-scraper](https://docs.stashapp.cc/in-app-manual/scraping/scraperdevelopment/#post-processing-options) post-processing.

Maista6969:
Looks like their search hits an API, might be easier to use that directly with a JSON parser instead of the XPath... they return the published date as a UNIX timestamp, but that's supported since stashapp/stash#2817

https://content.vrbangers.com/api/content/v1/search/call%20on%20me

freeagent1384:
> It uses JS for LazyLoad so you probably need to use [sub-scraper](https://docs.stashapp.cc/in-app-manual/scraping/scraperdevelopment/#post-processing-options) post-processing.

I'm not familiar with LazyLoad, I'll read up on it this weekend. @DogmaDragon, do you know of another scraper that has to address a similar issue?

freeagent1384:
> Looks like their search hits an API, might be easier to use that directly with a JSON parser instead of the XPath... they return the published date as a UNIX timestamp, but that's supported since stashapp/stash#2817
> 
> https://content.vrbangers.com/api/content/v1/search/call%20on%20me

Thanks @Maista6969, that definitely is a very viable option. I'm fascinated to know how you knew it was an API? Could be very useful in other scrapers.

Maista6969:
> > Looks like their search hits an API, might be easier to use that directly with a JSON parser instead of the XPath... they return the published date as a UNIX timestamp, but that's supported since [stashapp/stash#2817](https://github.com/stashapp/stash/pull/2817)
> > https://content.vrbangers.com/api/content/v1/search/call%20on%20me
> 
> Thanks @Maista6969, that definitely is a very viable option. I'm fascinated to know how you knew it was an API? Could be very useful in other scrapers.

I ran a search with the Developer Tools open and checked the Network tab: a lot more sites are using APIs these days and I got lucky with this one 😃

Developer Tools

PR #1436, opened 2023-08-09
Created scraper for rachel-steele.com as requested in  Scrapers Request #73

freeagent1384:


BlokeCDP:
Hi, thanks for this. Some feedback.

Tried this with:
https://rachel-steele.com/x-new/new-set.php?setid=110165
- Studio returned Rachel-Steele.com - StashDB's studio is Rachel Steele - better to scrape as Rachel Steele for consistency and less manual edits?
- Animated GIF is scraped - not sure what can/should be scraped instead.
- Description not scraped - <meta name="twitter:description" content=" has the description

Hope this is of some use.

Maista6969:
The studio doesn't seem to offer anything but the animated GIFs, so we'll just have to scrape them and let users decide whether or not they prefer to create their own thumbnails 👍

BlokeCDP:
Fair enough on the animated gif front, but it's not pulling the description across?

PR #1441, opened 2023-08-15
Created Babes.com scraper and updated SCRAPERS-LIST to reflect changes

freeagent1384:
SCRAPERS-LIST indicated this site was covered by Brazzers scraper but I found no reference to it in Brazzers.yml

### DIFF
--- a/scrapers/Babes.yml
+++ b/scrapers/Babes.yml
@@ -170,7 +170,7 @@ xPathScrapers:
       Measurements: $root//span[contains(text(),"Measurements:")]/following-sibling::span/text()
       Gender:
         fixed: "Female"
-      HairCollor: $root//span[contains(text(),"Hair Color:")]/following-sibling::span/text()
+      HairColor: $root//span[contains(text(),"Hair Color:")]/following-sibling::span/text()
   sceneScraper:
     common:
       $data: //div[contains(concat(' ',normalize-space(@class),' '),' hAewsL ')]
--- a/scrapers/Babes.yml
+++ b/scrapers/Babes.yml
@@ -1,0 +1,224 @@
+name: Babes
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - www.babes.com/scene
+    scraper: sceneScraper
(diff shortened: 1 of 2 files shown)


scrapers/Babes.yml
```diff
@@ -0,0 +1,224 @@
...
+#      Name: //a[contains(concat(' ',normalize-space(@class),' '),' one-list-hueuj4 ')]/text()
+#      URL:
+#        selector: //a[contains(concat(' ',normalize-space(@class),' '),' one-list-hueuj4 ')]/@href
+#        postProcess:
+#          - replace:
+#              - regex: ^
+#                with: https://www.babes.com
+  performerScraper:
+    common:
+      $root: //div[contains(concat(' ',normalize-space(@class),' '),' cPKtEE ')]
```
Maista6969, line 36: These CSS classes are regenerated every time the website deploys a new version so we cannot rely on them staying stable for any amount of time

Maista6969 (commented):
I don't think we can merge this as is, but I also feel like the layout of Babes.com is familiar enough that we can probably get what we want from some other scraper without duplicating all of this effort: I suspect the MindGeek scraper refactor might have us covered, but I'll get back to you on that

scrapers/Babes.yml
freeagent1384: Thanks for the info Maista. I did read your other reply and will totally stand by while you guys decide what to do. Out of interest, do you know of a better or alternate method for extracting the desired material not relying on the CSS class?

Maista6969: It's tricky and arguably less readable, but we can rely on the structure of the page instead: this means using the kinds of XPath selectors you get from right-clicking an element in Developer Tools and choosing `Copy > XPath`. Here's an example of selectors for the "scene element" at babes.com:
```
/html/body/div/div[1]/div[2]/div[2]/div[2]/div[1]
```

So if we use that as a common fragment we can select the description using this selector
```
$scene/div[2]/section/div/p
```

PR #1442, opened 2023-08-16
AnalVids: Update / Fix and Rename

woodgen:
Picking up  a scraper  I initially wrote.

This patch set renames `LegalPorno` to `AnalVids`, which was used in the URLs already. `LegalPorno` is the old name and redirects.

The main reason for this PR is an update to recent site changes, which broke the old scraper.
The Python script still worked perfectly and was just renamed.

While I was at it, I improved the scraper a bit, adding an alternative URL and also scraping performer / studio URLs for new entries.

The performer scraper was also updated in addition to the scene one.


Maista6969:
Thank you for this fix! 👍

PR #1448, opened 2023-08-22
Update Clips4Sale.yml

zzazzcdts:
New version of Clips4Sale scraper that is compatible with their new site design.


Maista6969:
Tested with every c4s link I could get my hands on and this works great, thank you for contributing! 👍

PR #1457, opened 2023-09-04
Created Naughty Natural scraper

echo6ix:
This creates a scene and gallery scraper for naughtynatural.com

The scraper does not scrape a scene image. The xPath scene scraper correctly parses the scene's jpg file, however the server returns the following:

```
<Error>
<Code>AccessDenied</Code>
<Message>Access Denied</Message>
<RequestId>W708A9V1FYCYGYMP</RequestId>
<HostId>Zxrl4lZH6o9mKpN2kJeAaWPpigw21xFcNk9sH7TIkQdf2fZnOoOKzrGdZ/xdCIzFm+WybHr8img=</HostId>
</Error>
```

This is beyond my scope of expertise. Perhaps someone with more advanced skills can figure it out, if anyone even uses this scraper.

### DIFF
--- a/scrapers/NaughtyNatural.yml
+++ b/scrapers/NaughtyNatural.yml
@@ -1,0 +1,54 @@
+name: Naughty Natural
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - naughtynatural.com/videos/
+    scraper: sceneScraper
+galleryByURL:
+  - action: scrapeXPath
+    url:
+      - naughtynatural.com/photos/
+    scraper: galleryScraper
+xPathScrapers:
+  sceneScraper:
+    common:
+      $url: //link[@rel="canonical"]/@href
+    scene:
+      Title: &title //h1[@class="entry-title"]/text()
+      Studio: &studio
+        Name:
+          fixed: Naughty Natural
+      Date: &date
+        selector: //meta[@property="article:published_time"]/@content
+        postProcess:
+          - replace:
+              - regex: (^\d{4}-\d{2}-\d{2}).*
+                with: $1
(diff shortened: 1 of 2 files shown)


scrapers/NaughtyNatural.yml
```diff
@@ -0,0 +1,54 @@
...
+        Name: //span[@class="categories"]/a
+      Performers: &performers
+        Name: //span[@class="post_author"]/a
+      # Unable to scrape image! Xpath correctly isolates jpg file, however AWS server blocks scraping of image
+      #Image:
+      #  selector: //script[@type="text/javascript"][contains(text(),"jwplayer")]
+      #  postProcess:
+      #    - replace:
+      #        - regex: .*(https.+jpg).*
+      #          with: $1
```
Maista6969, line 40: I noticed that while they are using AWS to store these images, they do also store it right on their CMS as well... we can do heinous regex stuff to grab it from there instead! If we grab the path from the first URL (the image) and the domain from the second URL (the video) and Frankenstein them together then it resolves and we can scrape the image

```
Image:
  selector: //script[@type="text/javascript"][contains(text(),"jwplayer") and contains(text(),"jpg")]
  postProcess:
    - replace:
        - regex: .*(wp-content.+jpg).*(https:\/\/[^\/]*).*
          with: $2/$1
```

Maista6969 (approved):
Looks good to me! I have a suggestion for how to grab the image so I'll wait for you to consider that and give the go-ahead before I merge this

scrapers/NaughtyNatural.yml
echo6ix: Perfect. Done.

PR #1462, opened 2023-09-12
Revised WankItNow (and associated sites) scraper

BlokeCDP:
Removed existing scrapers replaced by the revised WankItNow scraper. TODO added to scraper at request.


Maista6969:
Thank you, I love when we can consolidate scrapers! 👍

PR #1467, opened 2023-09-23
Fix Colby Knox scraper for new website design

evandeaubl:
Colby Knox launched a new website design a couple weeks ago, and naturally that broke the scraper. This PR fixes it for the new website design. Checked it against content of various vintages, and it worked okay on all of them. A couple of the selectors are a little less futureproof than I'd like, but the new site design doesn't have a lot of semantic classes or ids, so here we are...


Maista6969:
Thank you for the quick fix! Tested this and it's worked for everything I've thrown at it 👍

PR #1470, opened 2023-09-25
Fix PornHub sceneScraper

Diliu94:
Updates to make pornhub scraper work again

### DIFF
--- a/scrapers/Pornhub.yml
+++ b/scrapers/Pornhub.yml
@@ -112,7 +112,8 @@ xPathScrapers:
       Image: //div[@class="thumbImage"]/img/@src|//img[@id="getAvatar"]/@src
   sceneScraper:
     common:
-      $datablob: //script[contains(., 'videodata')]/text()
+      $datablob: //script[contains(., 'VideoObject')]/text()
+      $videowrap: //div[@class="video-wrapper"]
     scene:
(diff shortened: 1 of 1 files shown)


Maista6969:
Unclear on what this is fixing, can you provide a link for a video the old scraper is failing for? This could be a geolocation issue but the old scraper and your version produce identical results for me

Diliu94:
Huh thats odd. The old scraper does not give me the full data for any video because the videodata object that is loaded as $datablob does not exist, so it only returns Title, URL and Image. The VideoObject that is available instead does not have the same structure and does not contain as much information to scrape.

As you say it is possible that pornhub is rolling out new versions of the site gradually over the world. I would estimate the old scraper has been failing for me for at least a few weeks, maybe a couple of months now.

Maista6969:
I was going to say that I'd merge this anyway because both scrapers seemed to work the same, but that was an oversight on my part: the existing scraper works for me, but this version doesn't retrieve the image... we'll definitely have to keep an eye on Pornhub and I hope we can get more feedback from other users to figure out which way they're going with this

Ziron:
Sorry! That was a small error caused by careless copy-pasting as the last action!
The image scraping should be unchanged from the existing version and I have fixed that now

Maista6969:
Thanks, I should have spotted that 😅

PR #1480, opened 2023-10-03
Add BoyCrush network scraper

evandeaubl:
This adds a scraper for the two main BoyCrush network sites that have public scene pages: boycrush.com and baretwinks.com.

Both sites have public pages for the scenes of all 5 sites under the BoyCrush network, so the scraper will read which site the video is supposed to be a part of and assign the proper studio. The other 3 sites do not have public scene pages, so URLs for them are not added to the scraper. I figure the idea will be if you want a public link for those sites, you will use the boycrush.com one, because that is the main network site at this point.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -164,6 +164,7 @@ bangteenpussy.com|Teencoreclub.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 barbarafeet.com|FFCSH.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 barebackplus.com|CarnalPlus.yml|:heavy_check_mark:|:x:|:x:|:x:|-|Gay
 barelylegal.com|Hustler.yml|:heavy_check_mark:|:x:|:x:|:x:|CDP|-
+baretwinks.com|BoyCrush.yml|:heavy_check_mark:|:x:|:x:|:x:|-|Gay
 bathhousebait.com|AMAMultimedia.yml|:heavy_check_mark:|:x:|:x:|:x:|-|Gay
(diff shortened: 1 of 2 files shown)


Maista6969:
Works great, but I ran into an issue trying to scrape [Angel & Kevin Get It On](https://tour.baretwinks.com/bare-twinks-watch/bt025_angel_kevin/Angel+%2526+Kevin+Get+it+On) from BareTwinks.com because it uses a `<video>` element with the `poster` attribute instead of an `` tag: I have no idea how common this is, and I will still merge this if you don't consider it a blocking issue 👍

evandeaubl:
Good catch. I’ll check it out this afternoon and let you know.

evandeaubl:
Apparently the scene pages have those two different formats somewhat randomly (at least there doesn't appear to be a pattern, like newer are in one format, older in the other), but both formats are common enough to be worth handling both. I guess I was just "lucky" to get all of one when I was testing. 🤷

Fixed and tested against the two formats on both BoyCrush and BareTwinks using:

(//video/@poster format)
https://tour.baretwinks.com/bare-twinks-watch/bc941_jackbailey_interview/Meet+Sexy+Young+Twink+Jack
https://tour.boycrush.com/boy-crush-watch-gay-movies/bc941_jackbailey_interview/Meet+Sexy+Young+Twink+Jack

(//img/@src format)
https://tour.boycrush.com/boy-crush-watch-gay-movies/bc705_andytaylor_rykermadison/Ryker+Fucks+The+Hot+Ass+Of+Andy
https://tour.baretwinks.com/bare-twinks-watch/bc705_andytaylor_rykermadison/Ryker+Fucks+The+Hot+Ass+Of+Andy

Maista6969:
Excellent, thank you for your contribution 🙂

PR #1482, opened 2023-10-03
Fixed issue in Baberotica.yml with Studio name map

smcallah:
Studio name mapping was not working, the URL would appear as the Studio name when scraping.  

URLs under "- map:" need the trailing "/" in order for the mapping to the name to work.

### DIFF
--- a/scrapers/Baberotica.yml
+++ b/scrapers/Baberotica.yml
@@ -36,8 +36,8 @@ xPathScrapers:
           selector: //meta[@itemprop="url"]/@content
           postProcess:
             - map:
-                https://avidolz.com: AvIdolz
-                https://baberotica.com: Baberotica
-                https://baberoticavr.com: BaberoticaVR
-                https://teenthais.com: TeenThais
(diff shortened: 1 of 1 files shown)


scrapers/Baberotica.yml
```diff
@@ -36,8 +36,8 @@ xPathScrapers:
...
-                https://avidolz.com: AvIdolz
-                https://baberotica.com: Baberotica
-                https://baberoticavr.com: BaberoticaVR
-                https://teenthais.com: TeenThais
-# Last Updated June 30, 2021
+                https://avidolz.com/: AvIdolz
+                https://baberotica.com/: Baberotica
+                https://baberoticavr.com/: BaberoticaVR
+                https://teenthais.com/: TeenThais
+# Last Updated October 3, 2023
```
Maista6969, line 43: This has the same issue as your other PR, the date needs to be "October 03, 2023" 😅

Maista6969 (changes requested):

Maista6969 (approved):

PR #1486, opened 2023-10-04
Fix Blake Mason scraper

evandeaubl:
Blake Mason had a site redesign in April; this fixes the scraper to work with that new design, and also adds extracting the studio code.

Tested against a couple of newer (circa last year) scenes, and a couple of old scenes, and it seems to handle both and their minor differences.


Maista6969:
Worked with everything I threw at it! Thank you for this contribution 🙂

PR #1487, opened 2023-10-05
Added performer scraper and fixed scene scraper issues with Baberotica.yml, added 2 new sites, and updated SCRAPERS-LIST.md

smcallah:
Added performer scraper and fixed scene scraper issues with Baberotica.yml and updated SCRAPERS-LIST.md to add the check mark for Performer scraper to Baberotica sites.


Maista6969:
Worked with every performer I tried across all sites: very nicely done! For the future you can avoid using two regular expressions for hair color by using an inline flag to make it case insensitive: `(?i).*hair color:?\s*(.*)\s*$` 🙂

PR #1490, opened 2023-10-08
Update MatureNL.yml

1-gran:
Fix image selector

### DIFF
--- a/scrapers/MatureNL.yml
+++ b/scrapers/MatureNL.yml
@@ -53,7 +53,7 @@ xPathScrapers:
                   with: $1,
           split: ", "
       Image:
-        selector: (//video/@poster | //span[@id="spnPageUpdateTrailer"]/a/img/@data-src | //img[@class="img-responsive lazy"][1]/@data-src)[1]
+        selector: //video/@poster or //span[@id="spnPageUpdateTrailer"]/a/img/@data-src or //img[@class="img-responsive lazy"][1]/@data-src
       Date: &date
         selector: //span[@class="val-m"][1]
         postProcess:
@@ -72,4 +72,4 @@ xPathScrapers:
       Performers: *performers
       Date: *date
       Studio: *studio
-# Last Updated July 11, 2023
+# Last Updated October 08, 2023


Maista6969:
I can't see what this is fixing, could you provide an example URL that won't scrape with the current selector?

1-gran:
> I can't see what this is fixing, could you provide an example URL that won't scrape with the current selector?

Any URL. Im not sure why, but your code will only fetch the image succesfully once. Then if you scrape another URL it wont work. If I restart the app then it will work again, but only once.
Maybe its just on my end, and theres some weird caching going on, I dont know. I tried on Chrome and Firefox.
My PR fixes the issue for me atleast.

Maista6969:
> > I can't see what this is fixing, could you provide an example URL that won't scrape with the current selector?
> 
> Any URL. Im not sure why, but your code will only fetch the image succesfully once. Then if you scrape another URL it wont work. If I restart the app then it will work again, but only once. Maybe its just on my end, and theres some weird caching going on, I dont know. I tried on Chrome and Firefox. My PR fixes the issue for me atleast.

You are quite right that this is a caching issue with the xpath library that Stash depends on: the PR with a fix [has been merged](https://github.com/stashapp/stash/pull/4126) so this selector (and many others like it) will be fully reliable in the next release

1-gran:
> > > I can't see what this is fixing, could you provide an example URL that won't scrape with the current selector?
> > 
> > 
> > Any URL. Im not sure why, but your code will only fetch the image succesfully once. Then if you scrape another URL it wont work. If I restart the app then it will work again, but only once. Maybe its just on my end, and theres some weird caching going on, I dont know. I tried on Chrome and Firefox. My PR fixes the issue for me atleast.
> 
> You are quite right that this is a caching issue with the xpath library that Stash depends on: the PR with a fix [has been merged](https://github.com/stashapp/stash/pull/4126) so this selector (and many others like it) will be fully reliable in the next release

Oh, lol. Glad you've found the issue. I will just use my own code locally as a workaround until the next release then.

PR #1500, opened 2023-10-17
feat: ✨scraper DC OnlyFans (FansDB)

xantror:
This PR improves upon the existing `dc-onlyfans` scraper. I have left that one intact in case some people still prefer to use it.

Among other things, this scraper changes the following compared to the original:
- Uses `stashapp-tools (stashapi)` instead of `py_common`
- Content of the title to aid with FansDB guideline compliance (see README for details)
- Returns the `post_id` as studio code
- Sets the studio name to the format used by FansDB `<username> (OnlyFans)` with a parent studio of `OnlyFans (network)`
- Improves the lookup of performer name to also search for aliases.
- Improved the retrieval of performer images to provide multiple options for performer image; images are also cached to speed up bulk scraping (Scrape All)
- Ability to mark scenes and galleries with a tag for identification, as users might not want to submit those to FansDB


scruffynerf (approved):
Looks good to me.  At first I wondered why we left the original intact, but between the changes and the FansDB guidelines (which other might not want to follow), makes sense to start it fresh with new scraper and leave the old for now.

scruffynerf:
I'd still like @Maista6969 to review.

Maista6969 (approved):

PR #1506, opened 2023-10-19
add scraper for nvg network sites

reynn:
nvg is exclusively a paid site so if this doesn't belong I understand and we can just close it. I no longer have access to the site myself but this functioned as of about a month ago and they dont seem to update to often.

Not sure about the user-agent being a header here so if that seems wrong let me know and I can remove it, it does have to be in sync with the cookie though so I felt it was better to have here so if a user updates one seeing the other might remind them that they have to update both.


Maista6969:
Still valuable for members of NVG, happy to add this 👍

PR #1509, opened 2023-10-20
Fixed xpaths for Title, URL, and Image for IWantClips.yml

smcallah:
iwantclips.com appears to have changed xpaths that are scrapeable for Title, URL, and Image, so I have changed the xpaths in the scraper to accommodate.

Tested scraper with as many scenes as I could, the majority work fine, however some appear to only have an mp4 as the "image" for the scene and those cannot pull an image URL for stash.

### DIFF
--- a/scrapers/IWantClips.yml
+++ b/scrapers/IWantClips.yml
@@ -29,9 +29,9 @@ xPathScrapers:
       Image: //div[@class='clip-thumb-16-9']/a/img/@src
   sceneScraper:
     scene:
-      Title: //span[@class="headline hidden-xs"]/text()
+      Title: //h1[@class="headline hidden-xs"]/text()
       URL:
-        selector: //button[@class='btn btn-default cart addToCart']/@data-id
+        selector: //link[@rel="canonical"]/@href
         postProcess:
           - replace:
               - regex: ^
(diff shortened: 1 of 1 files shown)


scrapers/IWantClips.yml
```diff
@@ -29,9 +29,9 @@ xPathScrapers:
...
   sceneScraper:
     scene:
-      Title: //span[@class="headline hidden-xs"]/text()
+      Title: //h1[@class="headline hidden-xs"]/text()
       URL:
-        selector: //button[@class='btn btn-default cart addToCart']/@data-id
+        selector: //link[@rel="canonical"]/@href
         postProcess:
           - replace:
               - regex: ^
```
Maista6969, line 37: Since the link is absolute now we need to remove this postprocessing: tested with this URL, got this result
```
https://iwantclips.com/store/82932/Goddess-Blonde-Kitty/2945599/Had-a-Bad-Day-Watch-This
https://iwantclips.com/store/item/https://iwantclips.com/store/82932/Goddess-Blonde-Kitty/2945599/Had-a-Bad-Day-Watch-This
```

Maista6969 (commented):
Tested with a selection of clips from their front page and this seems to work exactly as you described 👍 

Will merge once you've addressed the comment on extra postprocessing

smcallah (commented):
Removed selector/postProcess from scene URL

Maista6969:
Thanks for fixing this! :)

PR #1511, opened 2023-10-21
Add StasyQVR XPath Scraper

meed69:
VR sister site of existing StasyQ scraper, but the VR site has a totally different layout so couldn't reuse the other scraper. Tested on a few videos older and newer. Performer page on VR site has zero details and didn't seem worth scraping.

Site is light on video details so scraper should be fairly straightforward, but feedback is welcomed.


Maista6969:
Works great, thank you for contributing 🙂

PR #1515, opened 2023-10-23
Update RealJamVR to include PornCornVR

meed69:
Brand new site (September 2023) made by the RealJamVR folks which seems to have an identical page layout. Updated the Studio name to no longer be fixed and added the new URL, but otherwise seems to work great without any other adjustments from the handful of scenes I tested.


Maista6969:
Excellent, thank you for this contribution! I love when we can remove fixed values and make scrapers more general :)

PR #1456, opened 2023-09-02
Add Algolia scraper for Falcon Studios

GayNerd:
Moves (some of) Falcon Studios from the Gamma scraper to the Algolia scraper

Tested with scenes from falconstudios, ragingstallion, and hothouse.

PR #1517, opened 2023-10-24
Remove FalconStudios from Gamma scraper

GayNerd:
This was supposed to have been done in #1456, but was missed.


Maista6969:
Thanks for catching my mistake! 😅

PR #1519, opened 2023-10-25
Add scraping for tags for Sayuncle.yml

GayNerd:
Add tag scraping for Sayuncle scenes.

IMO, they tag excessively, but someone may want (some of) them.


Maista6969:
Just like all of the Paper Street Media sites 😅 The good thing is that Stash will only add the tags you actually have automatically 👍

PR #1529, opened 2023-11-07
Fix TeenMegaWorld.yml scraper because of site changes, update SCRAPERS-LIST to add more TMW sites. Delete old tmwvrnet.com scraper

smcallah:
Fix TeenMegaWorld.yml scraper because of site changes, update SCRAPERS-LIST to add more TMW sites. Delete old tmwvrnet.com scraper


Maista6969:
Very nice, thank you for cleaning this up :)

PR #1530, opened 2023-11-07
Update TeenMegaWorld.yml

sierrax:
Updated scene scraper with new changes to the site, the XPath is probably not as elegant as it could be.


sierrax:
Closing as it's a subset of the other TMW pull request submitted recently

PR #1532, opened 2023-11-07
Update Clips4Sale.yml

zzfet:
Removed image link post-processing - previously, it was necessary to prepend 'https:' at the start of the link, however it seems that C4S have changed their site so this is no longer necessary.


scruffynerf (approved):
Looks good!

rudetuber:
Thank you! This had been really bothering me.

PR #1535, opened 2023-11-15
Fix for IfeelMyself lazy loading image tags

pops64:
This caused a None type to return. The video tag appears to be load right away and shares the same information need for the rest of the code to work correctly. Had to add a check to ensure there was a video tag. As some actress have galleries now and these are not lazy loaded and have no video tag.


pops64:
Made an additional improvement. The search function would return false positives in the cause of 4 and 48 both would match in the old method. As add a safety check in the case of direct url searches

PR #1536, opened 2023-11-18
Added Pure-XXX scraper (pure-xxx, pure-ts, pure-bbw)

l1xander:
Adds Pure-XXX scraper with support for following websites.
    - pure-xxx.com
    - pure-bbw.com
    - pure-ts.com

Existing Pure-BBW scraper is deleted to avoid conflict.


scrapers/Pure-XXX.yml
```diff
@@ -0,0 +1,40 @@
...
+                with: $1$3
+              - regex: \/\/(.*).com\/.*@~@(.*)
+                with: https://$1.com$2
+      Studio:
+        Name:
+          selector: //div[@id="logo"]//img/@alt | //div[@class="flexslider"]//a/@title
+          postProcess:
+            - replace:
+              - regex: (.*)\-(.*)
+              - with: $1 $2
```
Maista6969, line 32: The validator doesn't like that you've got an extra dash in front of `with:` here, and this is also causing scrapes from pure-xxx.com to fail because the studio ends up empty

Maista6969 (changes requested):
Looks good except for the typo in that one postprocessing action, happy to merge once you've fixed that 👍 

Thanks for fixing this scraper!

scrapers/Pure-XXX.yml
l1xander: Thank you for pointing it out. I have fixed the PR.

Maista6969 (approved):

PR #1543, opened 2023-11-27
Update IWantClips Scraper

Muldec:
Scrape Studio name instead of fixed "IWantClips"
Fix scraping of description when it has multiple paragraphs (ignored and squashed before, not anymore)


Maista6969:
This had been bugging me, thank you for fixing it 👍

PR #1550, opened 2023-12-02
Add EvilAngelVideo scraper

GroteKnuffelbeer:


Maista6969:
Tested with several scenes and movies and this works great! The description for scenes aren't really accurate, but they're the best we can hope for from a third-party storefront 👍

Thanks for contributing!

PR #1555, opened 2023-12-07
Added TitanMen.yml

Dark-Obsidian:
- Added TitanMen scene and movie scraper.

### DIFF
--- a/scrapers/TitanMen.yml
+++ b/scrapers/TitanMen.yml
@@ -1,0 +1,64 @@
+# yaml-language-server: $schema=../validator/scraper.schema.json
+
+# Author: KinksterGuy
+# Based on: MenAtPlay.yml
+# Last updated: 2023-12-07
+
+name: TitanMen
+sceneByURL:
+   - action: scrapeXPath
+     url:
+        - titanmen.com
+     scraper: sceneScraper
+movieByURL:
+   - action: scrapeXPath
+     url:
+        - titanmen.com
+     scraper: movieScraper
+xPathScrapers:
+   sceneScraper:
+      scene:
+         Title:
+            selector: //h1[@class="scene-header-title"]/text()
+         Details:
+            selector: //div[@class="col span_6_of_12"][1]/p/text()
+            concat: "\n\n"
+         Date:
+            selector: //div[@class="col span_6_of_12"][2]/p[1]/text()
+            postProcess:
+               - parseDate: Jan 2, 2006
+         Studio:
+            Name:
+               fixed: TitanMen
+         Performers:
+            Name:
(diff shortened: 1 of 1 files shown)


scrapers/TitanMen.yml
```diff
@@ -0,0 +1,64 @@
...
+            Name:
+               selector: //div[@class="col span_6_of_12"][2]/p[3]/a/text()
+         Movies:
+            Name:
+                selector: //div[@class="col span_6_of_12"][2]/p[2]/a/text()
+         Tags:
+            Name:
+               selector: //div[@class="col span_6_of_12"][2]/p[4]/a/text()
+         Image: //meta[@property="og:image"]/@content
+         # Doesn't work, this site does not have any og:image tags in the head
```
Maista6969, line 43: They've hidden away their images inside some JavaScript, but we can get our hands dirty
```suggestion
         Image:
            selector: //script[contains(text(), "scene-image-")]/text()
            postProcess:
               - replace:
                    - regex: .*image:\s*"([^"]+).*
                      with: $1
```

The resolution isn't great, but it's an official scene cover

scrapers/TitanMen.yml
```diff
@@ -0,0 +1,64 @@
...
+            selector: //div[@class="col span_9_of_12"]/p[3]/text()
+            postProcess:
+               - parseDate: Jan 2, 2006
+         Studio:
+            Name:
+               fixed: TitanMen
+         Performers:
+            Name:
+               selector: //div[@class="col span_9_of_12"]/p[4]/text()
+         FrontImage: //div[@class="header-container"]/img/@content
```
Maista6969, line 61: Try this instead, it's worked for every movie I've tested it on
```suggestion
         FrontImage: //div[contains(@class, "box-cover")]//img/@src
```

scrapers/TitanMen.yml
```diff
@@ -0,0 +1,64 @@
+# yaml-language-server: $schema=../validator/scraper.schema.json
+
+# Author: KinksterGuy
+# Based on: MenAtPlay.yml
+# Last updated: 2023-12-07
```
Maista6969, line 5: Not really necessary, your authorship will be reflected in the commit history 😁

We have started omitting the "Last updated" comment for the same reason: stashapp/stash#4242 will introduce a scraper management system that gets this information from the commit history as well

Maista6969 (changes requested):
Thank you for contributing! I hope some of my suggestions are helpful, do not hesitate to ask for clarification: I feel like I've given you a bunch of homework here, I hope it doesn't come across as negative 😅

Maista6969:
This has not moved in a month, but I think this is a great scraper so I'm adding it to the repo myself while of course giving you co-authorship credit

PR #1556, opened 2023-12-07
Added PerformerScraper to AdultEmpire

AshoPash:


scrapers/AdultEmpire.yml
```diff
@@ -102,4 +108,31 @@ xPathScrapers:
                 - regex: ^
                   with: "https://www.adultdvdempire.com"
       URL: //meta[@name='og:url']/@content
-# Last Updated October 09, 2023
+  performerScraper:
+    common:
+      $infoPiece: //*[@id="profileModal"]/div/div/div[2]/div[1]/ul
+    performer:
+      Name: //*[@id="content"]/section/div/div[2]/h1/text()
+      Birthdate: $infoPiecel/li[contains(text(), 'Born:')]/text()
```
Maista6969, line 116: There's a typo here (an extra l in `$infoPiecel`) and we should try to remove the prefix:
```suggestion
      Birthdate:
        selector: $infoPiece/li[contains(text(), 'Born:')]/text()
        postProcess:
          - replace:
              - regex: Born:\s+(.*)
                with: $1
```

scrapers/AdultEmpire.yml
```diff
@@ -102,4 +108,31 @@ xPathScrapers:
...
+      Gender:
+        fixed: "Female"
+      Height:
+        selector: $infoPiecel/li[contains(text(), 'Height:')]/text()
+        postProcess:
+          - replace:
+              - regex: ^(\d+).*
+                with: $1
```
Maista6969, line 124: Here's that same typo again, and this regex will never match: if the field contains `Height:` it will always be at the start of the line so we need to strip it out. We also convert height to metric because that's what Stash expects
```suggestion
      Height:
        selector: $infoPiece/li[contains(text(), 'Height:')]/text()
        postProcess:
          - replace:
              - regex: Height:\s+(.*)
                with: $1
          - feetToCm: true
```

scrapers/AdultEmpire.yml
```diff
@@ -102,4 +108,31 @@ xPathScrapers:
...
+      Image: //*[@id="content"]/section/div/div[1]/a[1]/img/@src
+      Country: $infoPiecel/li[contains(text(), 'From:')]/text()
+      Measurements: $infoPiecel/li[contains(text(), 'Measurements:')]/text()
```
Maista6969, line 127: Same typo, same procedure
```suggestion
      Measurements:
        selector: $infoPiece/li[contains(text(), 'Measurements:')]/text()
        postProcess:
          - replace:
              - regex: Measurements:\s+(\d\d\w*)\D+(\d+)\D+(\d+).*
                with: $1-$2-$3
```

scrapers/AdultEmpire.yml
```diff
@@ -102,4 +108,31 @@ xPathScrapers:
...
+      CareerLength: $infoPiece[text() = 'First Seen:']/../text()
```
Maista6969, line 128: I have not found any performers that had this field, but this might need the same treatment as the other fields: strip out the prefix

scrapers/AdultEmpire.yml
```diff
@@ -102,4 +108,31 @@ xPathScrapers:
...
+      Gender:
+        fixed: "Female"
+      Height:
+        selector: $infoPiecel/li[contains(text(), 'Height:')]/text()
+        postProcess:
+          - replace:
+              - regex: ^(\d+).*
+                with: $1
+      Image: //*[@id="content"]/section/div/div[1]/a[1]/img/@src
+      Country: $infoPiecel/li[contains(text(), 'From:')]/text()
```
Maista6969, line 126: Same typo, same procedure
```suggestion
      Country:
        selector: $infoPiecel/li[contains(text(), 'From:')]/text()
        postProcess:
          - replace:
              - regex: From:\s+(.*)
                with: $1
```

Maista6969: This still won't give the correct country, but that's a limitation of their website: they list this as "city, state" for American performers and "city, country" for everyone else

Still better than nothing 🙂

Maista6969 (changes requested):
Thank you for contributing! I have a few suggestions I'd like you to address before I can merge this

Maista6969:
Thanks again for this contribution 👍

PR #1558, opened 2023-12-12
Improve IWC cover scraping

Muldec:
Today IWC image scraper gets the image from a "poster" placeholder in the page. However, this placeholder can contain still images (jpg, png, ...), .gif, or .mp4 files resulting in the following behavior :
* still images => scraping OK
* gif => scrape an animated image that is against the guidelines
* mp4 => error while scraping (image doesn't show)

This PR updates the "poster" url to grab the cover that is displayed in the search results which results in :
* still images  => no change, scraping OK
* gif => scrape a jpg being the cover of the scene => OK
* mp4 => scraped a jpg sometimes, or gets an error (similar to previous behavior)

In essence, this PR improves the image scraping by avoiding GIF retrieval and some MP4 errors.


Maista6969:
I was wondering why some of the IWC submissions looked so bad, thanks a lot for fixing this!

PR #1559, opened 2023-12-13
Update MFC.yml to add gallery and performer scrapers

imagineimaginingthings:
Gallery scraper is a copy paste of the already working scene scraper. MFC shares can contain both videos and photos, but regardless of content the underlying HTML is identical.

sample URL: `https://share.myfreecams.com/a/pwghp8y6`

Performer scraper works on `profiles.myfreecams.com` URLs. It's self explanatory for the most part.

Image: gets their "main photo" if one is set, otherwise falls back to their avatar. Gets the full size image where possible.
Height/weight: MFC allows models to input values in either imperial or metric. But as was the issue with the PornHex scraper I wrote, since we cannot perform math in YAML files and/or handle cases for both imperial and metric I have chosen to only get height/weight data if metric values are found. I (or someone else) could update this scraper in the future to use python to handle both cases. I don't know if it's still like it now, but MFC always used to have significantly more non-US models than US, so I'd think metric is more common.

sample URLs:
`https://profiles.myfreecams.com/lea_amano` (imperial measures ignored)
`https://profiles.myfreecams.com/Sindyy` metric
`https://profiles.myfreecams.com/RocknRose` customized profile using CSS to hide all the fields but the underlying HTML is still there. No main photo is set so gets her avatar instead.

### DIFF
--- a/SCRAPERS-LIST.md
+++ b/SCRAPERS-LIST.md
@@ -1189,6 +1189,7 @@ private.com|Private.yml|:heavy_check_mark:|:x:|:heavy_check_mark:|:heavy_check_m
 privatecastings.com|privatecastings.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 privatesextapes.com|Wtfpass.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
 producersfun.com|ProducersFun.yml|:heavy_check_mark:|:x:|:x:|:x:|-|-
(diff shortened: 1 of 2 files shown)


Maista6969:
Great work again! Tested with several profiles and everything seems to work like described. In the next version of Stash we will have the ability to postprocess with Javascript (see stashapp/stash#4200) so we can revisit the imperial/metric height differences then 👍

PR #1561, opened 2023-12-13
Added search by name and some scenescraper fixes.

ryosaeba75:


scrapers/DorcelVision.yml
```diff
@@ -10,34 +10,65 @@ movieByURL:
...
+            - regex: (.*)
+              with: https://www.dorcelvision.com$1
+              
+      Image: 
+        selector: $card/a[@class="movies rolloverv2"]/img[last()]/@data-src
+        postProcess: 
+          - replace:
+              - regex: (.*)
+                with: https://www.dorcelvision.com$1
+      Details: $card/details/summary
```
Maista6969, line 42: Some of these have longer descriptions, try this:
```suggestion
      Details: $card/details//p[last()]
```

scrapers/DorcelVision.yml
```diff
@@ -10,34 +10,65 @@ movieByURL:
...
+
```
Maista6969, line 43: Let's grab the performers while we're at it 👍 
```suggestion
      Performers:
        Name:
          selector: $card/details//p[@class="persons"]
          postProcess:
            - replace:
                - regex: ^With:\s+
                  with:
                - regex: \s*,\s*
                  with: __SEPARATOR__
          split: __SEPARATOR__
```

Maista6969 (approved):
Looks great, I only have two suggestions but I can merge this right away and I can fix 'em if you'd like 👍

scrapers/DorcelVision.yml
ryosaeba75: I couldn't manage to make this work when I tried, it returns always the same list of performers for every result, even ones where is no performers listed.

Maista6969: I apologize, I thought I tested this but upon retrying it now I can see the exact issue you're describing: disregard this comment 😅

Issue #1566, opened 2023-12-17
Python Scrapers do not work folderized


ouenascender:
Python Scripts to not work with change to folderizing that has taken place. Using CopyToGalley as the scraper I had issues with.

Error Output:

```
23-12-17 19:12:55
Error   
scrapeSingleScene: input: scrapeSingleScene scraper CopyToGallery: could not unmarshal json from script output: EOF
2023-12-17 19:12:55
Error   
could not unmarshal json from script output: EOF
2023-12-17 19:12:55
Error   
[Scrape / Copy to Gallery] You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)
```

Cause:
Current Design of Scrapers Does not support reference using folder trees, change would need to be made to support folder trees or py_common would need to be referenced in the call the the plugin

twist3dimages:
same problem, works on some of my computers but can't figure out how to fix this

ouenascender:
Added Commit https://github.com/stashapp/CommunityScrapers/commit/85732bb6bfb03793b0e70f67f3f61942b0057e1f to Resolve MindGeekAPI also not working with same issue

Maista6969:
These scrapers did not end up in folders by accident: we're restructuring the repo in preparation for the changes in stashapp/stash#4242, and the fix will differ depending on whether or not stashapp/stash#4372 gets included: for now the fix is not moving them back to the base scraper directory, but to include this snippet at the top of the Python file:
```python
import os
import sys
# to import from a parent directory we need to add that directory to the system path
csd = os.path.dirname(os.path.realpath(__file__))  # get current script directory
parent = os.path.dirname(csd)  #  parent directory (should be the scrapers one)
sys.path.append(
    parent
)  # add parent dir to sys path so that we can import py_common from ther
```

PR #1567, opened 2023-12-17
Issue 1566

ouenascender:
Moving of CopyTo Scrapers back to root directory to resolve Issue #1566


ouenascender:
Added Commit 85732bb to Resolve MindGeekAPI also not working with same issue

Maista6969:
These scrapers did not end up in folders by accident: we're restructuring the repo in preparation for the changes in stashapp/stash#4242, and the fix will differ depending on whether or not stashapp/stash#4372 gets included: for now the fix is not moving them back to the base scraper directory, but to include this snippet at the top of the Python file:
```python
import os
import sys
# to import from a parent directory we need to add that directory to the system path
csd = os.path.dirname(os.path.realpath(__file__))  # get current script directory
parent = os.path.dirname(csd)  #  parent directory (should be the scrapers one)
sys.path.append(
    parent
)  # add parent dir to sys path so that we can import py_common from ther
```

PR #1569, opened 2023-12-17
Regroup all Pure Media sites into one scraper.

SpecialKeta:
Regrouping all [Pure Media Entertainment](http://www.puremediaenterprises.com), which are currently scattered across several scraper files, into one scraper.
Changed name from PureXXX to PureMedia, to reflect parent network studio.


Maista6969:
Thank you, I love it when we can consolidate scrapers! I tested this with scenes from all 6 sites, but this seems to fail on sissypov.com. I used [incredible tattooed sissy maid blows you away](https://sissypov.com/tour/trailers/incredible-tattooed-sissy-maid-blows-you-away.html) as a test case, and it looks like this particular site shares a layout with the final studio from the link you posted, pornstarbts.com

Since it's not unlikely that some of the other sites might end up migrating to what might be a newer layout, I would suggest adding a second scene scraper in the same file for this other layout so we can keep the whole network in the same file similar to what we've done in [ExploitedX.yml](https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/ExploitedX.yml#L3-L15)

Maista6969:
Excellent, thank you for your contribution! 👍

SpecialKeta:
👍🏻 
Fixed the scraper for SissyPOV, by adding it as a second scene scraper, like you mentioned.
Copied the working code from the old SissyPOV scraper, and tested it on a few scenes.
Added PornStarBTS, as it uses the same layout, and also tested on a few scenes.

PR #1572, opened 2023-12-19
Sexuria.net scraper was added

AshoPash:


scrapers/sexuria.yml
```diff
@@ -0,0 +1,34 @@
...
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //*[@id="dle-content"]/article/h1
+      Details:
+        selector: //h4[text()='Description']/following-sibling::div/text()
+        concat: " "
+      Date:
+        selector: //div[text()='Release Date:']/following-sibling::div/text()
```
Maista6969, line 20: Are you sure you wouldn't rather use `Publication Date` here? I'm not very familiar with Sexuria (it looks like a pirate site?) but every post I checked had a publication date and the ones with release dates were often partial, so they only included a year (which is not valid for local Stash)

Maista6969 (commented):
Thank you for contributing! I can merge this if my nitpick about the dates is incorrect, waiting to hear back from you (who is presumably more familiar with the site)

scrapers/sexuria.yml
AshoPash: I am not quite sure with the PublicationDate ( because it is always bigger than the ReleaseDate) so I thought maybe it is the date that the information is published in the site.

AshoPash: I didn't know how I could add the default values for missing parts ( like 01-01)

Maista6969: I see, you're entirely correct: the publication date has nothing to do with the scene/movie itself 😅

AshoPash:
@Maista6969  I would like to extend the scraper so that one can search the site. To do that I have to write a script to interact with the website to find the search element and send the search text via a form .Could you please tell  me if there is a sample source in the repo ? ( I know how to do it in C# but not in python )

Maista6969:
> @Maista6969 I would like to extend the scraper so that one can search the site. To do that I have to write a script to interact with the website to find the search element and send the search text via a form .Could you please tell me if there is a sample source in the repo ? ( I know how to do it in C# but not in python )

Most sites don't use POST requests for searches so I took a look at Sexuria: when you search it does indeed run a POST request to `/ajax.php`, but that in turn will simply run a GET request to the real search endpoint! Here's all you need to add to get search support in your XPath scraper:
```yml
sceneByName:
  action: scrapeXPath
  queryURL: https://sexuria.net/f/j.x.name-released;p.title={}/sort=date/order=desc/
  scraper: searchScraper
sceneByQueryFragment:
  action: scrapeXPath
  queryURL: "{url}"
  scraper: sceneScraper
```
And in the  `xPathScrapers` section (next to the existing `sceneScraper`):
```yml
  searchScraper:
    common:
      $result: //article[contains(@class, "short")]
    scene:
      Title: $result//h2
      URL: ($result//a)[1]/@href
      Details: $result//div[@class="author-cat"]
      Date: $result//div[@class="date"]
      Image:
        selector: $result//img/@data-src
        postProcess:
          - replace:
              - regex: ^
                with: https://sexuria.net
```

If you'd still like to give Python scraping a shot, you could look at how [The Score Group](https://github.com/stashapp/CommunityScrapers/blob/b8430242a378be820d388b82d44b81e95c6c7b9c/scrapers/TheScoreGroup/TheScoreGroup.py) and [WAPdB](https://github.com/stashapp/CommunityScrapers/blob/b8430242a378be820d388b82d44b81e95c6c7b9c/scrapers/WAPdB/WAPdB.py) scrapers do it: both of these are for sites that use POST requests to handle searches

There are other Python scrapers that implement search like this, you can search the repo for calls to `requests.post` to find some of them 🙂

PR #1573, opened 2023-12-20
AniDB.yml: scrape by episode URL, add movie scraper, improve regex

symptom6186:
This pull request adds several new features to the AniDB scraper. 

## Scrape by episode URL

The url scene scraper has been separated into two scrapers:
- One to scrape by anime URL ( e.g., https://anidb.net/anime/230)
- One to scrape by episode URL (e.g., https://anidb.net/episode/3262)

The previous implementation of the scraper used only the anime URL. However, as a series usually contains multiple episodes, this made scene titles ambiguous. The new episode scraper sets the title, the air date, and the anime URL.

Consequently, the anime URL scraper no longer sets the title. (This can be re-enabled, however, by un-commenting the corresponding line.)

The anime URL scraper should still be used to set the performers, tags, artwork, etc. 

## Add movie scraper

This scraper now effectively uses the movie page to group all episodes in the same anime series.  

To this end, the anime URL scraper now suggests a movie entry when scraping a scene. In addition to scraping the movie title (i.e., the anime series name), the movie URL is also set. 

On the movie page, the anime URL can be used to scrape the movie artwork, title, alias, date, etc.

## Improvements to scraping scene by fragment

Previously, the by-fragment scene scraper would only work iwhen the filename was a substring of the series title. For example, "Heavenly Del" would match with [Heavenly Delusion](https://anidb.net/anime/?adb.search=heavenly%20del), but "Heavenly Del - 01" [would not](https://anidb.net/anime/?adb.search=heavenly%20del%20-%2001).

The regex matching has been rewritten to strip superfluous strings in the filename and replace characters that Anidb dislikes. All of the following strings will now be be rewritten to the correct title.

```
My Lewd Anime.mp4                          ->  My Lewd Anime
[XX] My Lewd Anime (720p).mp4              ->  My Lewd Anime
(XX) My Lewd Anime [720p].mp4              ->  My Lewd Anime
[XX] My Lewd Anime - 01 (720p) (x264).mp4  ->  My Lewd Anime
[XX]_My_Lewd_Anime-01_(720p)_(x264).mp4    ->  My Lewd Anime
```

Note that this does not catch ever use case (e.g., `My Lewd Anime - s01e02.mp4` becomes `My lewd Anime - se`), but this matching schema works the majority of the time.

Also note that the tagger blacklist may interfere with these regex expressions, as the tagger blacklist appears to have priority (for instance, if the blacklist strips brackets, then the brackets will be deleted before a match to "[XX]" can be attempted).

## Testing the new changes

If you want to test all of these features, I would recommend doing the following.

1. `touch '[XX] heavenly delusion - 01.mp4'`, move to library
2. On the scene page, scrape with AniDB (i.e., using the by-fragment scraper)
3. In addition to setting the performers, tags, etc, this should create a new movie entry, so navigate to the newly create movie page to scrape the movie page by URL
4. On the scene page, scrape the episode URL (https://anidb.net/episode/264164) for the title and date.

### DIFF
--- a/scrapers/AniDB.yml
+++ b/scrapers/AniDB.yml
@@ -1,31 +1,68 @@
 name: AniDB
-################################################################################################################
-#                                                HOW TO SET UP                                                 #
-#                              Store this file in the ~/stash/scrapers/AniDB.yml                               #
-#                       (If the scrapers directory is not there it needs to be created)                        #
-#                                                 SET COOKIES:                                                 #
-#             Access the anidb.net website > login > right button > inspect > find cookies storage             #
-#       Copy the "Value" of "adbsess" and "adbuin" and replace in the cookies category of this document        #
-#    If your account is new, you need to access any NSFW anime and confirm that you want to see 18+ content    #
-#                    Do not change the order of the columns, as it can make it stop working                    #
-#                               I recommend creating a new account just for this                               #
-#                                               SET USER AGENT:                                                #
-#               Go to your Stash > Settings > Metadata Providers > Scraping > Scraper User Agent               #
-#                                      Use the User Agent of your choice                                       #
-#     I'm currently using: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0      #
-################################################################################################################
-#                                                  HOW TO USE                                                  #
-#                                                   SCENES:                                                    #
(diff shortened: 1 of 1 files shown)


Maista6969:
Thank you for this excellent pull request: I really appreciate the time you took to document this scraper and the reasoning behind it (I would have asked about the unset Title field for sure!)

PR #1582, opened 2023-12-30
Add support for loading markers for Vixen Media Group content

existing-paradox:
There are also some additional error handling features that may or may not be wanted

### DIFF
--- a/scrapers/vixenNetwork/vixenNetwork.py
+++ b/scrapers/vixenNetwork/vixenNetwork.py
@@ -50,6 +50,8 @@ LOCAL_PATH = r""
 SERVER_IP = "http://localhost:9999"
 # API key (Settings > Configuration > Authentication)
 STASH_API = ""
+
+# Automatically reattempt GraphQL queries to Vixen sites which fail with a 403 response 
 MAX_403_REATTEMPTS = 20
 
 SERVER_URL = SERVER_IP + "/graphql"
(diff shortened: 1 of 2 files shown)


existing-paradox:
I am working on updating to support match v0.24.x

existing-paradox:
Can confirm this now works with updated schema

existing-paradox:
I am fairly certain that is a limitation of the Stash GraphQL configuration, I am not seeing a query supporting ignore case. To handle this, I watched the warning logs and added aliases where necessary then rerunning the scrape. It's not perfect but doesn't require too much extra effort.

existing-paradox:
The tag alias method has been tested and works. The name of the marker will be named what ever the the chapter title is set as from Vixen API, not the name of the primary tag defined in marker.

This functionality is based on the functionality used by the MindGeekAPI.

existing-paradox:
I have also seen chapter titles misspelled. You can't always programmatically solve for every edge case 😁

Maista6969:
But we can certainly match tags without considering their case: would this work?
```python
def graphql_findTagbyName(name):
    query = """
        query {
            allTags {
                id
                name
                aliases
            }
        }
    """
    result = callGraphQL(query)
    normalized_name = name.lower().strip()
    for tag in result["allTags"]:
        if tag["name"].lower() == normalized_name:
            return tag["id"]
        if tag.get("aliases"):
            for alias in tag["aliases"]:
                if alias.lower() == normalized_name:
                    return tag["id"]
    return None
```

existing-paradox:
@Maista6969 good call! Not sure why I didn't think of that... Update coming soon

PR #1584, opened 2023-12-31
Fix faphouse tags scraping

existing-paradox:


existing-paradox:
I am working on updating to support match v0.24.x

existing-paradox:
This required no changes for updated schema

PR #1586, opened 2023-12-31
Refactor PMVHaven scraper

dopepump:
The old method using cloudscraper was heavy handed IMO, I have made these changes:

- New method takes video ID directly from input URL because that is the only valid video site URL format. Removed unnecessary imports
- Code deduplication (fail method)
- More helpful error logging
- Changed script layout to allow for possible future scrapeBy options
- More helpful var names
- Reverse search thumbnails as I found a case where a blank thumb was returned as it was a incompatible type for the video site, and this change will ensure the most recent thumb is pulled into stash
- Updated yml file to specify sceneByURL

### DIFF
--- a/scrapers/PMVHaven/PMVHaven.py
+++ b/scrapers/PMVHaven/PMVHaven.py
@@ -35,7 +35,8 @@ def getData(sceneId):
     return req.json()
 
 def getIMG(video):
-    for item in video['thumbnails']:
+    # reversed as we want most recent thumb
+    for item in reversed(video['thumbnails']):
         if item.startswith("https://storage.pmvhaven.com/"):
             return item
     return ""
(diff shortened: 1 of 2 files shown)


JackDawson94:
Love the improvements 👍

Not sure why you removed the SOCKS5 bugfix though
We should have it in all Python scrapers due to the incompatibility between Golang and Python implementations of Socks5(h)
But that is just my opinion, not sure it's a "priority" for the community, but if the code is there and it works 😁

dopepump:
> Love the improvements 👍
> 
> Not sure why you removed the SOCKS5 bugfix though We should have it in all Python scrapers due to the incompatibility between Golang and Python implementations of Socks5(h) But that is just my opinion, not sure it's a "priority" for the community, but if the code is there and it works 😁

ty! For the proxy fix, I haven't tested it using a proxy, I removed it as I was copying the layout of a newly updated scrapers and thought it was redundant. If you have a means of testing feel free to do so and commit it back in.

Maista6969:
I love it when we can remove dependencies! I feel like fixing the proxy issue should be done at a higher level, but who knows if that will ever get done 😅

PR #1591, opened 2024-01-03
Add support for Worst Behavior Productions to Assume the Position scraper

DuctTape42:
Assume The Position Studios (and a handful of other studios) use the same API.  This adds support for Worst Behavior Productions to the Assume The Position Studios scraper.

I also wanted a scrapeByFragment that can use the filename (which is basically studiocode_studio_bitrate) to find the correct url and scrape it.  This adds a python scraper to do the parsing and then it turns around and does a scrapeSceneURL to call the yml scraper.

### DIFF
--- a/scrapers/AssumeThePositionStudios.py
+++ b/scrapers/AssumeThePositionStudios.py
@@ -16,11 +16,7 @@ except ModuleNotFoundError:
 
 
 def debugPrint(t):
-    #sys.stderr.write(t + "\n");
-
-def call_graphql(query, variables=None):
-    return graphql.callGraphQL(query, variables)
-
+    log.debug(t + "\n");
 
 def scrape_scene(url):
     query = """
@@ -75,7 +71,7 @@ query scrapeSceneURL($url: String!) {
 }"""
 
     variables = {"url": url}
-    result = call_graphql(query, variables)
+    result = graphql.callGraphQL(query, variables)
     log.debug(f"result {result}")
     res = result["scrapeSceneURL"]
     res["url"] = url
@@ -99,7 +95,7 @@ if m := re.match(r"(\d+)_[a-z]+_[a-z]+_\d+_([A-Z]+).*", title):
         apiUrl = "https://assumethepositionstudios.com/api/site/13/updates/0"
         sceneUrlFragment = "https://assumethepositionstudios.com/trailer?updateId="
     else:
-        print("{ Unknown site code }" + siteCode)
+        debugPrint("Unknown site code " + siteCode)
         sys.exit();
 
     debugPrint("Searching for " + str(contentId) + " on " + apiUrl)
--- a/scrapers/AssumeThePositionStudios.py
+++ b/scrapers/AssumeThePositionStudios.py
@@ -1,0 +1,123 @@
+import json
+import sys
+import requests
+import re
+import os
+
+try:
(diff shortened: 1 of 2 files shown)


scrapers/AssumeThePositionStudios.yml
```diff
@@ -35,9 +47,9 @@ jsonScrapers:
           selector: data.scene.description
           postProcess:
             - replace:
-                - regex: (.*?)([A-Z,'\s]+)\.?$
+                - regex: (.*?)([A-Z,/'\s]+)\.?$
```
Maista6969, line 50: I think this fix also needs to be applied on line 35, the regex for description: if I'm reading it right that's supposed to filter out the trailing list of performers and tags that are all in uppercase, but some tags contain forward slashes so it fails for [this scene](https://worstbehaviorproductions.com/trailer?updateId=12016) for example

scrapers/AssumeThePositionStudios.py
```diff
@@ -0,0 +1,123 @@
...
+except ModuleNotFoundError:
+    print(
+        "You need to download the folder 'py_common' from the community repo! (CommunityScrapers/tree/master/scrapers/py_common)",
+        file=sys.stderr,
+    )
+    sys.exit()
+
+
+def debugPrint(t):
+    #sys.stderr.write(t + "\n");
```
Maista6969, line 19: This could just delegate to `log.debug`, it's very helpful when scrapers actually log something so we can find errors faster 🙂

You could also just replace all calls of `debugPrint` with `log.debug`, and adding a trailing newline is not necessary since the log module already takes care of flushing the output

scrapers/AssumeThePositionStudios.py
```diff
@@ -0,0 +1,123 @@
...
+
+def call_graphql(query, variables=None):
+    return graphql.callGraphQL(query, variables)
```
Maista6969, line 22: Same as above, no need for a wrapper function here 😅

scrapers/AssumeThePositionStudios.py
Maista6969: This needs to be fixed either way because it's a syntax error and this script won't run at all in its current state 😅

scrapers/AssumeThePositionStudios.py
```diff
@@ -0,0 +1,123 @@
...
+    siteCode = m.group(2)
+    debugPrint("Site code is " + siteCode)
+    if siteCode.casefold() == "WBP".casefold():
+        apiUrl = "https://www.worstbehaviorproductions.com/api/site/21/updates/0"
+        sceneUrlFragment = "https://worstbehaviorproductions.com/trailer?updateId="
+    elif siteCode.casefold() == "ATP".casefold():
+        apiUrl = "https://assumethepositionstudios.com/api/site/13/updates/0"
+        sceneUrlFragment = "https://assumethepositionstudios.com/trailer?updateId="
+    else:
+        print("{ Unknown site code }" + siteCode)
```
Maista6969, line 102: Use `log.warning` or `log.error`, never `print` in a Stash scraper script 🙂

Maista6969 (changes requested):

Maista6969 (approved):

Maista6969:
Thank you for this contribution! 🙂

PR #1592, opened 2024-01-03
Bugfix

JackDawson94:
- Bugfix due to a change in the PMVHaven code, found a simpler way to get the info 🥳
- Cleaned up some unused code

### DIFF
--- a/scrapers/PMVHaven/PMVHaven.py
+++ b/scrapers/PMVHaven/PMVHaven.py
@@ -2,22 +2,12 @@ import os
 import json
 import sys
 import requests
-import random
-import time
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


dopepump:
Beat you to it :P 
https://github.com/stashapp/CommunityScrapers/pull/1586

Have a look through mine as I added some more robust early fails and relevant logging. Also noticed that sometimes it would pull bad thumbnails since the site doesn't support PNGs right now. So if you upload a fixed thumbnail the scraper needs to fetch the most recent thumb, i.e. the last one in the array.

JackDawson94:
@dopepump  thanks, I'll take a look at your code, I'll also close this MR

PR #1593, opened 2024-01-03
PornWorld scapper was not pulling just about anything correctly.

TheStashMaster:
Updated scraper to pull data correctly now.  Commented out Studio Code as it was configured to use the scene title which I didn't think was correct but have no idea what to add there now.

### DIFF
--- a/scrapers/pornworld.yml
+++ b/scrapers/pornworld.yml
@@ -32,26 +32,29 @@ xPathScrapers:
           - replace:
               - regex: featuring.+
                 with: ""
-      Code:
-        selector: //h1
+#      Code:
+#        selector: //h1
+#        postProcess:
+#          - replace:
+#              - regex: .*(GP\d+).*
+#                with: $1
+      Date:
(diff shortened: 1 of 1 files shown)


SpecialKeta:
The scraper works fine for urls that are not starting with tour10.pornworld.com.
Scene urls starting with tour10.pornworld.com don't have a studiocode. The usual urls that start with www.pornworld.com have GP as studiocode. 

Maybe combine both in 1 scraper file? Like they've done in ExploitedX scraper?

Maista6969:
You had inadvertently removed some fields from the old PornWorld scraper when you split them: I have tested this with scenes from both [the "regular" site](https://pornworld.com/watch/34312/college_slut_tiffany_tatum_friends_get_caught_during_intense_classroom_orgy_gp160) and the [tour10 version of the site](https://tour10.pornworld.com/watch/695272-27190/redhead-babysitter-kendra-cole-seduces-hung-boss-into-ass-sex) and it works for both

PR #1594, opened 2024-01-05
Add HobbyPorn

IAmKontrast:
New scraper for hobby.porn

It's my contribution. Would be happy about some feedback about my work.

### DIFF
--- a/scrapers/HobbyPorn.yml
+++ b/scrapers/HobbyPorn.yml
@@ -1,0 +1,39 @@
+name: HobbyPorn
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - hobby.porn
+    scraper: sceneScraper
+sceneByQueryFragment:
+  action: scrapeXPath
+  queryURL: "{url}"
+  scraper: sceneScraper
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://hobby.porn/search/{}/
+  scraper: sceneSearch
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="title-block"]/h1/text()
+      Performers:
+        Name: //span[@itemprop="author"]/a/text()
+      Studio:
+        Name:
+          fixed: Pornhub
+      Date:
+        selector: //div[@class="video-data"]/span/text()[contains(.,'/')]
+        postProcess:
+          - parseDate: 01/02/2006
+      Tags:
+        Name: //a[@itemprop="genre"]/text()
+      URL: //link[@itemprop="url"]/@href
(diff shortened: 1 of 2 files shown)


scrapers/HobbyPorn.yml
```diff
@@ -0,0 +1,39 @@
...
+  scraper: sceneSearch
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[@class="title-block"]/h1/text()
+      Performers:
+        Name: //span[@itemprop="author"]/a/text()
+      Studio:
+        Name:
+          fixed: Pornhub
```
Maista6969, line 23: This doesn't seem quite right, but the since the site lists the studio (or "author") along with the performers we might need to hack around that

For this I used a [German Scout scene](https://hobby.porn/video/slim-dreadlocks-girl-nicky-pickup-and-public-fuck-at-street-casting/) (which has a studio and a performer) and a [Morgpie scene]() (where Morgpie is both the "studio" and the performer)

```suggestion
      Performers:
        Name: //span[@itemprop="author" or @itemprop="actor"]/a
      Studio:
        Name: //span[@itemprop="author"]/a
```

This will unfortunately lead to the studio being listed as a performer, but as far as I can tell there's no way to avoid that given the site layout

Maista6969 (changes requested):
Looks good to me! I will note that you do not need to use `/text()` at the end of most of these selectors: if you just select a node, Stash will automatically grab its text content

The only time we need `/text()` is when the node contains other children that we do not wish to include

Will be happy to merge once you review my comment about the studio name

IAmKontrast:
Thank you @Maista6969 for the quick and helpful response and comment!

I've found another url direct to the Pornhub video. Some (or maybe a lot?) of the videos might be deleted but I think it would be a good idea to add this url too.

`//div[@class="player"]//a/@href`

Do you know a way to add two separate urls with xPathScrapers?

Maista6969:
I agree, that is a good idea! Unfortunately the Stash code for XPath scrapers has not been updated to account for multiple URLs so the only way to scrape that would be to convert this into a script scraper and use something like Python 😅

IAmKontrast:
> I agree, that is a good idea! Unfortunately the Stash code for XPath scrapers has not been updated to account for multiple URLs so the only way to scrape that would be to convert this into a script scraper and use something like Python 😅

What a pity. Maybe in the future. Recoding in Python sounds like overkill for a lot of dead links. If the scene is recent enough I would prefer the Pornhub scraper anyway. HobbyPorn is a great source for deleted/older ones.

PR #1596, opened 2024-01-06
Fix private.com Movie Title

ferengi82:
Fix the broken Movie Title on Private.com
I'm not an XPath expert, so there might be better ways to do this.


Maista6969:
Looks good to me, thanks for fixing this 🙂

PR #1599, opened 2024-01-10
Add DMM.co.jp scraper

justwjx:
this is scraper designed to scraper metadata from DMM.co.jp in JAPANESE language with official release IMAGE

Please make sure scrape through a JAPAN based VPN, visit from IP outside of JAPAN may being blocked by this website.

Here is my thoughts regarding metadata language: 
since quality of all translated text from JAV japanese title or description may vary due to the tool or person, and AI driven tool are more intelligent, to keep soruce language text maybe is a better choice, so you can translate to any language properly at any time moving forward.

### DIFF
--- a/scrapers/DMM.yml
+++ b/scrapers/DMM.yml
@@ -1,0 +1,115 @@
+name: DMM
+sceneByFragment:
+  action: scrapeXPath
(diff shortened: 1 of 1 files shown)


Maista6969:
Tested this with several movies and it works great! Thank you for this contribution :)

PR #1603, opened 2024-01-11
Add several JAV scraper, all from official websites

justwjx:
All filenames of the scrape scripts containing '-JP' are associated with the Japanese version of metadata. Meanwhile, it's possible that these websites also offer an English version. The scrape scripts for the English version may or may not have already been created.

### DIFF
--- a/scrapers/arzon.jp.yml
+++ b/scrapers/arzon.jp.yml
@@ -1,0 +1,130 @@
+name: arzon.jp
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.arzon.jp/itemlist.html?t=&m=all&s=&q={filename}
+  queryURLReplace:
+    filename:
+      - regex: \..+$
+        with: ""
+  scraper: sceneQueryScraper
+sceneByURL:
+  - action: scrapeXPath
+    url:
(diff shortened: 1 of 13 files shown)


scrapers/arzon.jp.yml
```diff
@@ -0,0 +1,130 @@
...
+  cookies:
+    - CookieURL: "https://www.arzon.jp"
+      Cookies:
+        - Name: "PHPSESSID"
+          Domain: "www.arzon.jp"
+          Value: "l7srbei96iti9ffs1eftqn99q5"
+          Path: "/"
+driver:
+  useCDP: true
+  sleep: 1
```
DogmaDragon, line 127: Validator returns an error, should be:
```
driver:
  cookies:
    - CookieURL: "https://www.arzon.jp"
      Cookies:
        - Name: "PHPSESSID"
          Domain: "www.arzon.jp"
          Value: "l7srbei96iti9ffs1eftqn99q5"
          Path: "/"
  useCDP: true
  sleep: 1
```

DogmaDragon:
Some general comments:
- Last Updated lines doesn't use full month. 
- Contains debug/driver even when set to false.

PR #1608, opened 2024-01-14
Title and description fixes for Helix

evandeaubl:
This applies a couple of fixes to the Helix scraper needed due to recent site changes:

* The HTML title for scene pages is now what a regular person looking at the scene page would call the scene subtitle. This changes the scraper to pull the title that appears in the largest font on the scene page, which has historically been the scene title (which, prior to this site change, matched the HTML title).
* Helix added performer links to the scene description a while back, and the XPath query didn't account for them, omitting the text and adding spurious line breaks. This changes the scraper to use the entire contents of all non-empty `<p>` blocks.

### DIFF
--- a/scrapers/HelixStudios.yml
+++ b/scrapers/HelixStudios.yml
@@ -48,17 +48,13 @@ xPathScrapers:
       $info: //div[@class="text"]
     scene:
       Title:
-        selector: //title/text()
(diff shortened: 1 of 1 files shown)


Maista6969:
Great fix, thanks for contributing!

PR #1613, opened 2024-01-16
Updated scene scraper for Data18 to drop ' #1' from movie title series

TheStashMaster:
If a movie is a part of a movie series the first title doesn't have a number appended to the title but in the scene scrapper it is added.  I updated the scraper to drop that trailing #1.

### DIFF
--- a/scrapers/data18.yml
+++ b/scrapers/data18.yml
@@ -52,7 +52,12 @@ xPathScrapers:
         Name: $studio
         URL: $studio/@href
       Movies:
-        Name: $movie/text()
+        Name:
+          selector: $movie/text()
+          postProcess:
+            - replace:
+                - regex: '\s#1$'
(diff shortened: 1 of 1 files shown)


Maista6969:
Seems like a personal preference, what's to say other users don't want their movies labeled "Movie #1", "Movie #2", etc.? I think we're better off just staying true to the source here

TheStashMaster:
If you compare the first movies of a series from data 18 to other sites like adult empire you'll see data 18 adds something like vol. 1 to the movie title. Example:

Fuck My MILF Pussy Vol. 1
https://www.data18.com/movies/151134-fuck-my-milf-pussy

Fuck My MILF Pussy
https://www.adultempire.com/1380740/fuck-my-milf-pussy-porn-movies.html

PR #1615, opened 2024-01-18
Fix HotMovies Scraper

GenericMale:
This fixes the HotMovies scraper.

Note that AdultEmpire provides the same data as HotMovies with a different UI.
The AdultEmpire scraper currently searches for dvds and this HotMovies scraper searches for clips which i feel is a good split.

### DIFF
--- a/scrapers/Hotmovies.yml
+++ b/scrapers/Hotmovies.yml
@@ -57,9 +57,13 @@ xPathScrapers:
       URL: //meta[@name="og:url"]/@content
       Director: //a[@label="Director"][1]
       Duration:
-        selector: //meta[@property="og:video:duration"]/@content
+        selector: //strong[contains(text(), "Run Time")]/following-sibling::text()[1]
         postProcess:
-          - javascript: return new Date(value * 1000).toISOString().slice(11, 19)
+          - replace:
+              - regex: " hrs. "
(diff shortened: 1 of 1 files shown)


GenericMale:
@Maista6969 sorry, i fixed the validation now by getting rid of the javascript seconds to hh:mm:ss conversion

Maista6969:
> @Maista6969 sorry, i fixed the validation now by getting rid of the javascript seconds to hh:mm:ss conversion

Hehe, and I just fixed the validator so you could keep it 🙂 The `javascript` postprocessor is new enough that you're the first one to run into this validation error! You can revert your most recent commit and this should still pass and we can get it merged 👍

GenericMale:
> > @Maista6969 sorry, i fixed the validation now by getting rid of the javascript seconds to hh:mm:ss conversion
> 
> Hehe, and I just fixed the validator so you could keep it 🙂 The `javascript` postprocessor is new enough that you're the first one to run into this validation error! You can revert your most recent commit and this should still pass and we can get it merged 👍

it's fine this way though. end result is the same. not having some wonky javascript in there feels like a good idea if it can be avoided 😅

Maista6969:
Good point, there's too much javascript in the world as is 😁

PR #1625, opened 2024-01-21
Update NewSensationsNetworkSites.yml

Ronnie711:
Added FamilyXXX in as I hadn't realised it was still a separate scraper!


Maista6969:
You should delete `FamilyXXX.yml` while you're at it, otherwise they'll conflict and Stash will just use the first one it finds 🙂

Ronnie711:
> You should delete `FamilyXXX.yml` while you're at it, otherwise they'll conflict and Stash will just use the first one it finds 🙂

Also removed HotwifeXXX as I did that last week but it still existed seperately! :)

Maista6969:
Beautiful, I love it when we can consolidate another network scraper!

PR #1631, opened 2024-01-25
Update KBProductions scraper to include LucidFlix

MortonBridges:
LucidFlix is a new site under the KB Productions umbrella.  It has some unique characteristics, but with some small modifications the existing KBProductions scraper works just fine with it.

Modified SCRAPERS-LIST.md for the new URL
Modified KBProductions.yml file to add LucidFlix URLs for scenes and models
Modified KBProductions.py to account for unique characteristics of LucidFlix


Maista6969:
Thank you for this, I'm glad to see we could reuse the scraper for this network 👍

PR #1636, opened 2024-01-27
Gallery to image Gallery scraper

dogwithakeyboard:
This is a copy of "Copy to Gallery" intended to push Gallery metadata down to the attached Images. I've made an attempt at making it configurable; I couldn't find an example of a config.ini in use so hopefully this is implemented correctly!

### DIFF
--- a/scrapers/CopyToImages/CopyToImages.py
+++ b/scrapers/CopyToImages/CopyToImages.py
@@ -1,0 +1,151 @@
+import json
+import sys
+import os
+
+# to import from a parent directory we need to add that directory to the system path
+csd = os.path.dirname(os.path.realpath(__file__))  # get current script directory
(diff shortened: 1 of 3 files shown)


Maista6969:
Tested this and it works great, but I have a question (mostly because I don't collect galleries or images)

Would it make sense to use `ADD` instead of `SET` as a safer default for `tagsmode`, `performersmode` and `urlsmode`? In case the user already tagged some of their images the default setting would overwrite their existing tags/performers/URLs

I'll defer to you on this one, happy to merge this as is 🙂

dogwithakeyboard:
It's a good suggestion, I think you're right. tbh the defaults are _my_ settings that I'm inflicting on everyone else.

Maista6969:
Gotta be SOMEONE's settings, might as well be yours 😁

Thanks for the safer defaults, though 👍

PR #1652, opened 2024-02-16
Update MetroHD.py

Ronnie711:
Removing space in hookups as there is no mention anywhere on their website about the space!

### DIFF
--- a/scrapers/MetroHD/MetroHD.py
+++ b/scrapers/MetroHD/MetroHD.py
@@ -21,7 +21,7 @@ def metrohd(obj: Any, _) -> Any:
     match dig(obj, "studio", "name"):
         case "Deviant Hardcore":
             replacement = "devianthardcore.com"
-        case "Family Hook Ups":
+        case "Family Hookups":
             replacement = "familyhookups.com"
         case "Girl Grind":
             replacement = "girlgrind.com"


Maista6969:
You can see that they use the name "Family Hook Ups" internally if you check the [site directory on familyhookups.com](https://www.familyhookups.com/porn-sitemap/scenes)

This part of the code is using the name returned from the API to determine the correct domain name for URLs, it doesn't change how the studio name comes out

Try this instead:
```python
studio_map = {
    "Family Hook Ups": "Family Hookups",
    "Metro": "Metro HD",
}


def metrohd(obj: Any, _) -> Any:
    replacement = None
    match dig(obj, "studio", "name"):
        case "Deviant Hardcore":
            replacement = "devianthardcore.com"
        case "Family Hook Ups":
            replacement = "familyhookups.com"
        case "Girl Grind":
            replacement = "girlgrind.com"
        case "Kinky Spa":
            replacement = "kinkyspa.com"
        case "She Will Cheat":
            replacement = "shewillcheat.com"
        case _:
            replacement = "metrohd.com"

    # Replace the studio name in all URLs: even if there's no specific studio,
    # metro.com is wrong and needs to be replaced with metrohd.com
    fixed = replace_all(obj, "url", lambda x: x.replace("metro.com", replacement))

    fixed = replace_all(fixed, "name", replacement=lambda x: studio_map.get(x, x))
    return fixed
```

Maista6969:
Fixed this in 64a5a6664e8286681768414aa2ba588eb460b05e

PR #1657, opened 2024-02-19
Update rb configs

mmenanno:
Starting to update this, looks like nothing merged relyings on this common folder yet, so this only affects my future projects.

### DIFF
--- a/scrapers/rb_common/configs/README.md
+++ b/scrapers/rb_common/configs/README.md
@@ -2,29 +2,35 @@
 
 ## Adding your settings to an existing config
 
-Various ruby scrapers will use these configs so that each one doesn't separately need to ask for your api keys and endpoints. For example if you want to configure your stash instance you can open the `stash_config.rb` with any text editor and in the `USER_CONFIG`, add your details. For example you might change:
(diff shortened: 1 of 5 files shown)


Maista6969:
While we're waiting for stashapp/stash#4433 you should be aware that any scrapers that get updated/reinstalled by Stash will overwrite any files that the user has modified, which might include configuration 😅

This is usually just an annoyance: most script scrapers don't see that many updates... but for "packages" like `py_common` and `rb_common` it can be more of a headache since users don't expect an unrelated scraper install to affect their "main config" that they've set in these libraries 🙂

The workaround I've implemented in [py_common/config.py](https://github.com/stashapp/CommunityScrapers/blob/4a92a4fde2dac5af22af7765d02e8d1c2b55ad25/scrapers/py_common/config.py) is a helper function called `get_config` that will persist a `config.ini` file in the same directory as whatever file calls it, which in turn is used both internally in [py_common/graphql.py](https://github.com/stashapp/CommunityScrapers/blob/4a92a4fde2dac5af22af7765d02e8d1c2b55ad25/scrapers/py_common/graphql.py#L8-L16) and externally in scrapers like [GEVI.py](https://github.com/stashapp/CommunityScrapers/blob/4a92a4fde2dac5af22af7765d02e8d1c2b55ad25/scrapers/GEVI/GEVI.py#L15-L26)

It's not elegant, but hopefully we'll get proper scraper configuration in Stash itself before too long

mmenanno:
Ah thanks for the heads up!

feederbox826:
https://github.com/stashapp/stash/issues/6017

PR #1664, opened 2024-02-22
Adult empire single sceene rather than entire movie

TheStashMaster:
The current solution only scrapes for the movie as a whole. I've added an additional scraper to scrape a clip from the movie as well.  I've found the image on the page is in an iFrame so I tried to do a subscraper.  From the debug I can see it is finding the correct URL but no image is scraped out, not sure if you can provide any insight there.


TheStashMaster:
I've never done a subScraper before and it's still not working but I'm not able to work out why. Any advice or point me to where I can get a little education would be great.

PR #1672, opened 2024-02-27
fix: adjust Arx scraper to updated graphql API

nrg101:
# problem

The GraphQL API has changed slightly in one of the query field names, and the response structure is one level less for errors.

# fix

This fixes those issues by using the changed field name `legacyScene` instead of `scene`, and using the first level `errors` instead of second level.

# extra

A general GraphQL introspection query variable `GRAPHQL_INTROSPECTION` is added that can be used again in future to discover API fields, queries, etc. This was used temporarily this time to find the changed field name.

Studio code scraping is now added, using the scene's `id` field.

### DIFF
--- a/scrapers/Arx/Arx.py
+++ b/scrapers/Arx/Arx.py
@@ -281,6 +281,7 @@ def get_scene(url):
     ret['image'] = result.get('primaryPhotoUrl')
     ret['date'] = result.get('availableAt') and result.get('availableAt')[:10] \
         or result.get('createdAt') and result.get('createdAt')[:10]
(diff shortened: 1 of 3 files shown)


nrg101:
tested successfully:

- https://honeytrans.com/scenes/2589/lingerie-trans-latina-seduces-her-roommate-for-anal
- https://transroommates.com/scenes/2743/redhead-trans-rachael-belle-gets-her-dress-back

Maista6969:
Thanks for the fix and for that GQL introspection query, that looks like it will come in handy for sure :)

PR #1675, opened 2024-02-28
Add Kim Holland scraper

cambiph:


Maista6969:
Thank you for contributing this! I will see if I can add the `kimholland.nl` scraper from #1671 to this so that users can scrape both sites 🙂

PR #1680, opened 2024-03-01
Update MYLF to use API

ltgorman:
This update will now use the API for MYLF.  This is essentially the same as the TeamSkeetAPI pointing at the MYLF API endpoint and some other minor mods for MYLF.


Maista6969:
Oh yikes, I thought I merged this already! Thanks for the contribution 🙂

PR #1686, opened 2024-03-11
Adding TMDB for new schema

Stash-KennyG:
Unsure why/when we lost the TMDB scraper.  Very useful for movies.  This has been updated to new schema on the site and pre-scrape the artwork into the created movie object.

### DIFF
--- a/scrapers/TMDB.yml
+++ b/scrapers/TMDB.yml
@@ -1,0 +1,98 @@
+name: TheMovieDB
+
+movieByURL:
+  - action: scrapeXPath
+    url:
+      - themoviedb.org
+    scraper: movieScraper
+sceneByURL:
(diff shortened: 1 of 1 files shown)


Maista6969:
We've never had a TMDB scraper 🙂

Based on the comments and structure of this it looks like an edit of the AdultEmpire scraper, perhaps you've converted it to a TMDB so long ago that you assumed it was in CommunityScrapers? 😁

Anyway I've tested this and it gives me a 403 Forbidden unless I configure a User-Agent string in `Settings > Metadata Providers > Scraping > Scraper User Agent`: once I set that it's all good! I suggest we add it to the scraper itself:
```yml
driver:
  headers:
    - Key: "User-Agent"
      Value: stash-scraper/1.0.0
```

PR #1691, opened 2024-03-14
New scraper using Algolia for 3rd Degree Films

TheStashMaster:
I do not have a gallery to test with so I've left that bit commented out.


Maista6969:
Tested and this works great, even for galleries so I'll enable those too :)

PR #1695, opened 2024-03-16
Update BrokenLatinaWhores scraper

213788:
- Simplify some selectors
- Add performer scraper


scrapers/BrokenLatinaWhores/BrokenLatinaWhores.yml
```diff
@@ -0,0 +1,41 @@
+name: Broken Latina Whores
```
Maista6969, line 1: Should add the requirement here
```suggestion
# requires: py_common
name: Broken Latina Whores
```

scrapers/BrokenLatinaWhores/BrokenLatinaWhores.py
```diff
@@ -0,0 +1,70 @@
...
+    scraped = get_scraped(inp)
+    if not scraped:
+        return {}
+
+    tree = html.fromstring(scraped.content)
+    image = tree.xpath('//div[contains(@class, "model_picture")]/img/@src0_3x')[0].strip()
+    image = '{uri.scheme}://{uri.netloc}/{img}'.format(uri=urlparse(scraped.url), img=image[1:])
+    name = tree.xpath('//meta[@name="keywords"]/@content')[0].strip().capitalize()
+    birthdate = re.search("([0-9]{2})", "".join(tree.xpath('//div[@class="model_bio"]/text()'))).group(0)
+    birthdate = datetime.now().replace(year=datetime.now().year - int(birthdate)).replace(month=1, day=1).strftime(
```
Maista6969, line 51: This is only approximately correct for the models who have performed _this year_, not sure the false positives are worth it for a public scraper

Maista6969 (changes requested):
Looks good but needs to put the `py_common` requirement into the yml file 👍

scrapers/BrokenLatinaWhores/BrokenLatinaWhores.py
213788: The point was to get approximate age of performer. I agree it is not best, but I think it's good enough because we don't have any other data sources (?). After the date is obtained/calculated it could be rather removed before saving, if needed.

Maista6969: I believe that unlike sites like AnalVids that calculate the age dynamically based on an actual birth year, this site has ages that were input when the profile was created and are never updated: essentially their debut age

If we look at the profile for [Beba](https://brokenlatinawhores.com/models/beba.html) for example, it lists her as 18, 19 and 20 years old :)

We'll just have to watch out for this in the queue

Maista6969 (approved):

scrapers/BrokenLatinaWhores/BrokenLatinaWhores.py
213788: That's new, when I was writing that it looked like the age value was dynamic, will try to submit some change if I manage to come with anything. Thanks for heads up on that!

PR #1697, opened 2024-03-17
(Tested) Fix for SQLite timestamp converter deprecation

twist3dimages:


twist3dimages:
Lots of the databases have different timestamp (from previous versions of DC and now OF-DL so I got a little lazy ..

twist3dimages:
image

twist3dimages:
If the created at is supposed to be from the database not the file let me know

PR #1700, opened 2024-03-20
Fixed issue with regex where scraping would not work for video with brackets in name.

mrmedieval:
I also added URLs to the scraper.


Maista6969:
Thank you for contributing! 🙂

PR #1721, opened 2024-03-29
Update PegasProductions.yml

MortonBridges:
Added User-Agent header to prevent 403-Forbidden response.

Added scraper for second format that studio has for scenes.  It's less useful than their primary format, but adding this in the event that a scene is only available via the newer, less useful, format.

Changed studio name to match what StashDB uses for the studio.

Added Director field to the scraper for the primary format.

Added French support to the scraper for the primary format should someone prefer to scrape this site in French rather than English.

### DIFF
--- a/scrapers/PegasProductions.yml
+++ b/scrapers/PegasProductions.yml
@@ -8,7 +8,7 @@ sceneByURL:
   - action: scrapeXPath
     url:
       - pegasproductions.com/abonnements
-    scraper: abonnementsnScraper
+    scraper: abonnementsScraper
 
   - action: scrapeXPath
     url:
@@ -16,7 +16,7 @@ sceneByURL:
     scraper: sceneScraper
 
 xPathScrapers:
-  abonnementScraper:
+  abonnementsScraper:
     scene:
       Title: //meta[@itemprop="name"]/@content
       Date:
--- a/scrapers/PegasProductions.yml
+++ b/scrapers/PegasProductions.yml
@@ -58,29 +58,29 @@ xPathScrapers:
         postProcess:
           - replace:
               - regex: janvier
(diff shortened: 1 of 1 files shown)


scrapers/PegasProductions.yml
```diff
@@ -1,21 +1,90 @@
...
+                with: 09
+              - regex: octobre
+                with: 10
+              - regex: novembre
+                with: 11
+              - regex: d[eé]cembre
+                with: 12
+              - regex: \s
+                with: /
           - parseDate: 02/01/2006
```
Maista6969, line 86: This fails on the first 9 months of the year because yaml is a wack-ass format: `03` will be turned into just `3` unless you put it in quotes :)

```
DEBU[2024-03-29 14:25:40] Replace: 'mars' with '3'
DEBU[2024-03-29 14:25:40] Before: 26 mars 2024
DEBU[2024-03-29 14:25:40] After: 26 3 2024
<snip>
DEBU[2024-03-29 14:25:40] Before: 26 3 2024
DEBU[2024-03-29 14:25:40] After: 26/3/2024
Error parsing date string '26/3/2024' using format '02/01/2006': parsing time "26/3/2024" as "02/01/2006": cannot parse "3/2024" as "01"
```

I don't know which editor you prefer to use, but I can recommend VS Code with the [YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) from RedHat: it'll warn you about these common pitfalls PLUS it uses the schema file from the repository so you get Stash scraper-specific type hints and autocompletions 🙂

Maista6969 (changes requested):

Maista6969 (approved):

PR #1730, opened 2024-04-02
update PornDudeCasting to pull new and moved values

feederbox826:
- support `/models/` path, since /casting/CODE redirects
- fix:
  - Performer name
  - Tags
 
Fixed date parsing to use `article:published_time` which does match with all other metdata, rather than the rounded `Cast: MONTH YYYY`

- Requests 4K preview images from the server, 8K 404s but 1080p is still available


feederbox826:
FWIW, validation passed locally

PR #1747, opened 2024-04-08
Update TheScoreGroup.yml

Ronnie711:
Adding comment for reasoning behind scene studio selector as only 3 people currently know the reasoning ....


Maista6969:
I bet you'd reveal the Wu-Tang Secret if we told you that, too!!

Ronnie711:
> I bet you'd reveal the Wu-Tang Secret if we told you that, too!! 😉

Only if it saved me work in the edit queue 🤣

PR #1757, opened 2024-04-11
Add Pornmuschimovie network sites to MMP Network scraper

MortonBridges:
New sites in the Pornmuschimovie network that are (mostly) compatible with MMP_Network scraper:

      - dirty-chubby.com
      - dirtycasting.com
      - lustyxxxfamily.com
      - pornmuschimovie.com
      - sexgamesprivate.com
      - xxxrealityfuck.com

Date and Image selectors needed modification to handle formats used by new sites, but the rest of the selectors appear to work as a charm.

No performer information appears to be available in the new sites.

Some scenes pages are rather sloppy, with descriptions being labeled as Tags, so you may see some strange tags being suggested.


Maista6969:
I think these can be added to `MMP_Network.yml`, I can look into it tomorrow if you don't get around to it first 🙂

MortonBridges:
> I think these can be added to `MMP_Network.yml`, I can look into it tomorrow if you don't get around to it first 🙂

Looks like everything works in MMP, except for the image (at least on my test scene).  I'll look into it further.

PR #1759, opened 2024-04-14
Update x confessions image selectors

MortonBridges:
The previous scraper had issues with some scenes, where it would suggest an image that's not related to the scene being scraped.  This change modifies the image selector to attempt several methods to find a suitable picture:

- First, look in the script for a cover_title_picture value.   
- If that's not present, look for two consecutive images with UIDs in the URL.  The first is the movie poster, the second is the scene image. 
- If that's not present, look for a time stamp followed closely by two images.  Again the first is the movie version, the second is the image. 
- If we still don't have a match, look for a confession_cover_image value.  These are pretty good, but won't have the scene name. 
- Finally, if nothing matches, try the last image in the script.  It's often the right one.

Yeah...I know...weird.  But this reliably snags a suitable image for the scene, though often it will be the vertical movie poster instead of a horizontal scene image.  A solution to improve this may yet present itself in the future.

Other changes: 

Modified movie Front Image selector to grab the vertical image more suitable for movie poster images.

Passing along the URL when creating movies when scenes are scraped.

Replaced " for ' for strings, for consistency.


Maista6969:
Impressive! Thanks for including good explanatory comments and a solid commit message

I don't think we'll be able to replace this with anything less than a script scraper if it gets more complicated 😁

PR #1763, opened 2024-04-16
Add scrapers for AVMoo and AVSox

MortonBridges:
For avmoo.online as mentioned in [Issue 73](https://github.com/stashapp/CommunityScrapers/issues/73).  Included avsox.click, as that site uses the same layout.

New URL scene scraper for avmoo.online and avsox.click.
New URL movie and performer scrapers for avmoo.online.

avsox.click shares the same format as avmoo.online, except it doesn't appear to have longer length clips that would justify creating movies, nor does it appear that performer details are available at avsox.click (at least for those sampled when writing this scraper).

Anyone with more familiarity with the studios present on these sites are welcome to create studio maps for those with Japanese names, if such is desired.

Tested with the English versions of the URLs.


Maista6969:
This was an XML file instead of a YML file for some reason, but I've fixed it in master 🙂

PR #1765, opened 2024-04-17
New scraper for twisted factory

MortonBridges:
For twistedfactory.com, as requested in [Issue 73](https://github.com/stashapp/CommunityScrapers/issues/73).

New scraper for TwistedFactory, based on a combination of the Kink.yml and KinkMen.yml scrapers, with extensive additions to the performer scraper.

Added entry for twistedfactory.com, using TwistedFactory.yml.

This is a site under the Kink.com umbrella, and shares the same layout as kink.com and kinkmen.com.  This scraper was built first using kink.yml as the base, but pulling in the simpler performer and studio selectors from kinkmen.yml.  There is likely an opportunity to merge all three scrapers into one large scraper for the entire family of sites.  While that may be possible, for the benefit of those waiting on a scraper for this site, it was decided to build a dedicated one.  Primarily as each site / scraper have their own Scene and Performer search scrapers, which I didn't want to affect.

I'm trusting that the search options work properly for TwistedFactory.


MortonBridges:
Draft:  Found that kink.com has different tags for some performers in the script where I'm snagging the gender.  Also, kink.com has trans models as well, with the appropriate gender tags.  I will be expanding the twistedfactory gender selector to account for these, should the studio have scenes in the future with trans performers.

PR #1788, opened 2024-04-26
Add scraper for shopmaker sites

MortonBridges:
Added YML URL Scene scraper for sites that operate using Shopmaker.

Added 108 sites to scrapers list.  Scrapers list was resorted, which resulted in some previous entries being reordered as well.

Blackboundfox scraper deleted.  The site it served, blackboundfox.com, is defunct.  It appears that this site has resurfaced as blackboundfox.net and is a Shopmaker site.

One duplicate entry, mugfucked.com, was deleted from scrapers list.


MortonBridges:
108 sites.  Sorry.  It's about half of what some sources say is available in the Shopmaker program, but these are the ones I can find that are still alive and offer video content.

MortonBridges:
An additional 39 URLs added that are part of the Shopmaker program.

PR #1799, opened 2024-04-29
Add url scene scraper for Pornbcn (pornbcn.com)

MortonBridges:
Add Pornbcn (pornbcn.com) URL Scene scraper.

Add pornbcn.com to scrapers list using Pornbcn.yml.

### DIFF
--- a/scrapers/Pornbcn.yml
+++ b/scrapers/Pornbcn.yml
@@ -1,0 +1,34 @@
+name: PornBCN
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - pornbcn.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //img[@data-name='image']/@alt
+      Date:
+        selector: //div[@data-date_format]/input[@type='hidden']/@value
+        postProcess:
+          - parseDate: 20060102
+      Image:
+        selector: //div[contains(@style,'background-image')]/@style
+        postProcess:
+          - replace:
+            - regex: ^.+url\(([^\)]+).+
+              with: $1
+      Studio:
+        Name: //meta[@property='og:site_name']/@content
+      Tags:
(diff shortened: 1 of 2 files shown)


scrapers/Pornbcn.yml
```diff
@@ -0,0 +1,34 @@
+name: PornBCN
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - pornbcn.com
+    scraper: sceneScraper
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //img[@data-name='image']/@alt
```
Maista6969, line 10: Found that this is often the Spanish title even if you're scraping an English link, suggest we replace this with a different selector to capture what the user sees
```suggestion
      Title: (//div[contains(@class, "rowdesktop1nsd")]//div[contains(@class, "et_pb_text_inner")])[1]
```

scrapers/Pornbcn.yml
```diff
@@ -0,0 +1,34 @@
...
+          - replace:
+            - regex: ^.+url\(([^\)]+).+
+              with: $1
+      Studio:
+        Name: //meta[@property='og:site_name']/@content
+      Tags:
+        Name: //meta[@property='article:tag']/@content
+      Performers:
+        Name: //span[@class='namepornstarvi']//a
+      Details: //div[contains(@class,'toggle_content')]/p
```
Maista6969, line 27: They sometimes use `<br>` tags to add newlines to their descriptions (see [My First Client](https://pornbcn.com/en/mi-primera-clienta/)) and this wil unfortunately just mash those lines together

I found one scene that uses multiple `<p>` elements as well, so we can account for that as well

```suggestion
      Details:
        selector: //h5[contains(@class, "toggle_title")]/following-sibling::div[not(.//form)]//text()
        concat: "\n"
```

Maista6969 (changes requested):

PR #1821, opened 2024-05-08
Scraper for black girl fetish network

MortonBridges:
URL Scene scraper for blackgirlfetish.com, a network that includes:
Black Girls Fetish
Black Femdom Goddess
Bondage Fantasy Camp
Diamond Orgega
Fresh Flesh Amateurs

There's some excellent info for performers, but unfortunately their layout is a complete mess with no set format, so no performer scraper at this time (if ever...)


MortonBridges:
The site is a bit of a mess, unfortunately.  I've run across a few different formats for scenes and tried to account for them, but there may yet be more that I haven't encountered.  There may be more work yet before all scenes can be reliably scraped for all the available data.

PR #1839, opened 2024-05-14
Update IWantClips.yml

zzfet:
Fix title scraping for IWantClips after site redesign.


Maista6969:
Thanks for fixing this! I could have sworn I added it in 4519a71 but I must have forgot 😅

Issue #1843, opened 2024-05-16
X-Art scraper issues


FlashSpazzbo:
**Scraper name**
X-Art

**Scraper method**
Scene, scrape with URL

**Sample URL**
https://www.x-art.com/videos/crazy_for_stacy/

## Issue 1
Scraper fails without `www` sub-domain in URL.

### Description
<details>
Scraper should change URLs that don't contain `www` sub-domain to include sub-domain.

<br>For example, `https://x-art.com/videos/crazy_for_stacy/`, should be changed to `https://www.x-art.com/videos/crazy_for_stacy/
`

Website is an inconsistent mess and browsers will sometimes not provide the `www` URL on copy-paste actions.
</details>

## Issue 2
Scraper can't scrape members' site URLs

### Description
<details>
<p>Scraper should probably modify members' site URLs to public-facing URLs.</p>

The members' site URL for the sample URL above is `https://www.x-art.com/members/videos/crazy_for_stacy/
`. Dropping the `members` sub-folder will direct to the public-facing page.

<p>The public-facing pages also have relatively cleaner HTML & CSS. The members' scene pages also don't have the scenes' dates.</p>
</details>

## Issue 3
Date field not being scraped properly.

### Description
<details>
Scraper appears to be selecting first child `h2` for the date field, instead of the unclassed `h2` element the date information resides in.

<br>Website appears to use `h2` elements for visual formatting, instead of for semantic purposes.  In sample URL, there are seven unclassed ID-less `h2` elements in the main content section, the first four of which contain content for the scene description.

The date field, based on the last five or so scenes, appears to be in front of a `h2 + span`, such as this from the sample URL:
```html
<h2>May 09, 2024 </h2>
<h2>
    <span>featuring</span>
    <a href="https://www.x-art.com/models/stacy">Stacy</a>
</h2>
``` 

<p><strong>Note that the anchor tag is not always included.</strong></p>
</details>

## Issue 4
Header elements not being scraped as scene description text.

### Description

<details><summary>Details</summary>
In the sample URL above, anything not a paragraph element is being ignored. The full description text is:

```html
<h1><strong>WE ARE BACK.</strong></h1>
<h2>This 4K VIDEO is our second little teaser - letting you know that we are once again posting new content on our site... and quite a teaser it is!&nbsp;</h2>
<p>Model Stacy Cruz shows off every little bit of her slender form in this delightful video. She loves to show off for the camera - and for YOU!</p>
[...]

PR #1859, opened 2024-05-24
Update X-artcom.yml

MortonBridges:
Addressing points 2, 3, 4, and 5 from [Issue 1843](https://github.com/stashapp/CommunityScrapers/issues/1843)

Modified URL scene scraper so it reliably picks up date and gets all description text, preserving paragraphs.

Also, it now accepts /members/videos URLs, but maps them to /videos.

Simplified some selectors.

### DIFF
--- a/scrapers/X-artcom.yml
+++ b/scrapers/X-artcom.yml
@@ -3,7 +3,13 @@ sceneByURL:
   - action: scrapeXPath
     url:
       - x-art.com/videos/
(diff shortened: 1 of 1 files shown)


MortonBridges:
The Details selector is also snagging the Date.  Some clever person might be able to find the magic required to exclude this from the details, but that person is not currently me.

feederbox826:
LGTM ;)

PR #1863, opened 2024-05-26
Added ExCoGigirls site as scene URL in ExplotedX Scraper

pretty3muss6mortuary:
Noticed it wasn't available as a URL to scrape, confirmed it uses the same base template and can be scraped.
Tested and works as expected.


pretty3muss6mortuary:
I'm blind, missed the URl in the next code block section - deleting PR.

PR #1880, opened 2024-05-30
Update BlurredMedia.yml

Morfeel:
this update fixes BlurredMedia scraper with 
1. age restricition
2. using CDP
3. xpaths


feederbox826:
CDP isn't needed for age restriction bypass, this should be possible with just the cookie: `accessAgeRestriction=1`

PR #1881, opened 2024-05-30
Add multi domain support for MyDirtyHobby

IAmKontrast:
Added the German domain ".DE" to the url and also extended the cookie driver.

Can you (@bnkai and @Maista6969) review my code, please? I don't know if it's the correct way of doing this.

Also could you tell my about the "parseDate" function? `- parseDate: 02 Jan 2006` delivers the right date but in the edit dialog it's still in the MyDirtyHobby format and there was no post processing.

### DIFF
--- a/scrapers/MyDirtyHobby.yml
+++ b/scrapers/MyDirtyHobby.yml
@@ -3,6 +3,7 @@ sceneByURL:
   - action: scrapeXPath
     url:
       - mydirtyhobby.com/profil/
+      - mydirtyhobby.de/profil/
     scraper: sceneScraper
 xPathScrapers:
   sceneScraper:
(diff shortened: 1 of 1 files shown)


scrapers/MyDirtyHobby.yml
```diff
@@ -98,4 +99,11 @@ driver:
           Domain: ".mydirtyhobby.com"
           Value: "1"
           Path: "/"
-# Last Updated October 02, 2023
+  cookies:
```
Maista6969, line 102: This line can be removed: all cookies should be defined in the same `cookies` map :)
```suggestion
```

Maista6969:
The `parseDate` function is very precise in the date formatting it expects: it is currently `02 Jan 2006`, which works for the English version of the site, but the German version lists dates as `02 Jan. 2006` (with a period indicating the month is abbreviated)

This can be fixed by adding an extra `parseDate` pattern below the current one, so that the scraper will try both:
```yml
          - parseDate: 02 Jan 2006
          - parseDate: 02 Jan. 2006
```

Maista6969:
Superseded by 60ff569

PR #1889, opened 2024-06-03
Update GirlsOutWest.yml for latest site layout

MortonBridges:
Girls Out West has changed their layout, meaning a rewrite of the scraper is needed.

Scene and Performer URL scrapers have been modified to scrape the new layout.

It appears that their galleries are now behind a paywall, so it is not possible to scrape those.  I'd be happy to be proven wrong and return gallery scraping if a public URL can be provided.

### DIFF
--- a/scrapers/GirlsOutWest.yml
+++ b/scrapers/GirlsOutWest.yml
@@ -24,7 +24,8 @@ xPathScrapers:
         selector: //meta[@property='og:title']/@content
       Details: &details
         selector: //div[@class='description']//p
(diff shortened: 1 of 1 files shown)


MortonBridges:
Pulling this back so I can return gallery scraping now that I have a functional URL.

MortonBridges:
Galleries added back, as is the missing "details" section that I somehow overlooked.  But in my defense the original scraper didn't have those so I'll use that as a lame excuse for missing it.

PR #1896, opened 2024-06-06
bibian-av.yml: Fix javascript variable declaration

well-thats-funny:
Is Stash 0.26 this doesn't work anymore. Apparently, JavaScript engine is more strict in this version.

### DIFF
--- a/scrapers/bibian-av.yml
+++ b/scrapers/bibian-av.yml
@@ -39,7 +39,7 @@ xPathScrapers:
         postProcess:
           - javascript: |
               if (value && value.length) {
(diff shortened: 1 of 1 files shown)


Maista6969:
Thanks for fixing this! Looks like a consequence of stashapp/stash#4631 which replaced the old otto runtime with goja :)

well-thats-funny:
Oh, so it's a completely different engine now? Thanks, it may explain why some other scrapers I'm currently working on started to behave bit differently (e.g. console.log stopped working).

PR #1899, opened 2024-06-07
Fix image mapping for MYLF.com, add discrete image URL mapping for all TeamskeetAPI studios.

MortonBridges:
MYLF.com has changed their high resolution scene image URLs.  Image mapping routine added to get the best possible scene image resolution for TeamSkeet, SayUncle and MYLF URLs.

- SayUncle - use the default image from the scrape (it's high res to begin with) 
- TeamSkeet - use the original mapping that was in the previous version of the scraper 
- MYLF - Use new "bio_big" mapping that is new and unique to MYLF.


MortonBridges:
Thanks to @Ronnie711 for reporting the issue and finding the URL pattern for high res MYLF scene images.

PR #1910, opened 2024-06-18
add docker validator

feederbox826:
supercedes #1114 

This has the advantage of not needing a container built and it removes itself afterwards

Explanation:  
`-v` mounts the current directory (CommunityScrapers) to the docker container at `/app`
`--rm` removes the container once it exits  
`/bin/sh -c` takes us out of node-cli into a shell  
`cd /app/validator` moves us to the validator directory to install the dependencies  
`yarn install --silent` installs the dependencies needed for the validator  
`cd ..` takes us back out of the validator directory into `/app`
`node validate.js --ci` runs the validator

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,5 +102,5 @@ Specific scrapers can be checked using: `node validate.js scrapers/foo.yml scrap
 Instead of NodeJS being installed, Docker can be used to run the validator
 
 ```bash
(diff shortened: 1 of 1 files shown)


feederbox826:
#1114 can be closed :)

PR #1917, opened 2024-06-23
Add movie url to RagingStallion.yml

gimmeliina:


Maista6969:
Unclear on why you closed this but I tested it and it works fine: is there any reason I shouldn't add it to the scraper?

gimmeliina:
Deleted branch accidentally, so this closed automatically. Reopened.

PR #1936, opened 2024-07-14
Adding thepovgod.com to the Nubiles network scraper

tj2222:
This is a pretty new site (first scene is March 5, 2024) and is one of the standalone sites of the Nubiles network. It doesn't have galleries/photos, so I'm enabling the scene URL scraper for this site.


feederbox826 (approved):
LGTM. Might be eligible for other subscrapers, but time will tell

PR #1937, opened 2024-07-18
Get correct Oldje scene cover

jersprin:
Oldje used to have 2 types of scene cover, the montage and the poster, the poster was large, portrait and made up of multiple pictures, whereas the montage is normal size and landscape all in one jpg. 99% of Oldje scenes on StashDB use the montage card format. Oldje have now dropped the poster and gone for a screen cap, but the montage is still used. This change will automatically pick up the montage in keeping with StashDB covers. Montage cover is same format with movie- removed from URL

Example
Screen cap : https://www.oldje.com/sets/898/movie-morning-exercise.jpg 
Montage : https://www.oldje.com/sets/898/morning-exercise.jpg


Maista6969:
Thank you for fixing this (and for including the motivation!)

PR #1948, opened 2024-07-23
Create EuroPornstar.yml

L3GW0RK:
New performer scraper, testing and working for the past few months, thought I would share here

### DIFF
--- a/scrapers/EuroPornstar/EuroPornstar.yml
+++ b/scrapers/EuroPornstar/EuroPornstar.yml
@@ -1,4 +1,4 @@
-name: 06_europornstar
+name: EuroPornstar
 performerByName:
   action: scrapeXPath
   queryURL: https://www.europornstar.com/search.php?q={}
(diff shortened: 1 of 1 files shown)


scrapers/EuroPornstar/EuroPornstar.yml
```diff
@@ -0,0 +1,181 @@
+name: 06_europornstar
```
feederbox826, line 1: Do you mind chacning the name to reflect `EuroPornstar` ?

feederbox826:
other than trailing spaces and name discrepancy, LGTM

scrapers/EuroPornstar/EuroPornstar.yml
L3GW0RK: done, yeah my bad, that was the testing name I was using before copying it here, fixed it just now (I am used to codeberg and gitea so the process here is a bit new to me).

PR #1965, opened 2024-08-07
Rule34Video: Fix video description sometimes not getting read properly

Chr0nicallyT1red:
Fixed an issue where the video description would miss line breaks and would read wrong elements when the description is empty.  Would've wanted to insert correct line break ammount for mirroring the correct formating off of the site (e.g. multiple line breaks directly after one-and-other), but my skills are too limited for that


Maista6969:
Thanks for fixing this!

We've run into the issue of trying to recreate multiple linebreaks in scraped descriptions before and AFAICT it's currently not possible due to the way that Stash scraper internals strip the whitespace from node text so this is probably as good as we can get without rewriting as a script scraper 😅

PR #1985, opened 2024-08-21
Fetch studio code from Virtual Real Porn sites

Emilo2:
There's a convenient `<link rel="shortlink">` element that contains the studio code.


feederbox826:
A bit aggressive on the regex, but LGTM

PR #2010, opened 2024-09-05
Fix Private date parsing

SpecialKeta:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

[scene url](https://www.private.com/scene/marie-berger-and-catherine-knight-get-horny-in-an-anal-threesome/26724)

## Short description

When scraping scene urls, scraper returns date as ``2023-12-29T00:00:00+01:00``
Small fix so it returns ``2023-12-29``


feederbox826:
Tested on all available languages. LGTM

PR #2013, opened 2024-09-07
Add AlexLegend Scraper

WellDoneLiz:
Adds a sceneByURL xPathScraper for AlexLegend.com

## Scraper type(s)

- [x] sceneByURL

## Examples to test

[Recent Scene](https://alexlegend.com/video/skylar-snow-black-lingerie-surprise-sex-behind-the-scenes-1074.html)
[Popular Scene](https://alexlegend.com/video/bg-caitlin-bell-xmas-roommates-740.html)
[Old Scene](https://alexlegend.com/video/alex-grey-first-ir-dp-gangbang-288.html)

## Short description

Break out alexlegend.com from VNAGirls scraper. New site design that does not conform to VNA template.

### DIFF
--- a/scrapers/AlexLegend/AlexLegend.yml
+++ b/scrapers/AlexLegend/AlexLegend.yml
@@ -25,7 +25,7 @@ xPathScrapers:
         selector: //script[contains(text(),"myFluidPlayer")]
         postProcess:
           - replace:
-            - regex: ^.*posterImage\s?:\s?'(.*?)'.*
+            - regex: .+posterImage\s?:\s?'(.*?)'.+
               with: $1
       Studio:
         Name:
@@ -36,15 +36,13 @@ xPathScrapers:
           split: ", "
           postProcess:
             - replace:
-              - regex: ^Tags:(.+)
-                with: $1
+              - regex: 'Tags:'
       Performers:
         Name:
           selector: //div[contains(concat(" ", normalize-space(@class), " "), " main__models ")]
           split: ","
           postProcess:
             - replace:
-              - regex: ^Models:(.+)
-                with: $1
+              - regex: 'Models:'
 
 # Last Updated September 06, 2024
--- a/scrapers/AlexLegend/AlexLegend.yml
+++ b/scrapers/AlexLegend/AlexLegend.yml
@@ -8,21 +8,23 @@ sceneByURL:
 
 xPathScrapers:
   sceneScraper:
+    common:
+      $sceneWrapper: //section[contains(concat(" ", normalize-space(@class), " "), " content-sec ")]
     scene:
-      Title: //div[contains(concat(" ", normalize-space(@class), " "), " content-title ")]/h1
+      Title: $sceneWrapper//div[contains(concat(" ", normalize-space(@class), " "), " content-title ")]/h1
       Code:
-        selector: //link[@rel="canonical"]/@href
+        selector: /link[@rel="canonical"]/@href
         postProcess:
           - replace:
             - regex: .+-(\d+)\..+
               with: $1
-      Details: //div[contains(concat(" ", normalize-space(@class), " "), " content-desc ")][2]
+      Details: $sceneWrapper//div[contains(concat(" ", normalize-space(@class), " "), " content-desc ")][2]
       Date:
-        selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-date ")]/div[contains(concat(" ", normalize-space(@class), " "), " label ")]
+        selector: $sceneWrapper//div[contains(concat(" ", normalize-space(@class), " "), " content-date ")]/div[contains(concat(" ", normalize-space(@class), " "), " label ")]
         postProcess:
           - parseDate: 02.01.2006
       Image:
-        selector: //script[contains(text(),"myFluidPlayer")]
+        selector: $sceneWrapper//script[contains(text(),"myFluidPlayer")]
         postProcess:
           - replace:
             - regex: .+posterImage\s?:\s?'(.+?)'.+
@@ -32,13 +34,9 @@ xPathScrapers:
           fixed: AlexLegend
       Tags:
         Name:
-          selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-tags ")]/a
+          selector: $sceneWrapper//div[contains(concat(" ", normalize-space(@class), " "), " content-tags ")]/a
       Performers:
         Name:
-          selector: //div[contains(concat(" ", normalize-space(@class), " "), " main__models ")]
-          split: ","
-          postProcess:
-            - replace:
-              - regex: 'Models:'
+          selector: $sceneWrapper//div[contains(concat(" ", normalize-space(@class), " "), " main__models ")]/a[@title]
 
-# Last Updated September 06, 2024
+# Last Updated September 08, 2024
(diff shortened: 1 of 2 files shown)


scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -36,15 +36,13 @@ xPathScrapers:
...
+              - regex: 'Tags:'
       Performers:
         Name:
           selector: //div[contains(concat(" ", normalize-space(@class), " "), " main__models ")]
           split: ","
           postProcess:
             - replace:
-              - regex: ^Models:(.+)
-                with: $1
+              - regex: 'Models:'
```
MortonBridges, line 46: Needs a "with" right about here...

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+        postProcess:
+          - replace:
+            - regex: .+posterImage\s?:\s?'(.*?)'.+
+              with: $1
+      Studio:
+        Name:
+          fixed: AlexLegend
+      Tags:
+        Name:
+          selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-tags ")]
```
feederbox826, line 35: Instead of splitting the main element, get the `/text()` of each inner `/a`

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+      Title: //div[contains(concat(" ", normalize-space(@class), " "), " content-title ")]/h1
+      Code:
+        selector: //link[@rel="canonical"]/@href
+        postProcess:
+          - replace:
+            - regex: .+-(\d+)\..+
+              with: $1
+      Details: //div[contains(concat(" ", normalize-space(@class), " "), " content-desc ")][2]
+      Date:
+        selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-date ")]/div[contains(concat(" ", normalize-space(@class), " "), " label ")]
```
feederbox826, line 21: absurd selector that could've just been `//div[@class="content-date"]/div[@class="label"]/text()`

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+  sceneScraper:
+    scene:
+      Title: //div[contains(concat(" ", normalize-space(@class), " "), " content-title ")]/h1
+      Code:
+        selector: //link[@rel="canonical"]/@href
+        postProcess:
+          - replace:
+            - regex: .+-(\d+)\..+
+              with: $1
+      Details: //div[contains(concat(" ", normalize-space(@class), " "), " content-desc ")][2]
```
feederbox826, line 19: description is also available in meta when not available in the page

[example](https://alexlegend.com/video/skylar-snow-black-lingerie-surprise-sex-behind-the-scenes-1074.html)

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+      Date:
+        selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-date ")]/div[contains(concat(" ", normalize-space(@class), " "), " label ")]
+        postProcess:
+          - parseDate: 02.01.2006
+      Image:
+        selector: //script[contains(text(),"myFluidPlayer")]
+        postProcess:
+          - replace:
+            - regex: .+posterImage\s?:\s?'(.*?)'.+
```
feederbox826, line 28: `posterImage : '(.+?)'` since you're escaping a single space

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+sceneByURL:
+  - action: scrapeXPath
+    url:
+      - alexlegend.com/video/
+    scraper: sceneScraper
+
+xPathScrapers:
+  sceneScraper:
+    scene:
+      Title: //div[contains(concat(" ", normalize-space(@class), " "), " content-title ")]/h1
```
feederbox826, line 12: normalize space is unnecessary

scrapers/AlexLegend/AlexLegend.yml
```diff
@@ -0,0 +1,48 @@
...
+      Tags:
+        Name:
+          selector: //div[contains(concat(" ", normalize-space(@class), " "), " content-tags ")]
+          split: ", "
+          postProcess:
+            - replace:
+              - regex: 'Tags:'
+      Performers:
+        Name:
+          selector: //div[contains(concat(" ", normalize-space(@class), " "), " main__models ")]
```
feederbox826, line 42: Could've just grabbed `@title` of the element instead of stripping nbsp and `,`

feederbox826 (changes requested):
The site does seem to have fragmented, but the selectors are absurd and unnecessarily complex

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: That is a fundamentally different selector that does not accommodate modern dynamic styles of web pages. Something as simple a new style class would be likely to break the scraper which is the purpose of the contains normalize-space pattern.

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: That is a fundamentally different selector that does not accommodate modern dynamic styles of web pages. Something as simple a new style class would be likely to break the scraper which is the purpose of the contains normalize-space pattern.

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: This regex is run against a block of code which is likely to become changed in the future (and is likely dynamically generated). The group should be (.+?) instead of (.*?) but the spaces should be handled as something like a site library update would otherwise break the expression or match unrelated sections of the script.

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: fixed

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: Meta description is the page description and not related to the scene.
Meta description template: 'Watch {SCENE_NAME} on AlexLegend, the best HD...'

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: fixed

scrapers/AlexLegend/AlexLegend.yml
WellDoneLiz: updated to not match empty image value

feederbox826:
as noted in my comment https://github.com/stashapp/CommunityScrapers/pull/2014#issuecomment-2336836506

`contains()` achieves the purpose of your normalize-spaces, if you want to introduce the same redundancy then `contains(" target ")` will accomplish the same goal with more readable code

WellDoneLiz:
> `contains()` achieves the purpose of your normalize-spaces, if you want to introduce the same redundancy then `contains(" target ")` will accomplish the same goal with more readable code

Appending a space to the end of a class name will not match a className that only has a singular member. For example `class="target"` is not matched with contains(" target "). `class="target"` and `class="target nottarget"` are both matched by the normalize space match pattern while not matching `class="nottarget"`.

feederbox826:
While I would prefer to update more frequently instead of sacrificing readability, I'd rather not let the project suffer or get stuck on this.

PR #2014, opened 2024-09-07
remove normalize-space chains

feederbox826:
the `contains(concat(' ', normalize-space(@class),' ')` chain would ensure that the classname had successive spaces collapsed, which only guaranteed that it was part of a multi-class selector. Since only a single class is being selected, this is completely redundant since consecutive spaces don't impact anything.

The trailing normalize-space(.) would make sure anything ending with `...` would be collapsed to `.` but this was not observed on any of the pages, and when it was observed, it was intentional styling.

The PornHub scraper was unraveled and and a common fragment added for $videoActions

### DIFF
--- a/scrapers/Boundhoneys.yml
+++ b/scrapers/Boundhoneys.yml
@@ -9,7 +9,7 @@ xPathScrapers:
     scene:
       Title: //div[@class="updateVideoTitle"]
       Details:
-        selector: //div[@class="updateDescription"]/b/text()[normalize-space(.)]
+        selector: //div[@class="updateDescription"]/b/text()
         concat: "\n\n"
       Image: //meta[@name="twitter:image"]/@content
       Studio:
--- a/scrapers/CherryPimps.yml
+++ b/scrapers/CherryPimps.yml
@@ -46,7 +46,7 @@ xPathScrapers:
           - parseDate: January 2, 2006
       Performers: &perfAttr
         Name: //div[@class='update-info-block models-list-thumbs']//span/text()
-      Details: &detailsSel //h4[normalize-space('Description:')]/following-sibling::p/text()
+      Details: &detailsSel //h4[contains(text(),"Description:")]/following-sibling::p/text()
       Tags: &tagsAttr
         Name:
           selector: //ul[@class='tags']/li/a/text()
--- a/scrapers/Dickdrainers.yml
+++ b/scrapers/Dickdrainers.yml
@@ -9,7 +9,7 @@ xPathScrapers:
     scene:
       Title: //div[@class="videoDetails clear"]//h3
       Details:
-        selector: //div[@class="videoDetails clear"]/p//text()[normalize-space(.)]
+        selector: //div[@class="videoDetails clear"]/p//text()
         concat: "\n\n"
       Date:
         selector: //span[contains(text(), "Date Added")]/following-sibling::text()
@@ -29,6 +29,6 @@ xPathScrapers:
       Tags:
         Name: //li[@class="label"][contains(text(),"Tags:")]/following-sibling::li/a/text()
       Performers:
-        Name: //li[@class="update_models"]//a/text()[normalize-space(.)]
+        Name: //li[@class="update_models"]//a/text()
 
 # Last Updated January 08, 2021
--- a/scrapers/FreeonesCommunity.yml
+++ b/scrapers/FreeonesCommunity.yml
@@ -66,12 +66,12 @@ xPathScrapers:
     scene:
       Title: //h1
       URL: //link[@rel="alternate"][1]/@href
-      Details: $commonRoot//div[contains(concat(' ',normalize-space(@class),' '),' pb-2 ')]
+      Details: $commonRoot//div[contains(@class,"pb-2")]/text()
       Studio:
         Name: $commonRoot//span[@data-test="link_span_Studio"]
       Director: $commonRoot//span[@data-test="link_span_Director"]
       Date:
-        selector: //div[contains(concat(' ',normalize-space(@class),' '),' mid-content-pr-past-date ')]
+        selector: //div[contains(@class,"uploaded-date")]/text()
         postProcess:
           - replace:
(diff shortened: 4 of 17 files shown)


scrapers/Boundhoneys.yml
```diff
@@ -9,7 +9,7 @@ xPathScrapers:
     scene:
       Title: //div[@class="updateVideoTitle"]
       Details:
-        selector: //div[@class="updateDescription"]/b/text()[normalize-space(.)]
```
feederbox826, line 12: No spaces found

scrapers/CherryPimps.yml
```diff
@@ -46,7 +46,7 @@ xPathScrapers:
           - parseDate: January 2, 2006
       Performers: &perfAttr
         Name: //div[@class='update-info-block models-list-thumbs']//span/text()
-      Details: &detailsSel //h4[normalize-space('Description:')]/following-sibling::p/text()
```
feederbox826, line 49: normalize-space with no spaces surrounding it is useless since it collapses spaces. contains is more lenient

scrapers/Dickdrainers.yml
```diff
@@ -9,7 +9,7 @@ xPathScrapers:
     scene:
       Title: //div[@class="videoDetails clear"]//h3
       Details:
-        selector: //div[@class="videoDetails clear"]/p//text()[normalize-space(.)]
```
feederbox826, line 12: multiple `...` was intentional

scrapers/Dickdrainers.yml
```diff
@@ -29,6 +29,6 @@ xPathScrapers:
       Tags:
         Name: //li[@class="label"][contains(text(),"Tags:")]/following-sibling::li/a/text()
       Performers:
-        Name: //li[@class="update_models"]//a/text()[normalize-space(.)]
```
feederbox826, line 32: No normalized spaces to be found

scrapers/FreeonesCommunity.yml
```diff
@@ -66,12 +66,12 @@ xPathScrapers:
...
       Title: //h1
       URL: //link[@rel="alternate"][1]/@href
-      Details: $commonRoot//div[contains(concat(' ',normalize-space(@class),' '),' pb-2 ')]
+      Details: $commonRoot//div[contains(@class,"pb-2")]/text()
       Studio:
         Name: $commonRoot//span[@data-test="link_span_Studio"]
       Director: $commonRoot//span[@data-test="link_span_Director"]
       Date:
-        selector: //div[contains(concat(' ',normalize-space(@class),' '),' mid-content-pr-past-date ')]
+        selector: //div[contains(@class,"uploaded-date")]/text()
```
feederbox826, line 74: Updated, I guess?

scrapers/Mandyflores.yml
```diff
@@ -12,7 +12,7 @@ xPathScrapers:
     scene:
       Title: //span[@class="title_bar_hilite"]
       Details:
-        selector: $updateDesc$divCenter/span/span[@style]/text()[normalize-space(.)]|($updateDesc | $updateDesc/p)/text()[normalize-space(.)]|($updateDesc$divCenter/text())[1]|($updateDesc/text())[1]
+        selector: $updateDesc$divCenter/span/span[@style]/text()|($updateDesc | $updateDesc/p)/text()|($updateDesc$divCenter/text())[1]|($updateDesc/text())[1]
```
feederbox826, line 15: No evidence that it was normalizing anything

scrapers/Pornhub.yml
```diff
@@ -146,7 +147,12 @@ xPathScrapers:
...
+          selector: $videoActions//div[contains(@class,"categoriesWrapper")]//a/text()
+            |$videoActions//div[contains(@class,"tagsWrapper")]//a/text()
+            |$videoActions//div[contains(@class,"productionWrapper")]//a/text()
+            |$videoActions//div[contains(@class,"langSpokenWrapper")]//a/text()
+            |$videoActions//div[contains(@class,"relatedSearchTermsContainer")]//a/text()
+            |//div[contains(@class,"userInfo")]//div[contains(@class,"usernameBadgesWrapper")]//i/text()
```
feederbox826, line 155: exploded selectors

scrapers/Pornhub.yml
```diff
@@ -117,6 +117,7 @@ xPathScrapers:
     common:
       $datablob: //script[contains(., 'VideoObject')]/text()
       $videowrap: //div[@class="video-wrapper"]
+      $videoActions: //div[contains(@class,"video-actions-tabs")]
```
feederbox826, line 120: added common fragment

feederbox826 (commented):
explanation of other changes

WellDoneLiz:
Some changes like the CherryPimps one make sense but others like EuroPornstar introduce fragility into the selector. For example a wrapper element with a classname of about-desc-wrapper or a new element like about-desc-info would break the EP scraper when currently the scraper is resilient to such changes.
While increased readability is nice IMO most of the changes here are downgrades and should not be introduced.

feederbox826:
> a wrapper element with a classname of about-desc-wrapper or a new element like about-desc-info would break the EP scraper when currently the scraper is resilient to such changes.

that's what the contains() is for, it's more resilient since your selector only works if the class name is surrounded by other classes

`class="x target z"` while contains will wildcard match anywhere in the string, allowing for `x target`, `x target y z` and `target`

WellDoneLiz:
The wildcard match in the string is the problem with contains (ex. contains 'target' matches 'target-xyz'). The common pattern of normalizing the spaces within the class string and appending spaces to each end creates a standard formatting for exact matches.

Contains without the exact match normalization is far less resilient as unrelated classes with the same prefix (or even just as the name implies containing) also match.

feederbox826:
While I would prefer to update more frequently instead of sacrificing readability, I'd rather not let the project suffer or get stuck on this. Will port out the PH improvements to another PR. Closing this and merging the other PR

PR #2023, opened 2024-09-12
Update PervCity scraper

SpecialKeta:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

[Scene url](https://upherasshole.com/trailers/alt-babes-bonnie-rotten-juelz-ventura-get-nasty.html) (Up Her Asshole)
[Scene url](https://analoverdose.com/trailers/taboo-anal-sex-three-way-part-2.html) (Anal Overdose)
[Scene url](https://www.dpdiva.com/trailers/Stunning-Latina-Vanessa-Sky-Intense-DP.html) (DP Diva)

## Short description
- Tags for scenes
Tags can only be grabbed on the pervcity.com website, not on the individual subsites.
The slug is the same for a scene, so I used queryURLReplace to replace domain to pervcity.com to grab the tags.

- Updated regex for dates 
Dates for scenes are only displayed on the pervcity.com website, when searching for a scene.
So searching for exact title that has a ``’`` or ``,`` would return more than 1 scene, and most of the time the wrong one.

- Map the studio to their corresponding studiocode
So for example, studiocode ``ao-0328`` maps to studio ``Anal Overdose``
In some cases, it's hard to tell which studio that scene belongs to. For example when the studiocode is ``BTS`` or ``SOLO``, it could be any studio, so I've mapped those to the main PervCity studio. We could always move it to the correct studio later on.

    Out of the 1050 scenes I used this scraper on, I've found only 5 or so scenes which had the anal overdose studiocode, but only could be found on upherasshole.com.

- Higher quality image
Grabs higher quality image. I've noticed they started higher quality scene images around 2021-01, so anything before that will give you low-quality.

- Moved dpdiva.com to own scraper
Because I used queryURLReplace to map urls to pervcity.com and DP Diva scenes can **only** be found on dpdiva.com, I moved DP Diva to it's own subscraper

### DIFF
--- a/scrapers/PervCity.yml
+++ b/scrapers/PervCity.yml
@@ -1,4 +1,4 @@
-name: "PervCity"
+name: PervCity
 sceneByURL:
   - action: scrapeXPath
     url:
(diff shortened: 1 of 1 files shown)


feederbox826:
LGTM! Nice work!

PR #2037, opened 2024-09-20
Update CockyBoys.yml

gimmeliina:
Get higher quality thumbnail from comments.

### DIFF
--- a/scrapers/CockyBoys.yml
+++ b/scrapers/CockyBoys.yml
@@ -34,7 +34,11 @@ xPathScrapers:
               - regex: Description\s*
                 with:
       Image:
-        selector: //meta[@property="og:image"]/@content
+        selector: //html/head/comment()[contains(., "og:image")]
+        postProcess:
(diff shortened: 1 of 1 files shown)


feederbox826:
The thumbnail quality doesnt seem to improve, it looks like it's just a bicubic upscaled version of the 1x picture, with 4x being upscaled to 4x with no actual fidelity improvement. I'd rather keep the low-res one that isn't stretched up, unless you have found scenes with actual quality improvements

gimmeliina:
Old Scraper doesn't use even 1x image but special low-res thumbnail that is awful quality.

If you checked first scene on the page, (a secret place), those are series and exceptions and for those scenes this scraper return low quality scaled version.

feederbox826:
I find 2x to be indiscernable from 4x, I guess I did stumble on the exceptions. Since StashDB rescales anyways, I see no harm in 4x once they come up with proper highres

PR #2048, opened 2024-09-28
Update JakeCruiseMedia.yml

SailorSilly:
Submitting fix to fix incorrect class for Height.

## Examples to test

https://www.jakecruise.com/tour/models/eduardo.html
https://www.jakecruise.com/tour/models/jake-cruise.html

## Short description

When you try to scrape the performer, instead of pulling in the performers height, it's currently programmed to pull in their penis size. The above proposed fix, corrects the span class that the scraper is looking for.


feederbox826:
Thanks!

PR #2052, opened 2024-09-28
Update HelixStudios.yml

SailorSilly:
Replacing Measurements with PenisLength. Position will also now pull in as a tag.

## Examples to test

https://www.helixstudios.com/model/1006/hayden-clark
https://www.helixstudios.com/model/2545/adam-reid

## Short description

I was able to fix the PenisLength, but couldn't figure out if it was possible to also scraper whether they were cut or uncut. On the studio's website it's listed as 8 inches uncut and they don't have uncut or cut as it's own separate item.

### DIFF
--- a/scrapers/HelixStudios.yml
+++ b/scrapers/HelixStudios.yml
@@ -113,16 +113,19 @@ xPathScrapers:
         selector: //span[text()="Height"]/following-sibling::text()
         postProcess:
           - feetToCm: true
-        PenisLength:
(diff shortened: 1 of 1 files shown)


feederbox826:
Thanks! The cut/uncut is a bit tricky and it also excludes some less well-endowed performers <10in

feederbox826:
Argh, hit enter early. Added Cut/Uncute selector, merged the two regexes into one

PR #2059, opened 2024-10-01
Tested and working on AnalMom.com

TheStashMaster:
PaperstreetMedia scraper was not working for me on a couple sites so I did some updates.  I've tested on AnalMom, pervdoctor, and dadcrush but not the entire list.  Hope this helps.


feederbox826:
Thanks! I'll test the rest but it might be best to just adapt it to TeamSkeet in the long term

PR #2060, opened 2024-10-04
SmutPuppet scraper update

SpecialKeta:
## Scraper type(s)
- [x] sceneByURL

[scene url](https://porn-uk.com/update/2256/)
[scene url](https://dothewife.com/update/81/)

## Short description

Updated Smut Puppet scraper
- Added sub-sites that were missing. 
Removed smutpuppet.com/update from the urls, it always re-directs to porngutter.com/tour, no matter what.
- Grab studio name by searching for scene at networksite porngutter.com, found out it's the most reliable way. 
 For example, https://dothewife.com/update/81/ or https://smutmerchants.com/update/81/ both give you the same scene, but different studio when you scrape it. The search always gives you the correct studio.
- Grab date for scene from networksite porngutter.com. Not all subsites have a date for their scenes.
For example, this scene https://smutmerchants.com/update/1487/ from Smut Merchants, doesn't have a date, but https://porngutter.com/update/1487/ does.
- Moved ``Jeff's Models`` to it's own sub scraper. Their scenes are not listed on porngutter.com, so studio name & scene date lookup wouldn't work.
- Fixed code & url sanitizer for older scenes with just 2 digits in the url
- Fixed image scraping.
- Removed sceneByFragment block. The sceneByFragment should be used to manipulate the title/filename so it can find the scene from the query url, or pass it onto a script, but here it did nothing of that.

### DIFF
--- a/scrapers/SmutPuppet.yml
+++ b/scrapers/SmutPuppet.yml
@@ -81,7 +81,7 @@ xPathScrapers:
         concat: "|"
         postProcess:
           - replace:
-              - regex: \/static\/(\w+\.[a-z]{3})\/.+\/api\/update\/(\d{2,4})\/.+
+              - regex: \/static\/(.+?\.[a-z]{3})\/.+\/api\/update\/(\d{2,4})\/.+
                 with: https://$1/update/$2/
       # Grab date from porngutter.com
       Date:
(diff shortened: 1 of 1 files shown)


Maista6969:
Solid job, thanks for fixing this!

PR #2079, opened 2024-10-25
add scraper TabooHandjobs.yml

sholmesta:
adding scraper for TabooHandjobs (similar to TabooPOV and Jerk off Instructions).


feederbox826:
Ugh, I wish we could just merge the three into one file, will look into it before merging

feederbox826:
LGTM, having it split into three is annoying but would have to be merged into python

PR #2083, opened 2024-10-29
Fix Ifeelmyself scraper imf_prefs cookie

pops64:
The imf_prefs cookie needed updating adds `s:11:"search_type";s:7:"content";` to the cookie string. Also changed the debug prompts to unify the code format. Also very old videos have only 2 digits for the media-id. The regex wouldn't find it as it specified for three. 

No bugs apparent


feederbox826:
Thanks!

Issue #2099, opened 2024-11-09
GayAdultEmpire - Scraping incorrect data from individual scenes


AbyssDarkVoid:
Trying to scrape a scene from a movie.

The scene is linked here: https://www.gaydvdempire.com/clip/836895/scene-2-streaming-scene.html

When I try to scrape the URL, it is trying to pull in Protein Shake as the Studio name rather than Trailer Trash Boys for the Studio.

It's also wanting to pull in Scene 2 or Scene 2 from Protein Shake when it should be pulling in the group name of Protein Shake.

This is the link to the actual movie on GayAdultEmpire.

https://www.gaydvdempire.com/3190601/protein-shake-porn-videos.html

If I try and scrape that URL as a new scene, it will populate the information correctly

No error messages to share.

MortonBridges:
#2102 created for this issue.

PR #2102, opened 2024-11-12
Update GayAdultEmpire.yml

MortonBridges:
Changed sceneByUrl studio selector to account for scenes that have more than one studio label attribute.  Addressing #2099.

## Scraper type(s)
- [x] sceneByURL

## Examples to test
Example of a link having the issue before this change:
https://www.gaydvdempire.com/clip/836894/scene-1-streaming-scene.html

Example of a link that had only one studio label to verify these scenes are not affected by change:
https://www.gaydvdempire.com/clip/1698707/angel-elias-riding-marco-alessandro-streaming-scene.html

## Short description

Studio selector for sceneByURL was looking for a specific tag.  Most scenes would scrape file, while some would be in the format of "from MOVIE by STUDIO" where both "MOVIE" and "STUDIO" would have the tag used in the selector.  Before the change we were matching only on the first tag found.  The change will chain all values of the tag and select the last one in the list, on the assumption that the final tag will be the studio.

### DIFF
--- a/scrapers/GayAdultEmpire.yml
+++ b/scrapers/GayAdultEmpire.yml
@@ -78,7 +78,13 @@ xPathScrapers:
       Director: //a[@label="Director"]/text()
       Image: //a[@id="front-cover"]/@data-href
       Studio:
-        Name: //a[@label="Studio"]/text()
+        Name: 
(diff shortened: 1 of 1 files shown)


feederbox826:
LGTM

PR #2107, opened 2024-11-14
Update PMVHaven Scraper

S3L3CT3DLoves:
Restores functionality to the PMVHaven scraper.

The latest update to the scraper removed the Music from the description. Music is an essential part of PMVs, unfortunately Stash does not have a dedicated field for it.
So, restoring the feature: Music is appended to the description.


Maista6969:
I don't think the music was removed by any update to scraper, perhaps you had a local modification that was lost when the scraper was updated? Either way including the track list makes sense to me, thanks for contributing :)

PR #2133, opened 2024-12-11
Rename MFVIdeoXXX.yml to MFVideoXXX.yml

ryosaeba75:
## Short description

Apparently I can't write a name correctly...


Maista6969:
The filename doesn't really matter, Stash only uses the name defined in the YAML itself and that's correct 😁

ryosaeba75:
> The filename doesn't really matter, Stash only uses the name defined in the YAML itself and that's correct 😁

Maybe, but that doesn't mean that wouldn't have bothered me ;) Thanks for the quick merge.

PR #2138, opened 2024-12-16
fix: VirtualRealTrans: maintain newlines in scene description

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://virtualrealtrans.com/vr-trans-porn-video/best-compilation/
- https://virtualrealtrans.com/vr-trans-porn-video/symptoms-of-attraction/
- https://virtualrealtrans.com/vr-trans-porn-video/valentines-surprise/ (this contains VirtualRealTrans.com. in the description, which shows the change adds a newline for the correct `.` scenario)

## Short description

Currently, the description is parsed in such a way that the source newlines from the `<br>` tags are being dropped entirely, resulting in the last word and `.` in a paragraph being followed immediately by the first character of the next paragraph.

This change uses this consistent parsing behaviour to add newlines at these points, to maintain the intended paragraph separation.

### Notes

Simple regex has been used instead of using non-capturing, look-behind, look-ahead, `\w` (word characters), as it appears stashdb regex implementation is limited/basic and does not properly support these features.

### DIFF
--- a/scrapers/VirtualRealTrans.yml
+++ b/scrapers/VirtualRealTrans.yml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=../validator/scraper.schema.json
 name: VirtualRealTrans
 
 sceneByURL:
@@ -60,6 +61,10 @@ xPathScrapers:
       Details: &detailsSel
         selector: //div[@class="description_container"]/p
         concat: "\n\n"
+        postProcess:
+          - replace:
+              - regex: ([A-Za-z])(\.)([A-Za-z])
+                with: "$1$2\n\n$3"
       Tags:
         Name:
           selector: //div[@class="metaHolder"]//a/span/text()|//script[@type="application/ld+json"][contains(text(),"genre")]/text()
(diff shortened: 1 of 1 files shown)


feederbox826:
just for future reference
> stashdb regex implementation is limited/basic and does not properly support these features.

golang does not support backreferences https://groups.google.com/g/golang-nuts/c/7qgSDWPIh_E/m/OHTAm4wRZL0J

feederbox826 (changes requested):
how you break on <br> is not consistent with the actual page, I'd rather use the actual breaks in the site.

For example, https://virtualrealtrans.com/vr-trans-porn-video/symptoms-of-attraction/ has no <br> and should be presented as such

feederbox826:
https://guidelines.stashdb.org/docs/scenes/edit/scene-description/correcting-scene-descriptions/

there are no `<br>`s present on the page and the current scraper handles https://virtualrealtrans.com/vr-trans-porn-video/best-compilation/ correctly as it is displayed, with breaks between `<p>` paragraphs

PR #2145, opened 2024-12-21
Update Coomer to get correct performer information

QuentinJones:
## Short description

Lookup performer name by id and use it to return correct performer/studio name and performer url
Also removed a debug line from url lookup which always showed an error when there wasn't any


feederbox826:
will merge for now, but I think we might want to move it over to the FansDB repo in the long-term

PR #2148, opened 2024-12-24
[nhentai] Add Parodies into Stash Tags.

WeedLordVegeta420:
Updates nhentai scraper to include "Parodies" as Tags.


feederbox826:
Thanks!

PR #2150, opened 2024-12-28
Fixes cases where thumbnail is indexed at 0

Strad:
Sometimes the thumbnail index is 0 which Python was treating as False instead of 0 causing the condition to fail.


Maista6969:
Oof, what a clumsy mistake on my part 😅 Thanks for the fix!

PR #2153, opened 2024-12-31
[scrapeCoverFromFile] add

stg-annon:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [x] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL (formerly movieByURL)
- [ ] galleryByFragment
- [ ] galleryByURL

## Examples to test
Any scene with  a image file with the same name or a cover name in the same dir as the scene file

## Short description

"scrapes" an image file from the local file system into stash

### DIFF
--- a/scrapers/scrapeCoverFromFile/scrapeCoverFromFile.py
+++ b/scrapers/scrapeCoverFromFile/scrapeCoverFromFile.py
@@ -1,0 +1,57 @@
+import re
+import sys
+import json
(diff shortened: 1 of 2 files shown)


feederbox826:
I thought this was from something else, but I must have been mistaken. LGTM

PR #2175, opened 2025-01-22
Fix ashemaletube performerByURL

nrg101:
- [x] performerByURL

## Examples to test

- https://www.ashemaletube.com/model/domino-presley-796-m9nr/
- https://www.ashemaletube.com/model/bailey-jay-308-m0xx/
- https://www.ashemaletube.com/model/daisy-taylor-5700-m1h4/

## Short description

The ashemaletube.com site now has cloudflare protection. This refactor to Python makes use of `cloudscraper` to bypass the protection, and `FreeProxy` to improve the success chance of fetching the page successfully.

Each time a request is made via cloudscraper with a proxy, if a non-200 http status code is retrieved (i.e. usually 403 due to blocking) then another random proxy is used and the request re-attempted, until the response is a 200, or the max retries is reached. When the page is fetched successfully, the proxy used is used again to resolve any outgoing social media links (which are ashemaletube.com links that forward to the respective external URLs)

Due to the cloudflare protection, the scraping is not always successful, but it can be tried again until it succeeds. I would say the success rate is >95% for me, which is much better than it succeeding a constant 0% of the time.

### DIFF
--- a/scrapers/AShemaleTube/AShemaleTube.py
+++ b/scrapers/AShemaleTube/AShemaleTube.py
@@ -1,0 +1,79 @@
+import json
+from lxml import html
+import sys
+
+from py_common import log
+from py_common.deps import ensure_requirements
+from py_common.types import (
+    ScrapedPerformer,
+)
+from py_common.util import scraper_args
+
+ensure_requirements("cloudscraper")
+import cloudscraper  # noqa: E402
+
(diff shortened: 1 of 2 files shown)


Maista6969:
Thank you again for your consistently great contributions! I think we can reuse most of this logic for other sites that are giving us Cloudflare blocks

PR #2197, opened 2025-02-21
Add fragment scrape functionality to Voyeurweb

WithoutPants:
Videos downloaded from Homeclips/Funbags are in a format that the URL can be derived from.

Filenames are typically `xxxxx-xxxx.mp4` - for example `99974-6118.mp4`


feederbox826:
Thanks!

PR #2215, opened 2025-03-03
Coomer update

goawaylovestrike:
Fixed my last update and added new stuff.

Correctly finds tags in description. anything starting with # is extracted ending with a word break
Performers starting with @ should be extracted including all supported characters.
I noticed the title is just the first 65 characters of the description even if there are line breaks. Now the title is the first line of the description if it's 100 characters or less. If the first line is over 100 characters if checks for a string of emojis, then a . then a string of !s and ends it there. All of the words are also capitalized.
I updated the studio so it's now the username and site eg. "My_Name (OnlyFans)"

### DIFF
--- a/scrapers/Coomer/Coomer.py
+++ b/scrapers/Coomer/Coomer.py
@@ -14,8 +14,10 @@ headers = {
     'Referer': 'https://coomer.su/search_hash'
 }
 
-def extract_mentions(text):
-    return re.findall(r'@([^\s<]+)', text) if text else []
+def extract_mentions_and_tags(text):
+    mentions = re.findall(r'@([^\s<]+)', text) if text else []
+    hashtags = re.findall(r'#([^\s<]+)', text) if text else []
+    return mentions, hashtags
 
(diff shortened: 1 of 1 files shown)


goawaylovestrike:
I tested it on a bunch of posts, but I just found a couple of posts with very long first lines that its not working on. I'll fix it.

goawaylovestrike:
I fixed it. I'm not sure if it updated the pull request though, I'm new to github

feederbox826:
> I fixed it. I'm not sure if it updated the pull request though, I'm new to github

You have different changes in each pull request, choose one to keep and i'll close the other

feederbox826:
Conflicting merge changes in both, will just manually resolve and take fragments from https://github.com/FansDB/metadata-scrapers/tree/main/scrapers/Coomer/SHALookup

feederbox826:
Since all of these new improvements were already made in SHALookup on release, will close these since there's way too many file conflicts and will move to remove Coomer and redirect instead to SHALookup

PR #2217, opened 2025-03-04
Update Algolia.py - MAIN_CHANNELS_AS_STUDIO_FOR_SCENE logic

MortonBridges:
Changed logic for MAIN_CHANNELS_AS_STUDIO_FOR_SCENE processing in the determine_studio_name_from_json subroutine, as the check if a site is in the dictionary for studio name processing was not using the proper key.  The value returned in the JSON for this key is a dictionary itself.  Changed the check to match the SITES_SEGMENT_USING_MAIN_CHANNEL_AS_SCENE_STUDIO check in the same block.

Added some debugging lines so it's more obvious which IF block is being triggered to assist in future efforts to hunt down studio name processing issues.

Added BiPhoria to MAIN_CHANNELS_AS_STUDIO_FOR_SCENE  dictionary to allow the scraper to return BiPhoria as the studio instead of  the scene name.

## Scraper type(s)
- [x] sceneByURL

## Examples to test
The following should return "BiPhoria" instead of the scene names:
https://www.biphoria.com/en/video/biphoria/Saying-GoodBi---Scene-3/232742
https://www.biphoria.com/en/video/biphoria/Saying-GoodBi---Scene-1/232740
https://www.biphoria.com/en/video/biphoria/Saying-GoodBi---Scene-2/232741

Another random URL from a studio in the dictionary. Should return "Devil's Film Parodies" as the studio name
https://www.devilsfilm.com/en/video/This-Isnt-UFC---Ultimate-Fucking-Championship-02/29464

### DIFF
--- a/scrapers/Algolia/Algolia.py
+++ b/scrapers/Algolia/Algolia.py
@@ -35,6 +35,7 @@ NON_FEMALE = True
 # a list of main channels (`mainChannelName` from the API) to use as the studio
 # name for a scene
 MAIN_CHANNELS_AS_STUDIO_FOR_SCENE = [
+    "BiPhoria",
     "Buttman",
     "Cock Choking Sluts",
     "Devil's Film Parodies",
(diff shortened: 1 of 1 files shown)


feederbox826:
little typo on L640 but that's a nitpik

tested and working, thanks!

MortonBridges:
> little typo on L640 but that's a nitpik
D'h!

PR #2226, opened 2025-03-09
Fix for Kink scraping details

SpecialKeta:
Narrowed the regex for scraping details
For some scenes the details were incomplete, regex stopped after ``\",`` while the details continued
Scrape the url below to see this behaviour

Updated the regex for details, to remove any left-over whitespaces before a newline character

## Scraper type(s)
- [x] sceneByURL

## Examples to test
https://www.kink.com/shoot/36182


feederbox826:
Thanks!

PR #2230, opened 2025-03-19
Update Clips4Sale.yml

sMtzDczDq:
Scraper would extract tags as performers.


Maista6969:
This is working as intended: since Clips4Sale does not have a proper way to credit performers, they often appear among the tags

PR #2231, opened 2025-03-21
Add some Improvements, Make Scraper compatible with JF 10.9+

masterr2001:
Jellyfin 10.9 changed the Way Items can be referenced by URL. While the old pattern still works, it is deprecated now and will be removed in the future. Therefore the URL-Syntax has been updated to match the new Standard. Furthermore I implemented two configurable Options in the config.ini: If "Scene2Movie" is set to true, an Movie will be automatically created when scraping a new Scene. This is helpful if you have mainly Movies in Jellyfin and wants to create Movies in Stash accordingly. If "IgnoreSceneImage" ist set to true, the Primary Image for the Scene will not be pulled from Jellyfin. This is helpful if you have Coverart in Jellyfin for your Movies, but want to make sure that you only use it for movies, not for the Scenes


feederbox826:
I don't have a JF instance, would appreciate someone else being able to test

feederbox826:
Unable to test, will just merge and revert if any complaints

PR #2247, opened 2025-04-02
AniDB scraper pulls scene tags from complete tag box

spaceyuck:
## Scraper type(s)
- [X] sceneByURL

## Examples to test
(examples are intentionally SFW, as NSFW requires account, but tags work the same)

* https://anidb.net/anime/17617
* https://anidb.net/anime/17870

## Short description

The AniDB scraper was pulling the scene tags from a small box on the left, which only had a few preselected tags. This pulls tags from the big tag box towards that bottom, which is much more detailed.

the scene tags returned must fulfill some conditions:
* either no level (grey star), or at least 1 star (is just a XPath filter on the weight attribute)
* is from group indicators, elements, fetishes (ignoring stuff like technical aspects, origins, setting, themes)

### DIFF
--- a/scrapers/AniDB.yml
+++ b/scrapers/AniDB.yml
@@ -185,7 +185,9 @@ xPathScrapers:
         selector: //div[@itemprop="description"]//text()
         concat: " "
       Tags:
-        Name: $info//div[@id="tab_1_pane"]//span[@class="tagname"]
+        #Name: $info//div[@id="tab_1_pane"]//span[@class="tagname"]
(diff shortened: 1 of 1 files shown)


feederbox826 (approved):

feederbox826:
Thanks! Some of the weightless tags seem to be caught as well but that seems intentional and fits the scraper

spaceyuck:
> Some of the weightless tags seem to be caught as well but that seems intentional

Yes, there are some very ... cultured ... things in there, many sub-tags or deeper seem to be weightless.

PR #2249, opened 2025-04-04
Czech vr use cdp

TheStashMaster:
I had found that the scraper for www.czechvr.com had stopped working.  I added the CDP and that seemed to resolve the issue for me.


feederbox826:
LGTM without cdp, would recommend it as a troubleshooting step but not required yet

PR #2270, opened 2025-04-18
Update LucasEntertainment.yml

MortonBridges:
For those who happen to have full movies instead of individual scenes, but want the performers and other info in the scenes view, it is handy to scrape the movies URLs.  The existing LucasEntertainment scraper is nearly 100% compatible when scraping the movies URLs as scenes.

This update adds the movie URLs to the Scene Scraper, along with a modified Scene by URL Image selector to account for double quotes in the "movies" versions of the script where the image URL is taken.  The movie URLs also show the dates with the Jan 2, 2006 format, so a new parseDate postProcess was added, alongside the existing one that handles the format from the "scene" URLs.

Existing Scene By URL scrapes should work for individual scenes without issue.

## Scraper type(s)
- [x] sceneByURL

## Examples to test

Movie URL to scrape as a scene:  https://www.lucasentertainment.com/movies/view/flexed-and-worshipped
Scene URL to scrape, to ensure the changes did not introduce new problems:  https://www.lucasentertainment.com/scenes/play/michael-lucas-bottoms-for-louis-ricaute

### DIFF
--- a/scrapers/LucasEntertainment.yml
+++ b/scrapers/LucasEntertainment.yml
@@ -2,7 +2,9 @@ name: Lucas Entertainment
 sceneByURL:
   - action: scrapeXPath
     url:
+      - lucasentertainment.com/movies/
       - lucasentertainment.com/scenes/
+      - lucasentertainment.com/tour/movies/
       - lucasentertainment.com/tour/scenes/
     scraper: sceneScraper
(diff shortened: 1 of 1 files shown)


feederbox826:
Nice work and good catch on chaining parseDates, should probably add that to wiki

PR #2271, opened 2025-04-18
Sheer: Doing the 2025 annual update required for date scraping

robbydurano:
The comment prior to the updated code is pretty self-explanatory for why this commit is required.


feederbox826:
LOL a bit overdue, maybe would be better as a js function?

PR #2285, opened 2025-04-21
[Coomer] Avoid filling RAM for sha256, speedup by matching hash in filename

you-cant-see-me:
## Short description

Improves scraping Coomer sceneByFragment by:
- when calculating sha256, using `hashlib.file_digest()` which will hash 256kb at a time, thus not wasting RAM by reading the entire file at once
- if present, uses sha256 hash from filename, since many download tools include it, thus skipping the sha256 calculation and giving immediate results

I realized after making this work that `hashlib.file_digest()` was added in Python 3.11, is support for Python 3.10 (or earlier) required?

### DIFF
--- a/scrapers/Coomer/Coomer.py
+++ b/scrapers/Coomer/Coomer.py
@@ -2,6 +2,7 @@ import sys
 import json
 import hashlib
 import stashapi.log as log
+import pathlib
 import requests
 import re
 from bs4 import BeautifulSoup as bs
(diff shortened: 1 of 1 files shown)


feederbox826:
> when calculating sha256, using hashlib.file_digest() which will hash 256kb at a time, thus not wasting RAM by reading the entire file at once

it shouldn't eat the ram 4MB at a time, this was done for faster reads since each hash update is a cpu operation

> if present, uses sha256 hash from filename, since many download tools include it, thus skipping the sha256 calculation and giving immediate results

fair enough, I wanted to calculate sha256 deliberately to avoid corrupted files

> I realized after making this work that hashlib.file_digest() was added in Python 3.11, is support for Python 3.10 (or earlier) required?

CommunityScrapers targets latest python and nothing else

Also Coomer scraper here will be held in hiatus if I don't get around to deleting it, can you instead PR to https://github.com/FansDB/metadata-scrapers/tree/main/scrapers/Coomer/SHALookup

feederbox826:
#2297

you-cant-see-me:
cheers, i have been a bit busy lately. the chunked hashing method over there looks great :+1:

PR #2288, opened 2025-04-21
Add wifey to vixen network

MortonBridges:
Add Wifey (wifey.com) to Vixen Newtork scraper.

## Scraper type(s)
- [x] sceneByURL

## Examples to test
https://www.wifey.com/videos/victoria-and-leo-cute-blonde-hotwife-gets-fucked-doggystyle

## Short description
vixenNetwork.yml:  

Added wifey.com to sceneByURL scraper.

vixenNetwork.py:
Added Wifey to the studios dictionary.

Changed "isValidURL" processing to use split instead of lstrip and rstrip, as the previous implementation was coded under a wrong assumption on how the strip functions work, which resulted in the 'w' from wifey from being stripped.  The new implementation splits on '.' and takes the second element, so www dot wifey dot com becomes "wifey" instead of "ifey".

I've only tested scrape by URL.  Someone more comfortable with the other methods that this scarper supports should test those out (sceneByName, sceneByQueryFragment and sceneByFragment).

### DIFF
--- a/scrapers/vixenNetwork/vixenNetwork.py
+++ b/scrapers/vixenNetwork/vixenNetwork.py
@@ -272,7 +272,7 @@ class Site:
         up = urlparse(u)
         if up.hostname is None:
             return False
-        if up.hostname.lstrip("www.").rstrip(".com") == self.id.lower():
+        if up.hostname.split('.')[1] == self.id.lower():
             splits = u.split("/")
             if len(splits) < 4:
                 return False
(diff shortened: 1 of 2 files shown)


feederbox826:
weird, I thought I added support for it already

feederbox826:
Will give it some testing and some more love tomorrow

feederbox826:
A very nice side effect of this is that it now supports `members.site.com` links

feederbox826 (approved):

PR #2296, opened 2025-04-24
Babepedia remove url from aliases, fix image URL.

MortonBridges:
Remove URLs from aliases, fix image URL

## Scraper type(s)
- [x] performerByURL

## Examples to test
https://www.babepedia.com/babe/Tiffany_Mynx

## Short description
Two fixes:
1 - Added "catch all" to remove URLs from aliases.  The performer above had URL elements included as aliases prior to this fix.  After this change the leaf of the URL is provided as a potential alias.

2 - The IMG tag used in the Image selector has changed.  The original selector was matching against a single class, but a new class has been added to the tag.  This change uses a "contains" instead of strict equality to match the class.

Note:  Images still will not scrape properly due to hot linking protection at the studio site.  The image URL will appear in the StashApp logs, and can be used to update the image manually.


feederbox826:
> Note: Images still will not scrape properly due to hot linking protection at the studio site. The image URL will appear in the StashApp logs, and can be used to update the image manually.

the user agent was added to fix hotlinking, that worked for me and I was able to reliably pull the images, will give it another shot

feederbox826:
Image scraped successfully on this PR + v0.28.1

feederbox826 (approved):

DogmaDragon:
I think the header is not being passed correctly as part of the scraper. 

I tested with `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0` set as **Scraper User Agent** inside Stash and without. 

It worked with, but failed without.

feederbox826:
I set it in both locations, that might be why :/

Zambucco:
> I set it in both locations, that might be why :/

I'm new to this, how do you set the user-agent?

PR #2304, opened 2025-04-29
fix date scraping in virtualrealporn scraper

nrg101:
- [x] sceneByURL

## Examples to test

- https://virtualrealporn.com/vr-porn-video/anime-girl/
- https://virtualrealporn.com/vr-porn-video/dress-code/

## Short description

Currently the wrong element on the page is selected for scraping, e.g. "VR180 3D" as it is a fixed index of an array. As the regex pattern is reasonably robust to work for the whole div text, this change just concats the whole div to let the regex pick out the date string to parse.


feederbox826:
wouldn't it make more sense to grab the last element instead?

PR #2324, opened 2025-05-12
delete old scraper: all sites / scraper configs migrated to AdultTime

nrg101:
## Short description

The new AdultTime scraper (`AdultTime/AdultTime.yml` and `AdultTime/AdultTime.py`) includes all the sites and scraper configs in this old scraper, so this one can now be deleted.


Maista6969:
Thanks, I forgot about that one :)

PR #2326, opened 2025-05-12
fix lewood.com URL scraping via evilangel

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://www.evilangel.com/en/video/evilangel/Cream-Of-Cleavage-Sofis-Natural-Jugs/126145

## Short description

The lewood.com uses AdultEmpireCash so the scene IDs and URLs are completely different

e.g. corresponding scene URL: https://www.lewood.com/508940/lewood-black-haired-babe-uses-a-vibrating-wand-and-enjoys-a-big-dick-streaming-scene-video.html


feederbox826:
Not sure I understand why the change was made, will defer instead to @Maista6969

PR #2327, opened 2025-05-12
add erodougazo.com

### LINKED ISSUE #1929: Scrapers Request: erodougazo
## Notes

Scrapers Request: erodougazo
website language:  Japanese only

free site , no subscription
contain detailed Performer & film metadata, JAV only

### URLs that you would expect to scrape

[Scene example URL](https://erodougazo.com/actress/av/%E6%96%B0%E4%BA%95%E3%83%AA%E3%83%9E/)
[Performer example URL](https://erodougazo.com/actress/av/%E5%A1%94%E4%B9%83%E8%8A%B1%E9%88%B4/)

nrg101:
## Scraper type(s)
- [x] performerByURL
- [x] sceneByURL

### performerByURL

- https://erodougazo.com/actress/av/%E6%A4%BF%E3%82%8A%E3%81%8B/
- https://erodougazo.com/actress/av/%E5%A1%94%E4%B9%83%E8%8A%B1%E9%88%B4/

### sceneByURL

- https://erodougazo.com/av/video/445055/
- https://erodougazo.com/av/video/446950/

## Short description

This should fix #1929 (or at least make a decent start on it). I picked out what I could based on Google Translate for the data labels and then class names also helped.

### DIFF
--- a/scrapers/erodougazo.yml
+++ b/scrapers/erodougazo.yml
@@ -36,68 +36,43 @@ xPathScrapers:
         postProcess:
           - javascript: |
               if (value && value.length) {
-                // split the string by comma
-                var parts = value.split(",");
+                // create super regex to match the entire pattern
+                const allRegex = /([A-Z])?カップ,[^\d]*(\d+)?[^\d]*\/[^\d]*(\d+)?[^\d]*\/[^\d]*(\d+)?/;
+                const [_, cup, bust, waist, hips] = value.match(allRegex);
+                if (!cup || !bust || !waist || !hips) return `failed: ${cup}, ${bust}, ${waist}, ${hips}`;
 
-                // check if the first part contains a cup size
-                var cupSizeRegex = /[A-Z]+/;
-                var cupSizeMatch = parts[0].match(cupSizeRegex);
-                if (!cupSizeMatch) {
-                  return "";
-                }
-                var cupSize = cupSizeMatch[0];
+                // metric to imperial conversion
+                const cmToInch = (cm) => Math.round(cm / 2.54);
 
-                // check if the second part contains measurements
-                var measurementsRegex = /(\d+)[^\d]*(\d+)[^\d]*(\d+)/;
-                var measurementsMatch = parts[1].match(measurementsRegex);
-                if (!measurementsMatch) {
-                  return "";
-                }
                 // extract the measurements and convert to inches
-                var overbustCm = parseInt(measurementsMatch[1]);
-                var waistInches = Math.round(parseInt(measurementsMatch[2]) / 2.54);
-                var hipsInches = Math.round(parseInt(measurementsMatch[3]) / 2.54);
+                const overbustCm = Number.parseInt(bust);
+                const waistInches = Math.round(cmToInch(Number.parseInt(waist)));
+                const hipsInches = Math.round(cmToInch(Number.parseInt(hips)));
 
                 // from https://aimerfeel.com/blogs/news/how-to-measure-your-japanese-bra-size#:~:text=Find%20your%20appropriate%20size%20by,Your%20cup%20size%20is%20A.
-                var differenceMap = {
-                    "A": 10,
-                    "B": 13,
-                    "C": 15,
-                    "D": 18,
-                    "E": 20,
-                    "F": 23,
-                    "G": 25,
-                    "H": 28,
-                };
+                const differenceFromCup = (cup) => 10 + Math.ceil(2.5 * (cup.charCodeAt(0) - 65));
 
                 // from https://katherinehamilton.com/pages/international-bra-size-chart
-                var cupSizeMap = {
-                    "A": "AA",
-                    "B": "A",
-                    "C": "B",
-                    "D": "C",
-                    "E": "D",
-                    "F": "E",
-                    "G": "F",
-                    "H": "G",
-                };
+                const cupJPToUS = (cup) => cup === "A" ? "AA" : String.fromCharCode(cup.charCodeAt(0) - 1);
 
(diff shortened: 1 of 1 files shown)


scrapers/erodougazo.yml
```diff
@@ -0,0 +1,133 @@
...
+                // split the string by comma
+                var parts = value.split(",");
+
+                // check if the first part contains a cup size
+                var cupSizeRegex = /[A-Z]+/;
+                var cupSizeMatch = parts[0].match(cupSizeRegex);
+                if (!cupSizeMatch) {
+                  return "";
+                }
+                var cupSize = cupSizeMatch[0];
```
feederbox826, line 48: The logic around the regex matches are a bit questionable, will try to clean it up later

scrapers/erodougazo.yml
```diff
@@ -0,0 +1,133 @@
...
+                // check if the second part contains measurements
+                var measurementsRegex = /(\d+)[^\d]*(\d+)[^\d]*(\d+)/;
+                var measurementsMatch = parts[1].match(measurementsRegex);
+                if (!measurementsMatch) {
+                  return "";
+                }
+                // extract the measurements and convert to inches
+                var bust = Math.round(parseInt(measurementsMatch[1]) / 2.54);
+                var waist = Math.round(parseInt(measurementsMatch[2]) / 2.54);
+                var hips = Math.round(parseInt(measurementsMatch[3]) / 2.54);
```
feederbox826, line 59: There is also a metric to imperial postprocessing script but not sure if that would help much

scrapers/erodougazo.yml
```diff
@@ -0,0 +1,133 @@
...
+                  "F": 6,
+                  "DDDD": 7,
+                  "G": 7,
+                  "H": 8,
+                  "I": 9,
+                  "J": 10,
+                  "K": 11,
+                  "L": 12,
+                  "M": 13,
+                  "N": 14,
```
feederbox826, line 80: This chart cannot be correct because Japan does not have any double cups. IN fact they have a different system starting at A which would be <1in bust, corresponding to AA US, JP E would be 4in bust, or US D Cup

[1](https://thetokyofashionguide.com/learn-japanese-bra-sizes-chart/) [2](https://katherinehamilton.com/pages/international-bra-size-chart) [3](https://www.reddit.com/r/ABraThatFits/wiki/sizing_systems/#wiki_by_country)

feederbox826 (changes requested):
bust size calculations are incorrect! Japan's system is unfortunately much more complicated

scrapers/erodougazo.yml
feederbox826: I'm in the process of writing up a discourse wiki post about it 


https://bustyresources.fandom.com/wiki/Bra_sizing_by_country
tl;dr, one cup size in JP = +1cm, this can lead to up to 3 cup sizes of drift if you convert 1:1. JP also starts at A for 0 bust/band difference which would be US AA

feederbox826: https://discourse.stashapp.cc/t/bra-cup-sizes-and-conversions/1407 the last of my sanity going away in writing

feederbox826 (changes requested):
Still running into some problems 

https://erodougazo.com/actress/av/%E6%A1%9C%E3%82%8A%E3%82%87%E3%81%86%E3%81%8B/
https://erodougazo.com/actress/av/%E6%9C%A8%E6%9D%91%E6%84%9B%E5%BF%83/

 returns NaN for Bust without a cup size and birth date is mangled (Month out of range), will try to have a look myself as well

feederbox826:
I made a *proper* converter for it that derives what their likely US bra size would be https://github.com/feederbox826/CupConverter/blob/main/cupConverter.js, will see if I can fit it into the JS you already have

feederbox826:
```yml
      Measurements:
        # first selector is cup size, second is bust, waist, hips
        # japanese: スリーサイズ = english: measurements
        selector: $p1a/li/a[contains(@href, "/actress/index/cs")]/text()|$p1a/li[i[contains(text(), "スリーサイズ")]]/span
        concat: "," # concat so the cup size and measurements can be processed as one input value
        postProcess:
          - javascript: |
              if (value && value.length) {
                // create super regex to match the entire pattern
                const allRegex = /([A-Z])カップ,B(\d+) \/ W(\d+) \/ H(\d+)/
                // if no matches, return empty string
                const [_, cup, bust, waist, hips] = value.match(allRegex);
                if (!cup || !bust || !waist || !hips) return `failed: ${cup}, ${bust}, ${waist}, ${hips}`;
                // http://feederbox.cc/gh/CupConverter (MIT)
                function jpToUS(cupLetter, bustCm, waistCm, hipCm) {
                  const metricToImp = (metric) => Math.round(metric * 0.393701)
                  const lookupUSCup = (bustDiffIn) => bustDiffIn < 1 ? 'AA' : String.fromCharCode(Math.round(bustDiffIn) + 65 - 1)
                  // 1 letter = +1cm increment
                  // get ASCII code to increment letter, start at 1
                  return `${metricToImp(bustCm)}${lookupUSCup(metricToImp(cupLetter.toUpperCase().charCodeAt() - 65 + 1))}-${metricToImp(waistCm)}-${metricToImp(hipCm)}`
                }
                return jpToUS(cup, bust, waist, hips)
              }
      Birthdate:
```

This works quite well for me but the birth date does not work

feederbox826:
Ah that is my mistake, I thought the japanese measurements were underbust, not overbust. Will also adapt the rounding for US/UK

feederbox826 (approved):

Issue #2199, opened 2025-02-22
TheMovieDB Scraper - Date format issue


C-BoT-AU:
TheMovieDB Scraper seems to be importing the date from the site "As is" when trying to add a movie/scene via URL.

On certain scrapes, it is importing the date and switching the month and day around, adding it into stash as MM-DD-YYYY.
At first I thought it was simply the way that TMDb formatted the date on their site, however I noticed on refreshing a page, that the date changed formats. 
Wondering if it is more complicated and could be due to IP region filtering or cookies/something-smarter-than-I-am and how the site changes the date format to match where the user is viewing the site from.

I managed to get a screenshot of the same movie showing two different date formats.

<img width="242" alt="Image" src="https://github.com/user-attachments/assets/28595b6c-0427-4736-b2d4-d18d0fb58195" />

feederbox826:
> changes the date format to match where the user is viewing the site from.
yes

feederbox826:
their API is pretty open, will see what I can do with a proper client

PR #2328, opened 2025-05-12
fix for getting the video release date TMDB.yml

Druidblack:
according to some links, the scraper received a date format that stash did not perceive as correct

## Examples to test

https://www.themoviedb.org/movie/574475-final-destination-bloodlines
https://www.themoviedb.org/movie/1233069-exterritorial
https://www.themoviedb.org/movie/1197306-a-working-man

## Short description

Describe the general changes
Now the scraper always gets the date in the same stash-compatible format.

before

after

### DIFF
--- a/scrapers/TMDB.yml
+++ b/scrapers/TMDB.yml
@@ -34,7 +34,7 @@ xPathScrapers:
           - replace:
               - regex: \(.*$
                 with:
-          - parseDate: 02/01/2006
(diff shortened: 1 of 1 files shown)


feederbox826:
Nope, this is based on which region you're scraping from https://github.com/stashapp/CommunityScrapers/issues/2199. It was forced to be MDY

PR #2338, opened 2025-05-17
Update AdultPrime.yml

estellaarrieta:
## Examples to test

https://adultprime.com/studios/video/139270?site=goodmorningsex
https://adultprime.com/studios/video/139252?site=primelesbian
https://adult-prime.com/video.php?id=138838

## Short description

Site was updated, this PR fixes the scene scraper


Maista6969:
Looks like they've added a secondary site with worse metadata, I'll add it to the AdultPrime scraper but would still recommend people scrape from their primary site (to avoid cover image edit wars if nothing else)

PR #2355, opened 2025-05-25
AdultDVDMarketPlace CDP Age Prompt Click-through

WellDoneLiz:
## Scraper type(s)
Config update for all xPath scrapers.

## Examples to test

[sceneByURL: Rough Love 3](https://www.adultdvdmarketplace.com/dvd_view_416507.html)

## Short description

Add click on "ENTER" button when scraping.
Quick'n'dirty fix that adds unnecessary overhead, should move to a script in the future that calls "www.adultdvdmarketplace.com/xcart/adult_dvd/disclaimer.php?action=enter" once before scraping.

### DIFF
--- a/scrapers/AdultDvdMarketPlace.yml
+++ b/scrapers/AdultDvdMarketPlace.yml
@@ -70,4 +70,9 @@ xPathScrapers:
       Tags:
         Name: //span[text()="Category:"]/following-sibling::a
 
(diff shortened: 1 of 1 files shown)


feederbox826:
Yeah it sucks :/ Not too keen on rewriting the entire scraper but we'll see

PR #2371, opened 2025-06-01
Update Eporner.yml

Strassboom:
Allows for use of Eporner scraping tool using the tagger tool, or the "Scrape With..." button within an individual scene's page, so long as "EPORNER.COM - [videoidfromurlhere]" are in a given scene's filename.

## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [x] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

List a few links/titles/names/filenames/codes to test

EPORNER.COM - [TM1JDrf9Ykg] Leana Lovings - Nerdy (2160)
EPORNER.COM - [yXPCMxo3gI2] Abbey Rains Nerd Girls - Abbey Rain creampie nerd redhead (1080)
EPORNER.COM - [whatevercodeisinhere] followed by anything else .fileextension

## Short description

Describe the general changes

Eporner.yaml community scraper can now pull all media info for a video so long as the videoid is in the filename i.e. EPORNER.COM - [videoid] etc, as prepended when downloading any content through the website's user interface. This should work in the Edit tab of an individual scene, or in the Tagger section of the Scenes tab when selecting Eporner.

### DIFF
--- a/scrapers/Eporner.yml
+++ b/scrapers/Eporner.yml
@@ -17,7 +17,7 @@ sceneByFragment:
   queryURL: https://www.eporner.com/video-{filename}/ # Scrapes data on video so long as the EPORNER video id is in the filename
   queryURLReplace:
     filename:
-      - regex: '(EPORNER\.COM - \[)(\w+)(\]{1}.*)' # Filename must start with "EPORNER.COM - " followed by the video id in square brackets (see any EPORNER video url for example)
+      - regex: '(\[)(\w+)(\]{1}.*)' # Filename must start with "EPORNER.COM - " followed by the video id in square brackets (see any EPORNER video url for example)
         with: $2
   scraper: singleSceneScrape
 
--- a/scrapers/Eporner.yml
+++ b/scrapers/Eporner.yml
@@ -55,24 +55,24 @@ xPathScrapers:
       Image: //meta[@property="og:image"]/@content
       URL: //meta[@property="og:url"]/@content
     sceneScraper:
-    scene:
-      Title:
-        selector: //*[@id="video-info"]/h1/text()
-      Date:
-        selector: //script[contains(.,"uploadDate")]/text()
-        postProcess:
-          - replace:
-              - regex: .*uploadDate":\s"(\d{4}-\d{2}-\d{2})T.*
-                with: $1
-          - parseDate: 2006-01-02
-      Performers:
-        Name: //*[@id="video-info-tags"]/ul//li[@class='vit-pornstar starw']/a/text()
-      Tags:
-        Name: //*[@id="video-info-tags"]/ul//li[@class='vit-category' or @class='vit-tag']/a/text()
-      Studio:
-        Name: //*[@id="video-info-tags"]/ul/li[@class='vit-uploader']/a/text()
-      Image: //meta[@property="og:image"]/@content
-      URL: //meta[@property="og:url"]/@content
+      scene:
+        Title:
+          selector: //*[@id="video-info"]/h1/text()
+        Date:
+          selector: //script[contains(.,"uploadDate")]/text()
+          postProcess:
+            - replace:
+                - regex: .*uploadDate":\s"(\d{4}-\d{2}-\d{2})T.*
+                  with: $1
+            - parseDate: 2006-01-02
+        Performers:
+          Name: //*[@id="video-info-tags"]/ul//li[@class='vit-pornstar starw']/a/text()
+        Tags:
+          Name: //*[@id="video-info-tags"]/ul//li[@class='vit-category' or @class='vit-tag']/a/text()
+        Studio:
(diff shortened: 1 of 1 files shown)


scrapers/Eporner.yml
```diff
@@ -12,6 +12,14 @@ sceneByQueryFragment:
   action: scrapeXPath
   queryURL: "{url}"
   scraper: sceneScraper
+sceneByFragment:
+  action: scrapeXPath
+  queryURL: https://www.eporner.com/video-{filename}/ # Scrapes data on video so long as the EPORNER video id is in the filename
+  queryURLReplace:
+    filename:
+      - regex: '(EPORNER\.COM - \[)(\w+)(\]{1}.*)' # Filename must start with "EPORNER.COM - " followed by the video id in square brackets (see any EPORNER video url for example)
```
Maista6969: I'd suggest a more lax regex here and just looking for the first value within square brackets: the default filename template for tools like `yt-dlp` for example will be `Title of Video [video id].mp4`

scrapers/Eporner.yml
```diff
@@ -46,6 +54,32 @@ xPathScrapers:
...
+            - parseDate: 2006-01-02
+        Performers:
+          Name: //*[@id="video-info-tags"]/ul//li[@class='vit-pornstar starw']/a/text()
+        Tags:
+          Name: //*[@id="video-info-tags"]/ul//li[@class='vit-category' or @class='vit-tag']/a/text()
+        Studio:
+          Name: //*[@id="video-info-tags"]/ul/li[@class='vit-uploader']/a/text()
+        Image: //meta[@property="og:image"]/@content
+        URL: //meta[@property="og:url"]/@content
+  singleSceneScrape:
```
Maista6969: You've created an extra scraper object here, which isn't necessary: there should really only be one "canonical" scraper for scenes from this site, the only reason that search gets its own scraper is because it usually targets a page structure with multiple entries and needs different logic

scrapers/Eporner.yml
Strassboom: The reason I used EPORNER.COM is because the leading string followed by the id in brackets is the one thing I have not modified across all videos that are downloaded from that site, and ensures one is only scraping videos from there. That said, I'm happy to modify this regex.

scrapers/Eporner.yml
Strassboom: This new scraper object contributes additional functionality that the previous one doesn't. The original scraper object assumes that your video's filename is admissible as a search query string, and that the first video in search results is a match to the video you want metadata for, thus if the video is removed from the website you don't run the risk of getting metadata added from the wrong video. Another reason I'm relying on videoid's over the title, is that some video titles are so long they tend to be cut off when downloading, with one having to sometimes even manually change the file type from being the _title name_ for some reason back to .mp4. One last benefit I can see to this scarper, is that instead of relying on all the URL's to be correct, if a domain change occurs like with MISSAV.com to missav.ws, we can just change the QueryURL value in my new scraper and then add the ID's in. This allows for further future proofing than the sceneByURL scraper, as that one does not account for a potential future domain change. I will likely keep this second scraper around locally for personal use, but if you feel it is cluttering the Eporner.yaml then you're welcome to reject the PR and I can take a shot at working on a scraper for a different site.

Maista6969: Yes, I agree with that but what I was trying to say is that there's a `sceneScraper` on line 38, another one with the same name `sceneScraper` on line 57 and a new scraper called `singleSceneScrape` on line 76: these can all be consolidated into a single scraper, and removing the duplicate one will allow the CI to pass

Strassboom: I'd be down to take a crack at it! I completely missed the duplicate scraper. I'm a bit new so I wasn't aware one scraper could perform multiple scraping types! I'll read the docs a bit more and try to consolidate it down as much as possible. Should we keep this pr open or just close it and I can submit another for a full refactor?

Maista6969: We can just leave it open, there's no problem with a bit of a back-and-forth in the PR or in having multiple commits since they'll all get squashed when we merge anyway 🙂

Strassboom: Sounds good with me 🙂! I'll take a look at rewriting the yaml so one scraper can do the URL, Name, and ID approaches that are currently present in the file.

Strassboom (commented):
Consolidated scraping processes to one scraper as requested.

Strassboom:
Hi there, what is the change being requested? I don't see a comment or link specifying what I can do to satisfy the request.

PR #1893, opened 2024-06-05
Add BellesaPlus API scraper

feederbox826:
- Removed `bellesafilms.com` and `bellesahouse.com` since they redirect to `https://bellesaplus.co/`
- Explicitly removes support for `bellesa.co` since scenes are delayed and paywalled
- Removed old scrape since it was completely nonfunctional with new UI elements

closes #1892 
related: https://github.com/stashapp/CommunityScrapers/issues/1338#issue-1708737874

PR #2388, opened 2025-06-09
remove studio code parsing

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://porncornvr.com/scene/neighborly-beauties-rosalyn-sphinx/

## Short description

The find-replace regex currently finds a code after `/scenes/` in the video source URL. In the scene example above, that part of the URL is `/scenes2/735/` so the regex does not match.

Even if the regex was made to allow for digits after "scenes" like `scenes\d*/`, the following number is 735 in this case and a scene from a slightly earlier date has an ID here of 54136.

The "ID" term here relates to the number(s) in the CDN URL and it appears that the CDN has some scene media in `/scenes/` and newer scenes are in `/scenes2/`. This looks like a CDN architecture reason and nothing to do with the studio itself. For that reason I've now removed the `Code` scraping as there doesn't appear to be a studio code/ID determinable.


nrg101:
actually, the studio code may not simply be "735" for this scene as the most recently scraped studio scene with a code is https://stashdb.org/scenes/2d91573d-86cb-438e-8f8d-70908af9770c with "54136"...

so either the scene IDs got reset, or "scenes2" and "735" somehow gets combined (where scenes 2 is like 53305 or something... even though that's a really janky bad guess) to give 54140 or something

this assumes the studio codes ascend numerically over time and have not "reset" recently

nrg101:
converted to draft for further investigation

nrg101:
https://realjamvr.com/scene/realjamcasting-kayla-kayden/
- scrapes 51399 for studio code
- has URL at SLR: https://www.sexlikereal.com/scenes/realjamcasting-kayla-kayden-51399

that is the newest scene listed at SLR, and was published on realjamvr on Feb. 4, 2025, so SLR is several months behind (possibly that is an exclusivity stipulation or something of the studio before redistributing to partners like SLR)

feederbox826:
Maybe they have two seperate code conventions for redistribution like bellesa does? https://github.com/stashapp/CommunityScrapers/pull/1893

feederbox826:
So just remove studio code parsing?

PR #2390, opened 2025-06-12
Update vrpornnow.com

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://vrpornnow.com/videos/126-match-starring-karina-king/ --> note that this scene has the wrong cover image shown on the page for a different scene, this scraper improvement corrects that and picks the correct cover
- https://vrpornnow.com/videos/101-the-pleasure-retreat-starring-novella-night/
- https://vrpornnow.com/videos/dreams-come-true-cgi-passthrough/ --> this is a scene in the CGI sub-studio

## Short description

These changes includes:
- picking the correct scene image cover even if the web page shows a cover from a different scene (probably a niche case, but example above)
- adds Studio Code
- adds fixed tag `Virtual Reality` as well as the dynamically scraped tags
- adds current scene URL prefix
- picks URL from meta
- adds logic to determine studio (for CGI scenes, by presence of tag)

### DIFF
--- a/scrapers/VRpornnow.yml
+++ b/scrapers/VRpornnow.yml
@@ -35,25 +35,14 @@ xPathScrapers:
       Code:
         selector: *urlSel
         postProcess:
-          - replace:
+          - replace: &pickCode
               - regex: .*/videos/(\d+).*
                 with: $1
       Image:
-        selector: //script[@type="application/ld+json"]
+        selector: *urlSel
         postProcess:
-          - javascript: |
-              if (value && value.length) {
-                // parse JSON
-                const obj = JSON.parse(value);
-                // find WebPage item
-                const webPageItem = obj["@graph"].find(item => item["@type"] === "WebPage");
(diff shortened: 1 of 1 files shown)


nrg101:
change Details selector to get from meta tag instead

the current one scraping the p tag deletes newlines so that:

> It´s one of those boring weekends. You are home alone but wait what is this?                                                                                                      

> Ohana Petite is walking inside to return the favour from the last time you visitied her at the wellness spa. Can you make her cum?

becomes

> It´s one of those boring weekends. You are home alone but wait what is this?Ohana Petite is walking inside to return the favour from the last time you visitied her at the wellness spa. Can you make her cum?

note the newline removal after the "?" resulting in a bad scrape.

The meta tag does not have this newline, and matches the presentation in the web page's Description section

nrg101:
converting to draft while I investigate some scenes where the image URL assumptions are wrong, and also scraping the CGI sub-studio

PR #2398, opened 2025-06-18
add birthdate for ladyboygold performerByURL

nrg101:
## Scraper type(s)
- [x] performerByURL

## Examples to test

- https://www.ladyboygold.com/index.php?section=1832&actorid=2033 (Age: 24, Birthday: November 5) should scrape 2000-11-05
- https://tsraw.com/model/aline-tavares (Age: 27, Birthday: May 20) should scrape 1998-05-20

## Short description

This reuses the same Age + Month,DateOfMonth logic that was added for TSRaw. That code actually had a bug where it was adding the `extraYear` (for scenarios where the birthday is later in the current year), but it should actually be subtracted. That bug is fixed now.

For some reason I can't scrape a performer at tsraw.com, I think it might be some sort of JS lazily-loaded page as the `debug.printHTML: true` option doesn't show the same content as I can see in a browser even with:

```yaml
debug:
  printHTML: true

driver:
  useCDP: true
  cookies:
    - CookieURL: "https://tsraw.com"
      Cookies:
        - Name: consent
          Domain: tsraw.com
          Value: "true"
          Path: /
```

This could just be a "me problem" with geo/age blocking.

### DIFF
--- a/scrapers/LadyboyGold/LadyboyGold.yml
+++ b/scrapers/LadyboyGold/LadyboyGold.yml
@@ -39,6 +39,7 @@ sceneByQueryFragment:
 xPathScrapers:
   performerScraperLadyboyGold:
     common:
+      $modelStatsListItem: //div[@class="modelStats"]/ul/li
       $infoPiece: //div[@class="info-box info"]
     performer:
       Name:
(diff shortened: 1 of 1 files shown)


feederbox826:
It is a JS lazyloaded page, the CDP is required but it is a bit JS-scrapeable but still painful

feederbox826:
the TSRAW link provided Just Works for me?

feederbox826 (approved):

PR #2421, opened 2025-07-09
fix groobyvr: add alternate selector for details

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://www.groobyvr.com/tour/trailers/Amanda-Riley-In-A-Hot-Sticky-Mess.html

## Short description

Either the scene description has changed or some scenes use a different CSS layout. Whatever the case, this additional selector can now scrape the description.

I have to use CDP configured with a VPN in order to access the site as it has region-based age-verification. I set `driver.useCDP` to `false` however as this issue may only affect a few countries.

### DIFF
--- a/scrapers/GroobyNetwork-Partial.yml
+++ b/scrapers/GroobyNetwork-Partial.yml
@@ -271,7 +271,10 @@ xPathScrapers:
         postProcess:
           - parseDate: January 2, 2006
       Details: &detailsVR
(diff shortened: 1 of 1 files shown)


feederbox826:
https://github.com/stashapp/CommunityScrapers/commit/15b33ff8518bcab3b9f6479a47618397827fb059

no need for alternate selector and it targets the description box. went for //* for simplicity

tested against https://www.groobyvr.com/tour/trailers/Antonella-Spread-Wide.html as it also has nested spans

PR #2433, opened 2025-07-16
Update Tokyohot.py

finabit:
Add dynamic imports, normalize naming to real title, handle multi-part, use stable filename instead of title, add translation and translation cache of japanese tags, add some error handling

### DIFF
--- a/scrapers/Tokyohot/Tokyohot.py
+++ b/scrapers/Tokyohot/Tokyohot.py
@@ -3,14 +3,57 @@ import json
 import sys
 import os
 import re
+import subprocess
+import importlib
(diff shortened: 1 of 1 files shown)


feederbox826:
Thanks but your contributions don't exactly align with the scraper goals

- https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/py_common/deps.py
- translations are not preferred, original source content is

ghost:
In most cases that would make sense, except for the fact that on the target website the specific tags that are being translated are ONLY available in Japanese, irrespective of what language you select.

PR #2434, opened 2025-07-17
move groobyvr to own scraper to allow scene search

nrg101:
## Scraper type(s)
- [x] performerByURL
- [x] sceneByName
- [x] sceneByQueryFragment
- [x] sceneByURL
- [x] galleryByURL

### sceneByURL

- https://www.groobyvr.com/tour/trailers/Dicking-Down-Gigi-Ravine.html
- https://www.groobyvr.com/tour/trailers/A-Katt-In-Your-Lap.html

### sceneByName

- katt
- lap

Note that when the search results have more than 1 scene, the list of performers is repeated for each scene. I'm not sure if there is a way around this? This only affects the searchByName results list, and the searchByQueryFragment just scrapes the individual scene page normally.

### galleryByURL

this uses scene URLs, so the sceneByURLs above

### performerByURL

- https://www.groobyvr.com/tour/models/Cailey-Katts.html

## Short description

This adds a scene search for groobyvr.com. As a scraper file appears to only support a single sceneByName queryURL (and seems to have no per-domain ability), it made sense to move this scraper out of the (rather messy) GroobyNetwork-Partial scraper into its own one.

### DIFF
--- a/scrapers/GroobyVR.yml
+++ b/scrapers/GroobyVR.yml
@@ -1,0 +1,119 @@
+# yaml-language-server: $schema=../validator/scraper.schema.json
+name: GroobyVR
+performerByURL:
+  - action: scrapeXPath
+    url:
+      - groobyvr.com/tour/models
+    scraper: performerScraper
+sceneByURL:
+  - action: scrapeXPath
+    url: &urls
+      - groobyvr.com
+    scraper: sceneScraper
+sceneByName:
+  action: scrapeXPath
+  queryURL: https://www.groobyvr.com/tour/search.php?qall={}
+  scraper: sceneSearchResultsScraper
+sceneByQueryFragment:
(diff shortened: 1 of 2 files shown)


feederbox826:
Probably for the best since we had so many domain overrides

scrapers/GroobyVR.yml
```diff
@@ -0,0 +1,119 @@
...
+      Gender:
+        fixed: transgender_female
+      Image:
+        selector: //div[@class="model_photo"]/img/@src
+        postProcess:
+          - replace:
+              - regex: ^(/.*)
+                with: https://www.groobyvr.com$1
+              - regex: 2x
+                with: 3x
```
Maista6969, line 38: Replace `-2x` with `-full` instead to get the unscaled image in its original resolution

PR #2443, opened 2025-07-23
Add Nikko Jordan scraping

robbydurano:
Should be pretty easy merge.  Tested against https://nikkojordan.com/scene/8306715/the-anal-adventures-of-nikko-pearl


Maista6969:
You're right, that was easy to merge 🙂

PR #2459, opened 2025-08-07
GroobyVR: pick largest image

nrg101:
## Scraper type(s)
- [x] performerByURL
- [x] sceneByURL

## Examples to test

- https://www.groobyvr.com/tour/trailers/Getting-Into-Lolas-Jeans.html
- https://www.groobyvr.com/tour/trailers/Waking-Up-With-Ariel.html
- https://www.groobyvr.com/tour/trailers/Embers-Long-Stretch.html
- https://www.groobyvr.com/tour/trailers/I-Fucked-Gabrielly-Ferraz.html

## Short description

Someone recently pointed out that the "-1x" (or "-2x" or whatever number x) in the image URL can be changed to "-full" to get the largest version. This change does that

### DIFF
--- a/scrapers/GroobyVR.yml
+++ b/scrapers/GroobyVR.yml
@@ -121,5 +121,5 @@ xPathScrapers:
       
 # Use CDP with a configured proxy to bypass region-based age-verification
 driver:
-  useCDP: true
+  useCDP: false
 # Last Updated August 7, 2025
--- a/scrapers/GroobyVR.yml
+++ b/scrapers/GroobyVR.yml
@@ -36,11 +36,14 @@ xPathScrapers:
               - regex: content//
                 with: content/
               # ensure URLs are converted from relative to absolute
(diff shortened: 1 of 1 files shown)


nrg101:
There are some edge cases where the image is not being scraped at all (seems like the xpath is not matching). Putting into draft until resolved.

nrg101:
Brought out of draft for review

scrapers/GroobyVR.yml
```diff
@@ -30,13 +30,20 @@ xPathScrapers:
         fixed: transgender_female
       Image:
         selector: //div[@class="model_photo"]/img/@src
-        postProcess:
+        postProcess: &imagePostProcess
           - replace:
-              - regex: ^(/.*)
-                with: https://www.groobyvr.com$1
+              # errant double slash
```
feederbox826, line 35: // should be handled by most webservers, but worth catching

PR #2469, opened 2025-08-15
Add Frenchtwinks studio code and fix details

stash-of-awesomeness:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

* https://www.french-twinks.com/en/gay-porn-videos/in-the-backstage-01 (fta1080)
* https://www.french-twinks.com/en/gay-porn-videos/paul-and-gabriel-break-the-routine (fta984)

## Short description

This fixes an issue that existed when scraping the details where linebreaks were lost and the subtitle appeared smashed at the start of the description. This changes it to properly pull line breaks so the details match what is displayed visually on the site.

This also adds studio codes for French Twinks. These can be verified by looking at the `contentUrl` in the JSON-LD for the `VideoObject`.

### DIFF
--- a/scrapers/Frenchtwinks.yml
+++ b/scrapers/Frenchtwinks.yml
@@ -15,7 +15,7 @@ xPathScrapers:
             - regex: .*\/(\d+)\/\d+\.jpg
               with: "fta$1"
       Details:
-        selector: //div[@class="vid_txt_block2 vid_txt_block2017"]/article/h2 | //div[@class="vid_txt_block2 vid_txt_block2017"]/article/p//text() | //div[@class="vid_txt_block2 vid_txt_block2017"]/article/p//br
+        selector: //div[@class="vid_txt_block2 vid_txt_block2017"]/article/*/text()
         concat: "\n\n"
       Date:
         selector: //time[@datetime]


feederbox826:
why are you pulling the `<br>`s as well? you're already concat-ing with newlines

stash-of-awesomeness:
This needs to pull the `<br />` separated chunks in addition to the text node provided by the `<p>` because of how the XPath scraper breaks up elements which have sub-elements. XPath only returns the first text node with `text()` which ends at the first sub-element. By pulling in the `br` elements as well, we are able to get the additional lines which are split up within the `<p>` tag. It also pulls the `<h2>` tag because that is outside of the `<p>` but inside of the `<article>` that was being parsed before.

Ideally they wouldn't have use `<br />` within `<p>` tags considering the whole point of the tag is to separate out paragraphs, but alas here we are.

feederbox826:
found it - you just messed up your xpath with `//text()`

Issue #1967, opened 2024-08-09
Improve Model searching for Pornhub


Joly0:
Hey, i have noticed, that various Models on Pornhub are not found. This is due to the fact, that the original Pornhub.yml file uses this endpoint to search for models "https://www.pornhub.org/pornstars/search?search={}".
This doesnt show all Models, just ones that are listed as pornstars. There are also users, who are uploading stuff, but not listedd as pornstars. By trying out, i found "Lana Bee" for example. 

Though i have found out, that this endpoint gives back correctly the user "https://www.pornhub.org/user/search?username={}"

So there needs some improvement. I thought i could add this myself by changing the first few lines to this:
```
name: Pornhub
performerByName:
  - action: scrapeXPath
    queryURL: https://www.pornhub.com/pornstars/search?search={}
    scraper: performerSearch
  - action: scrapeXPath
    queryURL: https://www.pornhub.org/user/search?username={}
    scraper: modelSearch
```
 but apparentely, the scraper doesnt handle lists very well (it doesnt at all actually, just throws a bunch of unmarshal errors). So there needs another way to add the second one

Edit: I have opened this as an enhancement, but it might also be a bug. I am not sure

 The search should look something like this:
 ```
   modelSearch:
    performer:
      Name: //span[@class="usernameBadgesWrapper"]/a[@class="usernameLink"]/text()
      URL:
        selector: //span[@class="usernameBadgesWrapper"]/a[@class="usernameLink"]/@href
        postProcess:
          - replace:
              - regex: ^
                with: "https://www.pornhub.org"
 ```
 Though i am unable to get this to work, as "performerByName" only accepts one action, queryURL and scraper

Joly0:
Also here it would be good to change the behaviour of scraping scenes on pornhub. If i search a scene with a user, rather than a performer, the user is added as a studio, rather than a performer. This is wrong

Maista6969:
> but apparentely, the scraper doesnt handle lists very well (it doesnt at all actually, just throws a bunch of unmarshal errors). So there needs another way to add the second one

Scrapers currently only support one search per type, so we'll have to choose between `pornstar` and `model`: personally I lean towards `model` since that's _probably_ what users go to Pornhub for and more mainstream pornstars can be scraped from other sources

> Also here it would be good to change the behaviour of scraping scenes on pornhub. If i search a scene with a user, rather than a performer, the user is added as a studio, rather than a performer. This is wrong

[...]

PR #2475, opened 2025-08-26
[Pornhub] Find amateur channels by exact name

you-cant-see-me:

## Scraper type(s)
- [x] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

"carolina jackson"
"blake core"

## Short description

The current scraper uses `/pornstars/search?search=`; this only returns registered pornstars and no amateur channels. Using `/video/search?search=`, when supplied with the exact (case insensitive) name of the channel, finds amateur channels that the current method does not find, while still also finding registered pornstars by exact name.

Example of the issue: rip videos with a yt-dlp wrapper api, scrape scene by fragment which prompts to create a new performer for that channel with just the name, now you go to scrape the performer by name and nothing of relevance shows up. So you need to open up the video, go to file info, click the pornhub url, go to the channel, copy its link, put it into stash, then scrape performer by url.

With this change, these channels can be scraped easily. The scene scrape already filled in the exact name of the performer, so scraping the performer is a 2 click process.

The big downside is that `/video/search?search=` only returns 1 channel result and only by exact letter-by-letter match; this would mean simply searching for the performer by hand would work a lot worse.

Personally, this tradeoff is worth it for me. But I imagine it may not be the case for others. Regardless, I thought to make a PR and leave it as draft so others can use it or maybe if the tradeoff is worth for the majority of people, even merged.

Sidenote: the added `&o=mr` orders results by most recent; for this we of course dont care about video results, but ordering by most recent takes less processing time to return a response compared to most relevant which is the default.

### DIFF
--- a/scrapers/Pornhub.yml
+++ b/scrapers/Pornhub.yml
@@ -1,7 +1,7 @@
 name: Pornhub
 performerByName:
   action: scrapeXPath
-  queryURL: https://www.pornhub.com/pornstars/search?search={}
+  queryURL: https://www.pornhub.com/video/search?search={}&o=mr
   scraper: performerSearch
 performerByURL:
   - action: scrapeXPath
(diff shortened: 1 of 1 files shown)


feederbox826:
https://github.com/stashapp/CommunityScrapers/issues/1967 PornHubmodel was designed to search by model, that might align more with what you want?

you-cant-see-me:
i tried PornHubmodel and it gave no results for the channels i was looking for. im guessing "pornstars" are industry wide / professional grade, "models" are pornhub's own pornstars that while amateurs have some kind of contract with pornhub (not sure tho), and "channels"/"users" are anyone uploading videos to pornhub. so really it would need to be PornHubUser or PornHubChannel? i havent checked if /users/search returns something useful

feederbox826:
The new search uses the built-in search api but also returns all the results (models, pornstars), hopefully the best of all worlds, the one downside being that it needs a script to fetch PH token

PR #2478, opened 2025-08-30
Add Helix Studios to the AdultEmpireCash scraper

stash-of-awesomeness:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL
- [x] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

* Scenes
  * https://www.helixstudios.com/1725721/helix-studios-testing-the-tutor-streaming-scene-video.html
* Movies
  * https://www.helixstudios.com/4826020/twink-confessions-3-porn-videos.html

## Short description

It looks like they recently switched from their own system to using AdultEmpireCash. I tested it out locally and the scraper appears to work, though there are two issues that I noticed so far:

1. There are new scene covers for all of the scenes
2. Formatting appears to have been stripped during the migration, so paragraphs are no longer separated

I have also confirmed that the movie scraper works with the new Helix Studios website. Unfortunately with the migration comes new problems:

1. The synopsis appears to have been migrated, but there is some garbage showing up ahead of it in the scraped description.
2. The release dates are all wrong. Movies released in 2017 are showing up as 2024.

### DIFF
--- a/scrapers/AdultEmpireCash.yml
+++ b/scrapers/AdultEmpireCash.yml
@@ -48,6 +48,7 @@ movieByURL:
       - elegantangel.com/
       - filthykings.com/
       - forbiddenfruitsfilms.com/
+      - helixstudios.com/
       - lewood.com/
       - rodneymoorestore.com
       - severesexfilms.com/


feederbox826:
> The release dates are all wrong. Movies released in 2017 are showing up as 2024.

this sounds suspiciously like algolia lastUpdated, probably pulling the wrong value

stash-of-awesomeness:
> this sounds suspiciously like algolia lastUpdated, probably pulling the wrong value

While I would tend to agree, I think in this case it's an issue resulting from the migration. The scenes attached to the movies (but not necessarily the standalone scenes) all have a new release date as well.

It's also useful to note that the old Helix Studios scraper for movies didn't scrape the release date, so this isn't _losing functionality_ at least.

PR #2482, opened 2025-09-02
Improve Algolia API

nrg101:
## Scraper type(s)
- [x] performerByName
- [x] performerByFragment
- [x] performerByURL
- [x] sceneByName
- [x] sceneByQueryFragment
- [x] sceneByFragment
- [x] sceneByURL
- [x] galleryByFragment
- [x] galleryByURL

## Examples to test

This now logs out the scene number in a movie

- https://www.evilangel.com/en/video/evilangel/TS-ANDYLYNN-PAYNE-Seduces-EMMA-ROSIE/283513
- https://www.tsfactor.com/en/video/evilangel/Grazyeli--Shakira-TS-Glam-Girls-Date/282633

One of the performers has a trailing space in the API which causes a mismatch, this is now fixed.

- https://www.evilangel.com/en/photo/Trans-Active-31/145312
- https://www.tsfactor.com/en/video/evilangel/Grazyeli--Shakira-TS-Glam-Girls-Date/282633

The logged gallery photo set cover image URL can now be viewed directly in a browser, it currently spawns a file download which is less convenient as the cover image has to be downloaded as a file and then opened.

- https://www.evilangel.com/en/photo/Trans-Active-31/145312

## Short description

Some improvements to the AlgoliaAPI scraper and therefore any scraper that makes use of it:

- a scraped scene that belongs to a movie will now log out its scene number with the movie title in Algolia style
- performer names in scenes and photo sets are now stripped of leading and trailing whitespace (performer names in performer scraping are already stripped of leading/trailing whitespace)
- the cover image of a photo set (gallery) is now logged as a URL that is directly viewable in a browser

### DIFF
--- a/scrapers/AlgoliaAPI/AlgoliaAPI.py
+++ b/scrapers/AlgoliaAPI/AlgoliaAPI.py
@@ -29,6 +29,7 @@ T = TypeVar('T')
 CONFIG_FILE = 'AlgoliaAPI.ini'
 FIXED_USER_AGENT = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:79.0) Gecko/20100101 Firefox/79.0'
 IMAGE_CDN = "https://images03-fame.gammacdn.com"
+TRANSFORM_IMAGE_CDN = "https://transform.gammacdn.com"
 
 
 def slugify(text: str) -> str:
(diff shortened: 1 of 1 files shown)


feederbox826:
Spawning a file download is just content-disposition but if we can get the same quality, the url seems to be more reliable

PR #2496, opened 2025-09-18
Support sfw urls CorbinFisher.yml

gimmeliina:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [x] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Short description
All /tour/ urls seems to be blocked.
Just replacing it with sfw seems to be working for now, so added it to supported urls

### DIFF
--- a/scrapers/CorbinFisher.yml
+++ b/scrapers/CorbinFisher.yml
@@ -3,7 +3,7 @@ sceneByURL:
   - action: scrapeXPath
     url:
       - corbinfisher.com/sfw/trailers
-      - corbinfisher.com/tour/sfw/trailers
+      - corbinfisher.com/tour/trailers
     scraper: sceneScraper
 
 performerByURL:


feederbox826:
I do remember this being brought up before, sfw vs nsfw previews so will have to defer judgement to MOT

but it looks like it's not just geo-block, the entire site is redirected :/

gimmeliina:
Yeah, made this as temporary workaround. Some parts seems to be broken in the site entirely.

Tried through VPN and no luck, can't get to old site, don't know if sfw version is going to be only non-member option to access the site, and didn't dig out is there is way to figure out original cover and retrieve that in some way.

feederbox826:
https://discourse.stashapp.cc/t/sfw-vs-nsfw-covers/3712

I think in the interim, I'll merge the changes to support metadata for tour but explicitly exclude covers, since that's the only part that differs

PR #2497, opened 2025-09-19
[Babepedia] Fix performerByName search with Python script

TheStashMaster:
Wrote small python script which requires cloudscraper to all Performer Search by Name.

## Scraper type(s)
- [X] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Short description

Performer search by name appeared to be broken. I wrote some small Python script to use the php search to retrieve the JSON data and parse for search results.

### DIFF
--- a/scrapers/Babepedia/Babepedia.py
+++ b/scrapers/Babepedia/Babepedia.py
@@ -1,0 +1,40 @@
+import json
+import re
+import os
+import sys
+import cloudscraper
+import urllib
+
+from py_common import log
+from py_common.util import scraper_args
+
+scraper = cloudscraper.create_scraper()
+
+def performerName(_name: str, _site: str):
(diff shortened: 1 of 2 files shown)


feederbox826:
I think you're double calling babpedia in the args. The UA was also excluded sicne babpedia requires up-to-date UA, and the dependency structure for it is really weird

TheStashMaster:
I'm afraid I'm not following you. Are you  pointing out the POST request and then later scraping the resulting page?

feederbox826:
Was partially due to a misunderstanding on my part, the extra babepedia is not necessary

The User-Agent is excluded in the .yml since it overrides the stash one, and it's necessary to keep it up to date frequently

feederbox826:
Superceded by https://github.com/stashapp/CommunityScrapers/commit/8682bf923ced9cfce59ebbd69568282ab63c5f05, cloudscraper is not required for api search

PR #2506, opened 2025-09-23
Performer now scraping from both RealJamVR and PornCornVR

TheStashMaster:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [X] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

[List a few links/titles/names/filenames/codes to test]

https://realjamvr.com/scene/threesome-anal-teacher-for-stepsister/
https://realjamvr.com/scene/slutty-housewife-bunny-madison/

https://porncornvr.com/scene/latex-doll-ember-snow/
https://porncornvr.com/scene/threesome-passion-girls-prefer-anal-2/

## Short description

I don't love this pull request. Nearly everything works for both RealJam and PornCorn with two exceptions which is why there are two scrapers. The URL extraction hard codes the base URL. This is really only used if you use Scene Search By Name.  As far as I can tell I can't have two "search by name" items listed in the same Yaml file (one for PornCorn one for RealJam) created in the drop down even though I'd use the same Python code.  I really feel like this could be better consolidated but I suspect it's my inexperience that's creating the issue. Help would be appreciated.

### DIFF
--- a/scrapers/RealjamVR/RealJamVR.yml
+++ b/scrapers/RealjamVR/RealJamVR.yml
@@ -14,11 +14,11 @@ sceneByURL: &byURL
   - action: scrapeXPath
     url:
       - realjamvr.com/scene/
-    scraper: sceneScraperRealJamVr
+    scraper: sceneScraper
   - action: scrapeXPath
     url:
       - porncornvr.com/scene
-    scraper: sceneScraperPornCornVr
+    scraper: sceneScraper
 galleryByURL: *byURL
 performerByURL:
   - action: scrapeXPath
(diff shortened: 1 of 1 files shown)


feederbox826:
Oh. Oh I see. 

Personally, I would drop the URL altogether but if I throw myself at this brick wall long enough surely I can come up with a solution

feederbox826 (approved):
Just don't look too hard and don't think about it

PR #2519, opened 2025-09-30
Czech vr scene by name

TheStashMaster:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [X] sceneByName
- [X] sceneByQueryFragment
- [ ] sceneByFragment
- [X] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

- Brand New Threesome
- 267
- New Maid is Horny

## Short description

I've added sceneByName and updated sceneByUrl to extract the URL and the Code. I used the three digit code found in the scene title rather than the longer code in the URL, not sure which would be preferred.

### DIFF
--- a/scrapers/CzechVR.yml
+++ b/scrapers/CzechVR.yml
@@ -110,7 +110,6 @@ xPathScrapers:
       FrontImage: *image
 
 driver:
-  useCDP: true
   cookies:
     - CookieURL: "https://www.czechvr.com"
       Cookies:


feederbox826:
Thanks! I wasn't getting errors without cdp so curious why it was added

PR #2521, opened 2025-10-02
Improve CarnalPlus scraper

gimmeliina:
This PR
- Scrapes 1920px cover from both, studio and network site
- Removes custom title handling from studio site. This is to standardize titles between studio and network site.
- Dig performer name from image alt so no character name is used.
- Ensures that performer is scraped more often


feederbox826:
LGTM, however the cover seems to be limited on backend at 1500w so why stop at 1920 😁

PR #2539, opened 2025-10-16
[fandom.com] Fandom performer scraper

spaceyuck:
## Scraper type(s)
- [x] performerByURL

## Examples to test

https://metalgear.fandom.com/wiki/Solid_Snake
https://witcher.fandom.com/wiki/Yennefer_of_Vengerberg
https://cyberpunk.fandom.com/wiki/Rogue_Amendiares (birthdate without year, death only year)
https://rainbowsix.fandom.com/wiki/Ash_(Siege)
https://nier.fandom.com/wiki/YoRHa_No.2_Type_B (height, weight, meassurements as split field)
https://deadoralive.fandom.com/wiki/Christie (height in mutliple units, weight in mutliple units, measurements)
https://cyberpunk.fandom.com/wiki/Judy_%C3%81lvarez (birthday in regular format)
https://hitman.fandom.com/wiki/Diana_Burnwood (birthday in alternate order)
https://farcry.fandom.com/wiki/Faith_Seed (Birth and death year only)

## Short description

Another one of those scrapers maybe five people are looking for, but I made it, so might as well contribute it.

It should match all fandom.com wiki subdomains, and tries to extract as much performer information as possible. It will match localized versions too, but results may be impacted by the language change.

The results vary heavily between Wikis, since each can use different content structures and IDs, but this seem to work reliably for at least the basics like Name, Image, Tags. The birthdate is very messy, since many give partial infos, add countries, or even more specific locations or other stuff.

It does some extra work trying to wrangle the 1001 variants for height, weight and measurements, but it's a best effort kinda thing, sometimes it works, sometimes it doesn't.

### DIFF
--- a/scrapers/fandom-wiki.yml
+++ b/scrapers/fandom-wiki.yml
@@ -12,33 +12,13 @@ xPathScrapers:
       $infobox: //aside[contains(@class, 'portable-infobox')]
     performer:
       Name: 
-        selector: $infobox/h2//text()
-        postProcess:
-        - replace:
-          # remove citation markers like "[3]"
-          - regex: "\\[[ ]*[0-9]+[ ]*\\]"
-            with: ""
+        selector: $infobox/h2//text()[1]
       Aliases:
-        selector: $infobox//div[@data-source="full_name" or @data-source="full" or @data-source="also_known_as" or @data-source="known" or @data-source="aka" or contains(@data-source,"Name") or contains(@data-source,"name") or contains(@data-source,"alias") or contains(@data-source,"Alias")]/div/text()
-        postProcess:
-        - replace:
-          # remove citation markers like "[3]"
-          - regex: "\\[[ ]*[0-9]+[ ]*\\]"
-            with: ""
+        selector: $infobox//div[@data-source="full_name" or @data-source="full" or @data-source="also_known_as" or @data-source="known" or @data-source="aka" or contains(@data-source,"Name") or contains(@data-source,"name") or contains(@data-source,"alias") or contains(@data-source,"Alias")]/div/text()[1]
       Gender:
-        selector:  $infobox//div[@data-source="gender" or @data-source="Gender" or @data-source="sex"]/div//text()
-        concat: " "
-        postProcess:
-        - replace:
-          # remove citation markers like "[3]"
-          - regex: "\\[[ ]*[0-9]+[ ]*\\]"
-            with: ""
-          # reduce to single word
-          - regex: ^([^\s]+).*
-            with: "$1"
+        selector:  $infobox//div[@data-source="gender" or @data-source="Gender" or @data-source="sex"]/div//text()[1]
       Birthdate:
-        selector: $infobox//div[contains(@data-source,"Birth") or contains(@data-source,"birth") or @data-source="born" or @data-source="dob" or @data-source="DOB"]/div//text()
-        concat: " "
+        selector: $infobox//div[contains(@data-source,"Birth") or contains(@data-source,"birth") or @data-source="born" or @data-source="dob" or @data-source="DOB"]/div//text()[1]
         postProcess: &postprocessDate
         # there are many formats used, many only give partial info - some examples
         # September 26, 1981
@@ -48,9 +28,6 @@ xPathScrapers:
         # 2 May 1998 (aged 46/47)
         # 1997 [3] some more text
         - replace:
-          # remove citation markers like "[3]"
-          - regex: "\\[[ ]*[0-9]+[ ]*\\]"
-            with: ""
           # remove expressions in braces (likely only giving context)
           - regex: \([^)]*\)
             with: ""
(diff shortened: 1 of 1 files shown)


feederbox826:
Hm will take some time to look at this, I find fandom wikis to be extremely inconsistent so would prefer to try to hook into their backends or the raw api instead of parsing customized character boxes

spaceyuck:
I think at it's core Fandom Wikis are just MediaWikis like Wikipedia. You can get the source by appending `?action=raw` to the URL, but it's not much more structured than HTML, just the same IDs and values. Example: https://rainbowsix.fandom.com/wiki/Ash_(Siege)?action=raw

feederbox826:
Yes, I run mediawiki sites, but fandom has their own messed up structure

scrapers/fandom-wiki.yml
```diff
@@ -0,0 +1,91 @@
...
+xPathScrapers:
+  performerScraper:
+    common:
+      $infobox: //aside[contains(@class, 'portable-infobox')]
+    performer:
+      Name: 
+        selector: $infobox/h2//text()
+        postProcess:
+        - replace:
+          - regex: "[[0-9]+]"
```
feederbox826: what is the intention of this?

Is it to remove all numbers from names? If so, would strip

https://nier.fandom.com/wiki/YoRHa_No.2_Type_B
https://scp-db.fandom.com/wiki/SCP-999

scrapers/fandom-wiki.yml
```diff
@@ -0,0 +1,91 @@
...
+            with: "$1$2"
+          - regex: "\\(.*\\)"
+            with: ""
+          - regex: "[,]"
+            with: ""
+          - regex: "([0-9])(?:st|nd|th)"
+            with: "$1"
+          - regex: (\w+:)(\s)(\w+)(\s)(\d+)(\w+)(\s)(\w+)(\s)(\w+)(\s)(\d+)
+            with: $5 $10 $12
+        - parseDate: January 2 2006
```
feederbox826, line 52: this seems needlessly complicated but I guess it works

feederbox826 (changes requested):
some really wacky regex hacks but fandom seems to follow a pretty consistent format

spaceyuck:
I'm turning this into a draft, I need to reverse-engineer half the regexpes I wrote months ago, because I can't tell you anymore what examples they are there for. Lesson learned.

scrapers/fandom-wiki.yml
spaceyuck: Figured it out, I think I encountered some wikis where there were citation links with a text like "[3]", thats to strip these

scrapers/fandom-wiki.yml
spaceyuck: New version has a lot more comments. 

No idea what that last one was supposed to catch, did not encounter it again, so I removed it. The new patterns should be more flexible to pick up sane date variants.

Theres still a lot of garbage slipping through, but the birthdate really  is one of the worst fields in terms of extra content, partial information, and random order.

spaceyuck:
I've added a bunch of comments to the regexps, improved some things, and added some more examples with specific field infos to the PR (because meassurements are very rare, for example).

scrapers/fandom-wiki.yml
```diff
@@ -156,31 +100,18 @@ xPathScrapers:
...
-              console.log('weight has value');
-              const match = /^(\d+)lb$/.exec(value);
-              if (match) {
-                console.log('wight lbs regexp match');
-                const imperialWight = parseInt(match[1]);
-                return (imperialWight * 2.20462).toFixed(0);
-              }
-            }
-            return value;
+        - lbToKg: true
```
spaceyuck, line 110: this was done in JS explicitely because the content might be metric or imperial. The RegExps before were set up in such a way that imperial got reduces to number+unit, while metreic was reduced to just the number, for the Javascript rule to detect imperial and only convert it if imperial.

This makes the result wrong for a lot of wikis (that provide both units)

scrapers/fandom-wiki.yml
```diff
@@ -156,31 +100,18 @@ xPathScrapers:
...
       Measurements:
         # looks for one field with all values, or split fields (get combined into single string)
         # split fields really only work corretly if the field order is still B-W-H
-        selector: $infobox//div[@data-source="measurements" or @data-source="bust" or @data-source="waist" or @data-source="hips"]/div//text()
+        selector: $infobox//div[@data-source="measurements" or @data-source="bust" or @data-source="waist" or @data-source="hips"]/div//text()[1]
```
spaceyuck, line 114: This breaks at least the parsing of metrics for Wikis where the meassurements are 3 separate fields.

scrapers/fandom-wiki.yml
```diff
@@ -133,21 +89,9 @@ xPathScrapers:
...
-              const match = /^(\d)'(\d)$/.exec(value);
-              if (match) {
-                const feet = parseInt(match[1]);
-                const inches = parseInt(match[2]);
-                const imperialHeight = feet + inches / 12.0;
-                return (imperialHeight * 30.48).toFixed(0);
-              }
-            }
-            return value;
+        - feetToCm: True
```
spaceyuck, line 92: Same as weight, might be either metric or imperial, hence the JS solution.

scrapers/fandom-wiki.yml
```diff
@@ -48,9 +28,6 @@ xPathScrapers:
         # 2 May 1998 (aged 46/47)
         # 1997 [3] some more text
         - replace:
-          # remove citation markers like "[3]"
-          - regex: "\\[[ ]*[0-9]+[ ]*\\]"
-            with: ""
```
spaceyuck: I'm pretty sure this breaks examples that use day and month on one line and year on the next (those definitely exist)

scrapers/fandom-wiki.yml
feederbox826: > converts a string containing lbs to kg.
 
 postprocessor **should** be detecting it

spaceyuck: I don't think it does, the code in [mapped.go](https://github.com/stashapp/stash/blob/1b2b4c52218b25f053e6818fe6f089ffdd50e34e/pkg/scraper/mapped.go#L586) looks quite simple, it's just a parse as float, apply conversion factor, round. If you put KG through that, the number will just be wrong.

feederbox826:
Understood, thought there was a more elegant solution, will try to revert changes

PR #2541, opened 2025-10-17
Fix Motherless scraper to handle deleted videos

tomdotcash:
## Short description

Motherless returns HTTP 200 with error page content instead of 404 when videos are deleted:

```html
<div class="error-page view-page py-3">
    <h1 class="p-0 m-0">
        File not found. Nothing to see here!
    </h1>
            <p class="reason motherless-red text-huge py-2 m-0">
            • The member has deleted the upload •
        </p>
        <p class="text-huge m-0">
        Here's another image instead...
    </p>
</div>
```

Then below it, a random result is returned. The scraper will use the random result to populate the metadata, instead of a no-op as expected.

This PR adds error detection to prevent scraping replacement content by checking for "File not found" error div and returning null to fail the scrape.

## Examples to test

```
https://motherless.com/46E5D3E
```

### DIFF
--- a/scrapers/Motherless.yml
+++ b/scrapers/Motherless.yml
@@ -36,8 +36,17 @@ xPathScrapers:
   sceneScraper:
     common:
       $meta: //div[@class='media-meta']
+      $error: //div[@class='error-page view-page']
     scene:
-      Title: $meta//h1/text()
+      Title:
+        selector: $meta//h1/text()
+        postProcess:
+          - javascript: |
+              // Check if error page div exists (deleted video)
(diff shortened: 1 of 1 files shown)


feederbox826:
the $error selector fails since it also has the py-3 class, it needs contains()

it also does not fail, I'm not sure where you're getting context from but it doesn't return null. and any errors are ignored https://docs.stashapp.cc/in-app-manual/scraping/scraperdevelopment/#post-processing-options

adding a selector doesn't work since it errors out when the video is there as expected, opted for prepending it with the error string

PR #2542, opened 2025-10-17
Refactror SARJ-LLC.py

hyper440:
- Replaced lambdas with list comprehensions or helper functions
- all values are stripped - some tags had trailing spaces, this is now fixed
- added typing


feederbox826:
There are types in py_common.types

PR #2545, opened 2025-10-18
ScrapeWithURL: support multiple URLs

MikeLund:
ScrapeWithURL scraper would crash if any of the URLs on a scene were invalid (not a real hyperlink), and it would also fail if there's more than one URL on a scene.

Now, it does a good-enough sanity check (that the URL at least starts with 'h'), and it scrapes the scene's URLs one by one -- and completes upon first result.


feederbox826:
Thanks!

PR #2552, opened 2025-10-28
[hegre] performer add weight, tags

spaceyuck:

## Scraper type(s)
- [X] performerByURL

## Examples to test

https://www.hegre.com/models/hannah?locale=en
https://www.hegre.com/models/julietta?locale=en

## Short description

I noticed the Hegre scraper did not pick up the given performer weight and their tags. These might have been added to the site after the scraper was created.


feederbox826:
Thanks!

PR #2568, opened 2025-11-16
New python meanbitches scraper to replace old one

thismanyboyfriends2:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [x] sceneByName
- [x] sceneByQueryFragment
- [x] sceneByFragment
- [x] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [x] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

- https://megasite.meanworld.com/scenes/Ashlyn-In-Charge-Ashlyn-Peaks_vids.html
- https://megasite.meanworld.com/scenes/Darryl-Hanah---2009_vids.html
- https://megasite.meanworld.com/scenes/Obedient-Landlord-4-Hotwife-Lexi_vids.html

## Short description

A brand new MeanBitches scraper in python to replace the old simple scraper that only supported the old site (which is now defunct).

Now supports:
- New website
- fragment and search scraping
- multiple fallbacks for images if they don't exist
- searches parallel pages for querying
- caching for search results


feederbox826:
This smells like LLM and the new site looks like it can be parsed just fine with xPath but LLM is too lazy to implement properly

Manifest should not be included under any circumstance.

there's no reason why you would use urllib.request when requests is already part of the base requirements includes.

Converting to draft, the LLM is overcomplicating it, there's no need for async thread-safe python

thismanyboyfriends2:
So I already made an xpath scraper before I made the python one - the main issue really is the cover image. The image is not available on any page which also includes a trailer. Therefore the only way to get that image is to go back to the performer page, and iterates through the pages to find the thumbnail of that scene. The vast majority of the other data can be fetched with the xpath scraper, but I considered the image too important a piece of information to be omitted.

Yeah, the parallel search with async/threading was done by an LLM - this was an attempt at an optimisation because searching through lots of paginated pages just to find an image was taking a while. Happy to remove it, I just found it worked better.

I'll submit a re-do in a bit with your suggestions - but I do think the fact it's python is necessary.

feederbox826:
Cover is available on the scene page, just had to update some selectors and additional processing for title since I went the easier selector way.

there is no need to overcomplicate things with a misguided LLM, they're much too attached to python

PR #2577, opened 2025-11-22
Adding support for scraping image from MensRush scenes

decisive72:
## Scraper type(s)
- [x] sceneByURL

## Examples to test
https://www.mensrush.tv/single.php?id=GT-2330_DL

## Short description
Adding support for scraping image from MensRush scenes

### DIFF
--- a/scrapers/MensRush.yml
+++ b/scrapers/MensRush.yml
@@ -12,8 +12,10 @@ xPathScrapers:
       Details:
         selector: //div[@class="movie_detail"]/h3[1]/following-sibling::p[1]/text()
         concat: "\n\n"
+      Image:
+        selector: //div[@id="movie_area"]/div[@class="player"]/div/img/@src
       Studio:
         Name: //div[@class="movie_detail"]/p/a[contains(@href,"maid")]/text()
       Tags:
         Name: //div[@class="movie_detail"]/p/a[contains(@href,"geid")]/text()
-# Last Updated May 23, 2025
+# Last Updated November 22, 2025


decisive72:
Oops, totally missed the comment in [#2352](https://github.com/stashapp/CommunityScrapers/pull/2352#issuecomment-2914999769).
> Saw the image but it's heavily watermaked, censored and low-res so wasn't sure if I wanted it included
- Heavy watermark: varies significantly studio-by-studio.
- Censored: If this is referring to the genitals, it is a legal requirement, the videos themselves are censored too (and so are any auto-generated thumbnails). Performers' eyes/faces are sometimes censored too for privacy: they don't seem to account for too many of the scenes so this isn't much of a problem in my opinion.
- Low-res: seems like they all come in 800x450, which is typically lower than the videos' 720p or 1080p dimensions, but it's also clear enough for identification.

feederbox826:
Thanks for the info, was not informed at all 😅, didnt want to populate with shit images if better ones were available elsewhere since they are **BAD**

decisive72:
For their in-house/exclusive studios these are as good as they get, at least officially. They also sell 3rd party scenes which are usually available at other storefronts too: for those scenes I usually wouldn't even consider scraping from this one just because MensRush usually adds in a bunch of their own annotations and disclaimers in the description which vary from scene to scene (i.e. a manual or NLP task to figure out what are and aren't their additions, at which point you're better off scraping from a different storefront).

feederbox826:
Thanks!

PR #2578, opened 2025-11-22
[Kink] sceneByURL - Fix Tags

WellDoneLiz:
## Scraper type(s)
- [X] sceneByURL

## Examples to test

[Old Scene](https://www.kink.com/shoot/5757)
[Recent Scene](https://www.kink.com/shoot/46314)
[New Scene](https://www.kink.com/shoot/106962)

## Short description

Update selector to find tags on scene pages.

### DIFF
--- a/scrapers/Kink.yml
+++ b/scrapers/Kink.yml
@@ -137,12 +137,11 @@ xPathScrapers:
           split: ","
       Tags:
         Name:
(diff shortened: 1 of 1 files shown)


feederbox826:
Why select the parent instead of targeting the /a below it? it'll clean up the replace code quite a bit

PR #2580, opened 2025-11-23
fix: ScrapeWithURL crash on invalid result

Sz945Kv:
## Short description

Fixes an issue where ScrapeWithURL would crash and hang up the scraping process whenever one of the many urls did not return a valid result. Resolved by only printing the results when it's not None.

### DIFF
--- a/scrapers/ScrapeWithURL/ScrapeWithURL.py
+++ b/scrapers/ScrapeWithURL/ScrapeWithURL.py
@@ -83,8 +83,8 @@ for url in urls:
             result = scrape_scene(url)
             result = filter_nones(result)
             log.debug(f"result {result}")
-            print(json.dumps(result))
             if result:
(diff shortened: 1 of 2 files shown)


feederbox826:
Not sure if this does what you intend for it, let me take some time to review and make sure

Sz945Kv:
> Not sure if this does what you intend for it, let me take some time to review and make sure

please do!

My tests has been against a scene with a valid & invalid url where the latter did not match any installed scraper. Another was a scene with two urls with matching scrapers but one slug was modified to no longer match.
I'm unsure however if this works because of some quirk in the spec or if it's fully intentional.

feederbox826:
It doesn't do **exactly** what it's supposed to but gets around the error in a weird way. Technically it should be caught by try: but this works as well. 

The code instead returns on the first successful result

Issue #2579, opened 2025-11-22
[PMVHaven] Get video id or download hash from filename rather than title


rororosi:
Assuming that a title of a scene will always be the download name can lead to issues. If a scene title is generated by another scraper or identifier, the PMVHaven scraper will be unable to find the scene, despite the id still being accessible.

Python is not my language of choice but this in sceneByFragment:

```python
    if not (title := dig(params, "title")):
        fail("JSON blob did not contain title property")

    if not (match := re.search(r"([a-z0-9]{24})", title)):
        fail(f"Did not find ID from video title '{title}'")
```

could turn to this:

```python
   fileName = dig(params, "files", 0, "path")

    if not (title := dig(params, "title")):
        fail("JSON blob did not contain title property")

    if not (match := re.search(r"([a-z0-9]{24})", title)):
        if not (match := re.search(r"([a-z0-9]{24})", fileName)):
            fail(f"Did not find ID from video title '{title}' or fileName '{fileName}'")
```

Happy to open a PR if people think this is a good idea.

feederbox826:
Happy to accept a PR! #1814 for context

rororosi:
Thank you! Just opened one up here: #2582

PR #2582, opened 2025-11-24
[PMVHaven] add file name fallback for scene identification

rororosi:
## Short description
Adds file name fallback for sceneByFragment. Sometimes a user will change a scene's title before scraping with this scraper. This ensures that we can at least check the file name, which should still have the download hash.

Resolves:  #2579


feederbox826:
Thanks!

conhowdiesel:
This broke scraping for my machine, put up a PR to fix: #2583

PR #2593, opened 2025-12-01
LadyboyGold: make scene search work for more network sites

nrg101:
## Scraper type(s)
- [x] sceneByName

## Examples to test

- winking
- tayla
- femboy

## Short description

The searchByName implementation has been extended to work with multiple sub-studio site searches.

### DIFF
--- a/scrapers/LadyboyGold/LadyboyGold.yml
+++ b/scrapers/LadyboyGold/LadyboyGold.yml
@@ -1,5 +1,5 @@
 # yaml-language-server: $schema=../../validator/scraper.schema.json
-name: LadyboyGold / TSRAW
+name: LadyboyGold (Network)
 performerByURL:
   - action: scrapeXPath
     url:


feederbox826:
What's the reason for disabling the cache?

nrg101:
When I was testing it with the cache code active the script kept producing an EOF error as if the code was throwing when attempting json.load(f) and I couldn't figure out why, even when wrapping with try/except and adding log lines there were no meaningful errors or logs to say what had gone wrong.

nrg101:
I'll try to uncomment the cache results read from file and see if I can get it working... moving to Draft in the meantime

nrg101:
it seems to be working fine today, no idea what was happening yesterday, will check a few things before removing Draft status

feederbox826:
Can you revert the name change? Since there's no ID it'll appear under a new scraper

nrg101:
> Can you revert the name change? Since there's no ID it'll appear under a new scraper

ah you mean for updating the previous "version" of the scraper, the name needs to remain the same because there's no other identifier? I thought the folder/filename of the YAML did that. Let me know if you still want the name changing back and I can make small PR reverting that line.

feederbox826:
No, it's probably better this way since it no longer reflects the old version

PR #2599, opened 2025-12-05
Update GEVI.py to ignore "none available" in notes field.

Captain-Organize:

## Examples to test

List a few links/titles/names/filenames/codes to test
https://gayeroticvideoindex.com/performer/141367  # Performer with "none available"
https://gayeroticvideoindex.com/performer/31139   # Performer with text in notes field

## Short description

Update Scraper to ignore "none available" in the Notes field so it isn't imported to Performer Details.


Maista6969:
Thanks for fixing this :)

PR #2613, opened 2025-12-09
GroobyNetwork-Partial: fix title and image scraping for transvr.com

nrg101:
## Scraper type(s)
- [x] sceneByURL

## Examples to test

- https://www.transvr.com/tour/trailers/Is-Up-In-Your-Face.html
- https://www.transvr.com/tour/trailers/Dasha-Dimes-Big-Night.html

## Short description

The title appears to be reliably found in the meta, so I added that as a possible selector, and then the cover image only appears in a script tag/block in at least the examples above.

CDP/VPN no longer appears to be necessary as Grooby appear to have at least SFW versions of the site for previously age/geo blocked regions.

### DIFF
--- a/scrapers/GroobyNetwork-Partial.yml
+++ b/scrapers/GroobyNetwork-Partial.yml
@@ -285,7 +285,12 @@ xPathScrapers:
         Name: *tagsSel
   sceneScraperTransVR:
     scene:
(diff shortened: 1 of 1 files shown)


feederbox826:
Missed that this only targets the TransVR block oops!

PR #2619, opened 2025-12-16
feat(AlgoliaAPI): add endowment parsing and unit conversion

xantror:
This PR adds endowment field parsing and unit conversion for performer height/weight into `AlgoliaAPI`.

I'm not sure if some sites that use this already have metric values for height/weight or if endowment is limited to `penis_size` or if it is used differently for females so additional input is welcome.

I tried spot checking with some of the scrapers that used this but all the performers I tried to test had no weight/height/endowment values - needs additional testing.

The only site i managed to test is `chaosmen.com` where it functions as expected.

### DIFF
--- a/scrapers/AlgoliaAPI/AlgoliaAPI.py
+++ b/scrapers/AlgoliaAPI/AlgoliaAPI.py
@@ -173,15 +173,6 @@ def scene_url(site: str, sitename: str, url_title: str, clip_id: str) -> str:
     "Generates URL for a scene"
     return f"{homepage_url(site.lower())}/en/video/{sitename.lower()}/{url_title}/{clip_id}"
 
-def parse_endowment(performer_from_api: dict[str, Any], gender: str | None) -> dict[str, Any] | None:
-    "Parses penis length if gender is male"
-    if (
-        gender == "male"
-        and (endowment := dig(performer_from_api, "attributes", "endowment"))
-    ):
-        return {"penis_length": feet_to_cm("0'" + endowment.strip())}
-    return None
-
 def to_scraped_performer(performer_from_api: dict[str, Any], site: str) -> ScrapedPerformer:
     "Helper function to convert from Algolia's API to Stash's scraper return type"
     performer: ScrapedPerformer = {}
@@ -203,8 +194,8 @@ def to_scraped_performer(performer_from_api: dict[str, Any], site: str) -> Scrap
         performer["height"] = feet_to_cm(height.strip())
     if weight := dig(performer_from_api, "attributes", "weight"):
         performer["weight"] = lb_to_kg(weight.strip())
-    if endowment := parse_endowment(performer_from_api, performer.get("gender")):
-        performer.update(endowment)
+    if endowment := dig(performer_from_api, "attributes", "endowment"):
+        performer["penis_length"] = feet_to_cm("0'" + endowment.strip())
     if home := dig(performer_from_api, "attributes", "home"):
         performer["country"] = guess_nationality(home.strip())
     if performer_from_api.get("has_pictures") and (pictures := performer_from_api.get("pictures")):


scrapers/AlgoliaAPI/AlgoliaAPI.py
```diff
@@ -156,6 +156,15 @@ def scene_url(site: str, sitename: str, url_title: str, clip_id: str) -> str:
     "Generates URL for a scene"
     return f"{homepage_url(site.lower())}/en/video/{sitename.lower()}/{url_title}/{clip_id}"
 
+def parse_endowment(performer_from_api: dict[str, Any], gender: str | None) -> dict[str, Any] | None:
+    "Parses penis length if gender is male"
+    if (
+        gender == "male"
```
nrg101: If `endowment` also applies to transgender_female, then this should be:

```python
gender in ("male", "transgender_female")
```

NOTE: this is now redundant, see more recent comment

xantror: Do you happen to have some example cases I can test against?

nrg101: hmm, I need to check the gender values from Algolia, but I could find any trans-female actors with endowment attribute populated 

what I mean is, the sites I checked don't have attributes populated for any actor, so I'll have to check a bunch of sites

nrg101: hmm, I need to check the gender values from Algolia, but I could find any trans-female actors with endowment attribute populated

xantror: That was the exact issue I ran into when I wanted to test outside of my main use-case `chaosmen.com` 🤣

nrg101:
Having looked at this more, I don't think you need the separate function and could attempt to get the endowment attribute in the same way that ethnicity is attempted, and then assign performer["penis_length"] with the feet_to_cm converted value, i.e. just two lines. 

You shouldn't have to include the gender in the condition either, and can assume that any actor with an endowment attribute populated is referring to their penis length. I can't think what else endowment would be measuring.

scrapers/AlgoliaAPI/AlgoliaAPI.py
nrg101: I'd suggest this within `to_scraped_performer`:

```python
if endowment := dig(performer_from_api, "attributes", "endowment"):
    performer["penis_length"] = feet_to_cm("0'" + endowment.strip())
```

and remove the `def parse_endowment` function

xantror:
Updated and rebased

xantror:
> The code looks good, but can you provide one or two examples so I can test each conversion taking place?

1. https://www.chaosmen.com/en/pornstar/view/Blake/93119
   ```json
                    "attributes": {
                        "eye_color": "",
                        "hair_color": "",
                        "body_type": "",
                        "ethnicity": "",
                        "alternate_names": "",
                        "best_know_for": "",
                        "weight": "185.00",
                        "height": "6'5\"",
                        "endowment": "8\"",
                        "home": "",
                        "exclusive": "",
                        "centerfold_date": ""
                    }
   ```
3. https://www.maskurbate.com/en/pornstar/view/Brad/33779
   ```json
                    "attributes": {
                        "eye_color": "Blue",
                        "hair_color": "Blonde",
                        "body_type": "",
                        "ethnicity": "",
                        "alternate_names": "",
                        "best_know_for": "",
                        "weight": "0.00",
                        "height": "5'11\"",
                        "endowment": "8\"",
                        "home": "",
                        "exclusive": "",
                        "centerfold_date": ""
                    }
   ```
4. https://www.girlcore.com/en/pornstar/view/Abigail-Mac/34517
   ```json
                    "attributes": {
                        "eye_color": "",
                        "hair_color": "Brown",
                        "body_type": "",
                        "ethnicity": "",
                        "alternate_names": "Abi",
                        "best_know_for": "",
                        "weight": "107.00",
                        "height": "5'2\"",
                        "endowment": "",
                        "home": "",
                        "exclusive": "",
                        "centerfold_date": ""
                    }
   ```

scrapers/AlgoliaAPI/AlgoliaAPI.py
```diff
@@ -191,9 +191,11 @@ def to_scraped_performer(performer_from_api: dict[str, Any], site: str) -> Scrap
     if alternate_names := dig(performer_from_api, "attributes", "alternate_names"):
         performer["aliases"] = alternate_names.strip()
     if height := dig(performer_from_api, "attributes", "height"):
-        performer["height"] = height.strip()
+        performer["height"] = feet_to_cm(height.strip())
     if weight := dig(performer_from_api, "attributes", "weight"):
-        performer["weight"] = weight.strip()
+        performer["weight"] = lb_to_kg(weight.strip())
```
nrg101, line 196: In the example https://www.maskurbate.com/en/pornstar/view/Brad/33779, the API has the weight attribute as:

```
"weight": "0.00",
```

This is invalid in the populated field in stash web UI, so you could add a condition for non-zero like:

```python
kg = lb_to_kg(weight.strip())
if kg > 0:
    performer["weight"] = kg
```

and probably do the same approach for `height` and `penis_length`

xantror:
Merge after #2627

feederbox826:
Thanks!

PR #2628, opened 2025-12-27
feat(py_common): Update py_common types to match Stash schema

xantror:
Align py_common TypedDicts with Stash schema: add stored IDs, replace deprecated single url with urls, rename ScrapedMovie → ScrapedGroup, and remove deprecated fields.

- ScrapedTag/Performer/Studio/Group/Gallery/Scene/Image: added stored_id, codes, photographers, directors, durations, and urls where appropriate.
- ScrapedMovie renamed to ScrapedGroup; ScrapedMovie kept as alias.
- Removed deprecated single url fields and other deprecated fields; added deprecation docstrings where applicable.

### DIFF
--- a/scrapers/py_common/types.py
+++ b/scrapers/py_common/types.py
@@ -114,9 +114,9 @@ class ScrapedGroup(TypedDict, total=False):
     studio: ScrapedStudio
     tags: list[ScrapedTag]
     front_image: str
-    "This should be a base64 encoded data URL"
+    "Image can be a URL or base64-encoded data URL"
     back_image: str
-    "This should be a base64 encoded data URL"
+    "Image can be a URL or base64-encoded data URL"
 
 
 # ScrapedMovie is deprecated in favor of ScrapedGroup
@@ -144,7 +144,7 @@ class ScrapedScene(TypedDict, total=False):
     urls: list[str]
     date: str
     image: str
-    "This should be a base64 encoded data URL"
+    "Image can be a URL or base64-encoded data URL"
     studio: ScrapedStudio
     tags: list[ScrapedTag]
     performers: list[ScrapedPerformer]


feederbox826:
don't add back url

scrapers/py_common/types.py
```diff
@@ -29,14 +29,19 @@
...
     "Name is the only required field"
+    stored_id: str
+    "Set if tag matched"
 
 
 class ScrapedPerformer(TypedDict, total=False):
     name: Required[str]
     "Name is the only required field"
+    stored_id: str
+    "Set if performer matched"
```
Maista6969, line 44: What does this mean? Are there cases where scrapers do their own matching against Stash's database? I think this is an internal field that scrapers do not need to be concerned with

scrapers/py_common/types.py
```diff
@@ -85,70 +88,82 @@ class ScrapedPerformer(TypedDict, total=False):
...
+    "Must be in the format YYYY-MM-DD"
+    rating: str
     director: str
+    urls: list[str]
     synopsis: str
     studio: ScrapedStudio
-    rating: str
+    tags: list[ScrapedTag]
     front_image: str
+    "This should be a base64 encoded data URL"
```
Maista6969: Images can (and I often prefer them to be) regular ol' URLs that the Stash frontend can resolve

The only reason they'd need to be [data URIs](<https://en.wikipedia.org/wiki/Data_URI_scheme>) is if they're inaccessible due to authentication or CORS

scrapers/py_common/types.py
xantror: The field is supported by Stash as far as I could see e.g. if somebody were to do some gql query to match within their scraper they could chose to return this field for a more exact match than what Stash might provide internally.

I don't think that any of the current scrapers use this functionality but who knows what people might need in the future.

scrapers/py_common/types.py
xantror: Ah, I think i just reused the original comment on one of the image fields for all of them - I'll update that.

feederbox826 (approved):
Long overdue but LGTM. I'll try to resolve some other PRs to avoid merge nightmares or just YOLO if I don't get to it in time

PR #2632, opened 2026-01-02
Update fc2.yml

bennyz327:
Remove invalid or unavailable items, and update the currently available ones.

## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [x] sceneByFragment
- [x] sceneByURL
- [x] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

https://adult.contents.fc2.com/article/1130822/
https://adult.contents.fc2.com/article/339284/

## Short description

I updated some incorrect path retrievals and removed parts that are essentially unable to get metadata.

### DIFF
--- a/scrapers/fc2.yml
+++ b/scrapers/fc2.yml
@@ -49,11 +49,7 @@ xPathScrapers:
         selector: //section[@class="items_comment_sellerBox"]/div/div/h4/a
       Studio:
         Name:
-          selector: /html/head/title/text() 
-          postProcess:
-            - replace:
-                - regex: .*(FC2).*
-                  with: $1
+          fixed: FC2 
       Code:
         selector: //meta[@name="description"]/@content
         postProcess:
@@ -86,12 +82,9 @@ xPathScrapers:
         selector: //section[@class="items_comment_sellerBox"]/div/div/h4/a
       Studio:
         Name:
-          selector: /html/head/title/text() 
-          postProcess:
-            - replace:
-                - regex: .*(FC2).*
-                  with: $1
+          fixed: FC2 
       Tags:
         Name: //a[@class='tag tagTag']/text()
 # Last Updated January 3, 2025
+
 


scrapers/fc2.yml
```diff
@@ -22,57 +22,53 @@ groupByURL:
...
-      Performers:
-        Name: //*[@id="top"]/div[1]/section[1]/div/section/div[2]/ul/li[3]/a/text()
-      Tags:
-        Name: //a[@class='tag tagTag']/text()
+      Title: 
+        selector: //div[@class='items_article_headerInfo']/h3/text()
+        concat: " "
+        postProcess:
+          - replace:
+            - regex: '\s+'
```
feederbox826: why?

scrapers/fc2.yml
```diff
@@ -22,57 +22,53 @@ groupByURL:
...
+              with: ''
+      URLs: /html/head/meta[@property="og:url"]/@content
       Image: 
         selector: //div[@class='items_article_MainitemThumb']/span/img/@src
         postProcess:
             - replace:
                 - regex: ^
                   with: "https:"
+      Director:
+        selector: //section[@class="items_comment_sellerBox"]/div/div/h4/a
```
feederbox826, line 49: Looks to be the seller, not the director

scrapers/fc2.yml
```diff
@@ -86,11 +82,14 @@ xPathScrapers:
...
       Studio:
         Name:
-          selector: /html/head/title/text() 
-          postProcess:
-            - replace:
-                - regex: .*(FC2).*
-                  with: $1
-# Last Updated June 26, 2024
+          selector: //section[@class="items_comment_sellerBox"]/div/div/h4/a
```
feederbox826: Reused director tag, so which is it?

scrapers/fc2.yml
bennyz327: the seller is usually both the director and the performer. If this is inappropriate, I can remove it.

bennyz327:
Change list

1. **Title**: Use the product name instead of the page title to obtain a cleaner name.
2. Update **URL** → **URLs**.
3. Add director field population (on FC2, the seller is usually both the director and the male performer; if this is inappropriate, it can be removed).
4. **Code**: Fix the source of code population to obtain the format `FC2-PPV-XXXXXXX`.

bennyz327:
@feederbox826 I’ve cleaned up the code, and it should be easier to read now.

scrapers/fc2.yml
bennyz327: because the text inside the h3 is split into multiple text nodes by span elements. When Stash evaluates //h3/text(), it only keeps the first text node and silently drops the rest. Adding concat ensures all text nodes are merged into a single string before post-processing, otherwise part of the title is lost at the XPath stage and cannot be recovered later.

feederbox826: that's the concat, why the regex for \s+

bennyz327: It’s just personal preference. I can keep only the concat and remove the regex

scrapers/fc2.yml
bennyz327: the seller is usually both the director and the performer. If this is inappropriate, I can remove it.

scrapers/fc2.yml
```diff
@@ -56,23 +59,24 @@ xPathScrapers:
                 - regex: .*(FC2).*
                   with: $1
```
feederbox826: Is this always supposed to be FC2? 

https://docs.stashapp.cc/in-app-manual/scraping/scraperdevelopment/#fixed-attribute-values

might be more helpful than a static regex replace

bennyz327: yeah, this always been parsed as FC2

PR #2638, opened 2026-01-11
rule34.xxx fragment scrapers

ghost:
**NOTE: This scraper was mostly generated by Claude Code (Sonnet), and has been validated and tested by me with minimal modification.**

Adds a scene and image fragment scraper which matches filenames to either an md5 hash or the general format of `r34_{post_id_number}_{...}.{ext}`.

Tag categorization:
- Characters → Performers
- Artists → Studio (first non-voice actor artist)
- Voice actors → Performers
- Copyright → Tags
- General → Tags
- Meta → Tags

Open to feedback on this one. I have been treating Studios as a direct equivalent to Artists in my personal instance, but it's a bit of a bummer I can't assign multiple to the same media to keep parity with sites like `rule34.xxx` and other booru.

### DIFF
--- a/scrapers/Rule34xxx/README.md
+++ b/scrapers/Rule34xxx/README.md
@@ -1,0 +1,11 @@
+# Rule34xxx Stash Scraper
+
+This scrapes the html to isolate performers from tags and artists.
+
+Some things to be aware of:
+
+- This scraper works for files named with their md5 hash, or files named in the general format of `r34_{post_id_number}_{...}.{ext}`.
+- Meta, Copyright, and General tags get collapsed into normal stashapp tags.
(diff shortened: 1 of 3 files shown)


feederbox826:
just going to nip this in the bud immediately, python isn't required for this.

Stash does md5 calculations already and supports query via {checksum} https://docs.stashapp.cc/in-app-manual/scraping/scraperdevelopment/#scrapexpath-and-scrapejson-use-with-scenebyfragment-and-scenebyqueryfragment

also having multiple seperate scrapers instead of just reusing one since they all use the similar format

ghost:
Thanks, I'll take a look at simplifying this when I get a chance, if possible.

PR #2668, opened 2026-02-22
Babepedia: fix country extraction and alias whitespace

FlashSpazzbo:
## Scraper type(s)

- [x] performerByURL

### performerByURL
- **Country:** Replaced text-based nationality extraction with ISO code parsed from Babepedia's flag icon CSS classes (e.g. `fi-gb` → `GB`). The old approach returned malformed values like `British),` for performers with multiple nationalities. Takes the first nationality only.
- **Aliases:** Added `.strip()` before splitting the AKA string, removing leading/trailing whitespace that was causing spurious "replace aliases?" prompts in Stash.
- `fake_tits`: Replaced `str(not real_breasts)` logic which was storing Python's `"True"`/`"False"` with a `sanitize_fake_tits()` function that maps Babepedia's labels to Stash's valid values (`"Fake"`, `"Natural"`). Unrecognised values leave the field unset.

## Test URLs
- https://www.babepedia.com/babe/India_De_Beaufort (multi-nationality, alias whitespace)
- https://www.babepedia.com/babe/Seltin (single nationality, alias whitespace)
- https://www.babepedia.com/babe/Heather_Jo_Hughes 

## Notes
Built on top of the recent Babepedia changes in upstream — no conflicts.

### DIFF
--- a/scrapers/Babepedia/Babepedia.py
+++ b/scrapers/Babepedia/Babepedia.py
@@ -45,6 +45,22 @@ def sanitize_hair_color(str) -> HairColor:
         return "Brunette" # type: ignore
     return str # type: ignore
 
+def sanitize_fake_tits(value: str) -> str | None:
+    # Maps Babepedia's breast type labels to Stash's valid fake_tits values:
+    # "Fake", "Natural", or "Na". We never return "Na" here — if Babepedia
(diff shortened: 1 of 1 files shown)


feederbox826:
Will merge and replay changes on top of it, since it's using the new proxy library

PR #2669, opened 2026-02-23
PlayboyPlus: fix country mapping, performer images, and hair color

FlashSpazzbo:
## Scraper type(s)

- [x] performerByURL
- [x] sceneByName

## Examples to test

- https://www.playboyplus.com/en/model/view/Alanna-Hensley/118653
- https://www.playboyplus.com/en/model/view/Adeanna-Cooke/119840

## Short description

**Images:** The `pictures` field in the Algolia response can contain broken legacy URLs. Switched to `multicontent_data.nsfw` as the primary image source, filtering for the `modelHeroMobile` entry (1000×1600 portrait) — this is the same structure already used for scene images and consistently points to valid CDN-hosted files. `pictures` is retained as a fallback for performers where `multicontent_data` is absent. Also removes a previous hardcoded fallback that guessed a URL suffix number which varies per performer.
 
**Country:** Replaced fragile string-splitting logic with a `COUNTRY_MAP` dict. The `home` field is a full location string (e.g. `"Burlington DC United States"`), so matching uses `endswith`. Fixes `"Kingdom"` for United Kingdom and `"N/A"` being stored as a country value. Also corrects `"USA"` → `"US"` (alpha-2). Unknown values are logged and omitted.
 
**Hair color:** Added `sanitize_hair_color()` with `HAIR_COLOR_MAP`. Fixes `"Brown"` → `"Brunette"` to match StashDB's internal value.

### DIFF
--- a/scrapers/PlayboyPlus/PlayboyPlus.py
+++ b/scrapers/PlayboyPlus/PlayboyPlus.py
@@ -2,7 +2,6 @@ import json
 import re
 import sys
 import html
-import urllib3
 from urllib.parse import quote_plus
 
 from py_common import log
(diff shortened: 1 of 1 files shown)


feederbox826:
https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/py_common/util.py#L352

and colour map can be partial passthrough with .get() instead of populating all values

PR #2671, opened 2026-02-27
Babepedia: fix performer-by-name URL encoding

FlashSpazzbo:
## Scraper type(s)
- [x] performerByName

## What changed:
 `performer_by_name` now uses `params=` to properly URL-encode the search term, and truncates to the first word of the name since the Babepedia API only matches reliably on the start of the name.

## Test:
 Search for `Jessica-Jane Clement` — previously returned `[]`, now returns the correct result.

## Known limitation:
 Hyphenated first names like `Marie-Claude Bourbonnais` may return incomplete or incorrect results due to a Babepedia API limitation outside our control. `performer-by-url` works correctly for these performers.

### DIFF
--- a/scrapers/Babepedia/Babepedia.py
+++ b/scrapers/Babepedia/Babepedia.py
@@ -204,11 +204,11 @@ def map_performer_search(performer) -> PerformerSearchResult:
     return result
 
 def performer_by_name(name) -> list[PerformerSearchResult]:
-    url = f"https://www.babepedia.com/ajax-search.php?term={name}"
-    scraped = scraper.get(url)
+    search_term = name.replace('-', ' ')
+    scraped = scraper.get("https://www.babepedia.com/ajax-search.php", params={"term": search_term})
     scraped.raise_for_status()
     data = scraped.json()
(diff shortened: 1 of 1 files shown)


FlashSpazzbo:
You're right, and thanks for pushing back on this. After more testing I found that full-name search does work correctly for non-hyphenated names — the problem was specifically that hyphens in the search term were breaking the API match. Searching `Jessica-Jane Clement` (with hyphen) returns `[]`, but stripping the hyphen to `Jessica Jane Clement` returns the correct result. Same for `Marie-Claude Bourbonnais`.

So the actual fix is just stripping hyphens before sending the query:

```python
search_term = name.replace('-', ' ')
```

I've updated the commit with this approach instead of the first-word truncation. Tested against:
- `Jessica-Jane Clement` → returns correct result
- `Marie-Claude Bourbonnais` → returns correct result  
- `Jill` → returns expected broad results (same as before)
- `Jill Scott` → returns only Jill Scott results (full-name matching working correctly)

Apologies for the noise on the first attempt.

PR #2675, opened 2026-02-28
Fix 10Musume-JP scraper ActressesJa mapping

codykee:
The 10Musume.com JSON endpoint has replaced the key `Actresses` with `ActressesJa`. This PR updates the `10Musume-JP` YAML scraper to correctly match Performers again.


feederbox826:
Looks like everything else prioritizes jp content

PR #2678, opened 2026-03-01
PlayboyTV: add Algolia-based scene scraper for current site

FlashSpazzbo:
## Scraper type(s)
- [x] sceneByURL

## Examples to test
https://www.playboytv.com/en/episode/view/Episode-1/261641
https://members.playboytv.com/en/episode/view/Episode-1/261641
https://www.playboytv.com/en/episode/playboytv/Episode-1/261641
https://members.playboytv.com/en/episode/playboytv/Episode-1/261641

## Short description
PlayboyTV relaunched their site and the old XPath scraper no longer works. This adds a Python scraper using the Algolia API (same pattern as PlayboyPlus). Scrapes title, date, studio, description, tags, image, code, and URLs. Performer data isn't available — `actors` is consistently empty in the API response.

Title is constructed as `Serie Name - S01E01` since PlayboyTV doesn't use unique episode titles. The site has two URL path formats (`/episode/view/` and `/episode/playboytv/`), so the YAML matches both and all four variants (www + members × both formats) are written back on scrape.

The legacy XPath scraper is preserved for archived `playboy.tv/watch/` URLs.

### DIFF
--- a/scrapers/PlayboyTV/PlayboyTV.py
+++ b/scrapers/PlayboyTV/PlayboyTV.py
@@ -37,8 +37,6 @@ def _build_urls(api_scene: dict) -> list[str]:
     url_title = api_scene.get("url_title", "")
     clip_id = api_scene.get("clip_id", "")
     return [
-        f"https://www.playboytv.com/en/episode/view/{url_title}/{clip_id}",
-        f"https://members.playboytv.com/en/episode/view/{url_title}/{clip_id}",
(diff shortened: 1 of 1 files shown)


feederbox826:
SX EX notation should be derived as-is on the site, `Season 1, Episode 1`

Any reason you're not using the ALgolia(API) package instead?

FlashSpazzbo:
Thanks for the pointers! Assumed the SxEx notation, and was modelling off PlayboyPlus which hits Algolia directly via requests — wasn't aware AlgoliaAPI.py existed.

Issue #2623, opened 2025-12-19
TeamSkeet / Paper Street / Reptyle scrapers are both fragmented and unreliable


AdultSun:
## Fragmentation

The family of studios under the recent "Reptyle" re-branding and reorganization are split between two different scrapers, one named after primary network TeamSkeet and another after parent company Paper Street Media. Most of their scenes have several valid studio URLs from various official domains.

Different data is returned depending on which of the two scrapers is called, determined by which domain happens to be in the URL. Typically data from the PSM scraper is less complete compared to its TS counterpart. For example, networks are often returned instead of the sub-studio and studio codes are not added.

I believe that all of the Reptyle studios now share the same database and API, so we should be able to use a single scraper to cover all of them.

## Unreliable

With that said, I have run into several examples where the TS scraper fails to find a match through the API, even though the PSM scraper is successful.

Reptyle does not include a UUID in their public scene URLs, relying on a brittle title slug instead. I have seen examples of scenes listed in a performer's profile that redirect to a completely different scene with the same title. I have also seen Reptyle change a scene's name after release, causing the old URL to redirect to a completely different scene similar to the old title.

I believe the TS scraper often fails because it relies on the title slug to find a match through the API. If the title changes, if it's not unique, or if it uses special characters / punctuation (which is very common), the scraper can't find a match.

Lately, it feels like the TS scraper only works about half the time.

## Member-Exclusive Data

This has been an enduring issue for StashDB, but there might not be much we can do about it. Specifically, Reptyle's members area includes the male performers, but the public pages and public API scrapes do not. Without a membership, we have to rely on visual ID for these performers or wait / hope that Data18 scrapes with a membership for us.

I believe there is already a way to scrape this data through the API, but only if you have a member's access token. Again, not sure there's anything else we can do about that, but it's worth looking into.

## Possible Solutions

Ideally we would use TS's internal UUID instead of the title slug, but unfortunately that's a bit of a catch-22. The easiest way to find the studio code is by scraping it, so how do we scrape it without already having the studio code?

Using [this scene](https://stashdb.org/scenes/019b221e-5099-7fbf-981a-4c7cb9547cb6) as an example, I can find the UUID of `31374` in a couple different places in page source of both the `teamskeet.com` and `freeuse.com` domains. Hopefully that means we can extract it through a scrape of the webpage, then feed it into the API to get richer data.

[...]

PR #2680, opened 2026-03-03
Fix sayuncle scraping in TeamskeetAPI.py

gimmeliina:
Sayuncle.com api implementation doesn't work with special character in title. Nowadays nearly all titles have some special characters, so changed api handling using same as swappz and freeuse do.


feederbox826:
related #2623

the /_search returns 6 results, instead of the guaranteed one we get from the /doc lookup, are there any other endpoints that are more reliable instead of ?size=1?

gimmeliina:
Not played with API much, but didn't figure out anything.
Thought this as quick fix as ~80% of new scenes returns error with old implementation. Didn't get wrong scenes when testing.

feederbox826:
turns out it's in the way they strip/encode their index, and also a byproduct of query-param searching. ended up convering to bundle searching since /doc id is unreliable

PR #2682, opened 2026-03-03
Add scene numbering support and category_tags for AlgoliaAPI.py

gimmeliina:
- Add support for parsing scene number to title (e. g. Falcon Studios uses titles without scene numbering, requiring manual editing)
- Add category tag to tag list (e. g. Gay)
- Add tags to movies

PR #2683, opened 2026-03-03
Move HotHouse, Falcon, Ragingstallion and netdoorstudios to AlgoliaApi

gimmeliina:
#2682


feederbox826:
Missing some changes from Algolia to AlgoliaAPI, will fix in a follow-up commit since I can't seem to figure out how to checkout this PR

Tested FalconStudios

PR #2689, opened 2026-03-14
[nhentai] added galleryByFragment with ID from filename / path prefix

spaceyuck:
Just for convenience,  for stuff downloaded with gallery-dl. The same thing already exists for ehentai.

### DIFF
--- a/scrapers/nhentai.yml
+++ b/scrapers/nhentai.yml
@@ -6,9 +6,7 @@ galleryByFragment:
   queryURL: https://nhentai.net/g/{filename}/
   queryURLReplace:
     filename:
-      - regex: "[^a-zA-Z\\d\\-._~]" # clean filename down to basic charatcers to remove any brackets etc
-        with: ""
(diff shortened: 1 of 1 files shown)


scrapers/nhentai.yml
```diff
@@ -1,9 +1,23 @@
 name: nhentai
+
+# intended to extract gallery ID from folder / zip / cbz file names as produces by gallery-dl ("1234567 some gallery name.zip")
+galleryByFragment:
+  action: scrapeXPath
+  queryURL: https://nhentai.net/g/{filename}/
+  queryURLReplace:
+    filename:
+      - regex: "[^a-zA-Z\\d\\-._~]" # clean filename down to basic charatcers to remove any brackets etc
```
feederbox826: any reason not to just grab the numeric sequence directly? or is it indexed and size keeps on growing

spaceyuck: Yeah, I had examples where the ID in the file name sometimes was wrapped in brackets (like "[123456] title abcd...", presumably because some path metadata parsers ignore stuff in brackets), and sometimes not, so this is to support more possible variations.

PR #2690, opened 2026-03-15
Update House of Gord scraper

Misofist:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

 - `https://www.houseofgord.com/d/sbi-i---trouble-at-the-house-of-gord`
 - `https://www.houseofgord.com/d/the-box-dilemma`
 - `https://www.houseofgord.com/d/wenona-siloed`

## Short description

Changes have been made to the Gord website which have rendered the old scraper non-functional.

This change fixes the scraper to work against the current website.

NOTE: I was unable to get image scraping to work. The image URLs were valid and worked in my browser, but Stash gave an error when trying to save scenes with them.

But HoG's images aren't really suitable for use as thumbnails anyway, Stash's thumbnail generator produces better results. So I've just disabled the image scraping.

### DIFF
--- a/scrapers/HouseOfGord.yml
+++ b/scrapers/HouseOfGord.yml
@@ -9,28 +9,26 @@ xPathScrapers:
   sceneScraper:
     scene:
       Title:
-        selector: //div[@id="detail_title"]/text()
+        selector: //*[@class="detail_title_etc"]/div[contains(concat(' ', @class, ' '), ' leading-none ')]/text()
       Date:
-        selector: //div[@class="preview_feature_date"][contains(text(),"Video")]/text()|//div[@class="preview_feature_date"][1]/text()
+        selector: //*[@class="detail_title_etc"]/div[@class="links_to_pages"]/div[@class="text-xl"][contains(text()[1],"Video")]/text()[1]|//*[@class="detail_title_etc"]/div[@class="links_to_pages"]/div[@class="text-xl"][1]/text()
         postProcess:
           - replace:
               - regex: .*\s+added\s+(.*)
                 with: $1
           - parseDate: 02 Jan 2006
       Details:
-        selector: //p[@class="preview_description"]/text()
+        selector: //div[@id="main_content"]//p[@sanitize="false"]/text()
         concat: "\n\n"
       Performers:
         Name:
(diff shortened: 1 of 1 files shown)


feederbox826:
"remastered" sounds like it should have date precedence but i'm unfamiliar with the site

Misofist:
I believe that when they remaster a scene they update the original scene page, rather than creating a new page for the remastered scene.

My understanding is that Stash's Ministry of Truth generally prefers for scrapers to report the original / earliest metadata available for a scene based on whatever unique identifying property is used? As in, given a certain URL, we should aim to return data as close as possible to whatever was first published at that URL?

Could be wrong, please tell me your thoughts :)

feederbox826:
> For [remasters/redistributions/re-releases](https://guidelines.stashdb.org/docs/scenes/create/re-releases/), the earliest release date refers to that particular release and not the original scene

https://guidelines.stashdb.org/docs/scenes/edit/scene-date/preferred-scene-dates/

should be possible with a selector for the "Remastered" date and then `|` for the other date

Misofist:
> Remasters [...] and re-releases [...] are allowed **when they have a unique scene entry and URL** on a studio website.

In House of Gord's case, the remasters are added by modifying the existing scene entry, using the same URL as the original release.

I think the current logic is correct.

feederbox826:
But it's a remaster taking over the original URL. shouldn't we take the date of the remaster?

Misofist:
The remaster does not have a unique URL, it's re-using the original URL. So based on the guidelines, the remasters / re-releases aren't eligible for their own entry, only the original release gets an entry. And "The earliest known release date for a scene is always the preferred scene date."

This is consistent with what I've been told regarding scenes that have had their titles modified too, they need to be submitted under their original title unless they were re-released at a new URL.

feederbox826:
Ugh the site labelling it as a remaster doesnt help if they're messing with it in -place

PR #2695, opened 2026-03-17
Fix the issue of CzechAV failing to obtain the Title

DDXDB:


feederbox826:
should the title include the series prefix? it seems inconsistnt at this point and we should probably scrape from //title instead

DDXDB:
> should the title include the series prefix? it seems inconsistnt at this point and we should probably scrape from //title instead

Now, the strings will be merged and the possible series prefixes will be ignored.

PR #2699, opened 2026-03-20
Fix CarnalPlus title and add jockpack.com

gimmeliina:
- Barebackplus and jockpack doesn't have working <title> -tag atm (maybe related to site rebranding?) so using meta og:title 
- added jockpack.com to supported site (BB+ is rebranding)

test link: https://jockpack.com/videos/hunter-nash-cumdmp-1.html

### DIFF
--- a/scrapers/CarnalPlus.yml
+++ b/scrapers/CarnalPlus.yml
@@ -45,6 +45,7 @@ sceneByURL:
       - barebackplus.com/videos/
       - carnalplus.com/videos/
       - ftmplus.com/videos/
+      - jockpack.com/videos/
     scraper: networkScraper
 xPathScrapers:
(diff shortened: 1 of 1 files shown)


feederbox826:
//video-detail-h2 ? it looks like its also including series title atm

gimmeliina:
> //video-detail-h2 ? it looks like its also including series title atm

As an alternative?
No, because network and individual studio sites have slightly different layout. Title tag is same across them.

Old version used that and that leaded a lot of duplicates and confusion in stashdb. Site title also includes series title if available.

I will look ahead if after rebranding they changing something, but this is quick temporary fix. 

Maybe at some point whole scraper should written in Python as there is lot of quirky things there..

PR #2720, opened 2026-04-20
Update SARJ-LLC.py

ghost:
Update SARJ-LLC to use "descriptionSfw" API data when "description" is empty.

### DIFF
--- a/scrapers/SARJ-LLC/SARJ-LLC.py
+++ b/scrapers/SARJ-LLC/SARJ-LLC.py
@@ -192,14 +192,9 @@ def map_media(data, studio, base_url):
                         directors.append(name)
     director = ", ".join(directors).strip()
 
-    details = data.get('description', "") or None
-    # If "description" is missing or empty, try "descriptionSfw"
-    if not details:
-        details = data.get("descriptionSfw") or None
-
     return {
         'Title': data['name'].strip(),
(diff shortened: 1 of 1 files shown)


feederbox826:
https://guidelines.stashdb.org/docs/scenes/edit/scene-description/preferred-scene-descriptions/

If it's sanitized text, not taking it

zoemcastillo:
@feederbox826 From what I can tell, the API returns a scene description either in "description" or "descriptionSfw" field, not both. It does not look like the studio is writing two versions of description, they just stick what they have into one or the other. If both fields were to be populated, then the updated script would still select "description" data. In the current state, this scraper just leaves details empty if the only description field returned by the API is "descriptionSfw."

Here are some examples:
1. Data is present in "description" but not in "descriptionSfw":
https://www.metartx.com/api/movie?name=CHRISTMAS_WISH_2&date=20251225
2. Data is present "descriptionSfw" but not in in "description":
https://www.metartx.com/api/movie?name=EFFORTLESS_CLASS_2&date=20260420

feederbox826:
Not familiar with it, i'll try to have a look and play around with vpn locations. the title of the field is quite worrying

feederbox826 (approved):

feederbox826:
Thanks!

PR #2729, opened 2026-04-30
Improve Dream Tranny

nrg101:
## Scraper type(s)
- [x] performerByName
- [x] performerByFragment
- [x] performerByURL
- [x] sceneByName
- [x] sceneByQueryFragment
- [x] sceneByFragment
- [x] sceneByURL
- [x] galleryByFragment
- [x] galleryByURL

### performerByName

This gives a list of performers that have a name that partially matches the search string.

- Cintia
- natal
- lan

### performerByFragment

Use a name from above or URL from below.

### performerByURL

- https://dreamtranny.com/models/393/
- https://dreamtranny.com/models/406/

### sceneByName

This uses the site search and scrapes the first page of results.

- machine
- emylie

### sceneByFragment / sceneByQueryFragment

Use a partial title from above or URL from below.

### sceneByURL

- https://dreamtranny.com/update/1260/?nats=MC4wLjMuMTUzLjAuMC4wLjAuMA&step=2
- https://dreamtranny.com/update/1258/
- https://dreamtranny.com/update/1262/

### galleryByFragment

Matches by title via scene search

- Ayra Crams Her Pleasure-Pole Up a Dude's Rectum

### galleryByURL

Galleries appear on the same page as the scene.

- https://www.dreamtranny.com/update/1260/

## Short description

The scene search was a bit limited by the xpath scraper. It can also be used for performer and gallery searching. Refactoring to Python makes it possible to scrape each and keep them separate instead of mangling search results together.


feederbox826:
Hm really dislike the python rewrite, the only thing I see that might need it is the performers search. Would it not be possible to have both without mangling just with selectors?

nrg101:
Hmm, you mean a hybrid xpath and python scraper? I guess it could be done. Did you want to keep the xpath scraper for simplicity? 

The main problem I had is that the stash search ByName xpath scraping seems to have a bug/shortcoming in that it combines all the performers in a results page into one big list that is set for each sceneByName result, I think tags too. Also, attempting to sub-scrape each search result ended up just sub-scraping the first search page result URL and scraping that scene into every sceneByName result. 

I could back out the python refactoring for what already was fully working with xpath scraping, but I thought it made more sense to just migrate everything to python if implementing the searches in python.

If you can get scene searchByName working with xpath scraping with the same attributes scraped as I've done in python, I'd be interested to see how it is achieved.

feederbox826:
> If you can get scene searchByName working with xpath scraping with the same attributes scraped as I've done in python, I'd be interested to see how it is achieved.

I'll give it a shot. If it can't be done I will succumb to python rewrite :/

nrg101:
Ok, nice. Just to clarify, I mean a search that gives multiple scenes, e.g. search for something like: redhead

I could not get the performers for each scene to be listed for each scene.

PR #2731, opened 2026-05-02
Update KellyMadisonMedia scraper

tj2222:
* Fixes date extraction for PF/TF/KM
* Adds 8k sites to the list of scrapable sites; corrects capitalization of 5Kporn to match the site itself


feederbox826:
gross. at least it's still in ldjson

PR #2732, opened 2026-05-02
[nhentai] fix performers, caputre more tags, studio from artist or group

spaceyuck:
Current state of scraper was definitely wrong, artists are not performers, some tags were lost.

Essentially, the changes are to capture the same values into the same fields as before the needed switchover to the API

I also commented out the API key since it seems to be working without one.

### DIFF
--- a/scrapers/nhentai.yml
+++ b/scrapers/nhentai.yml
@@ -29,6 +29,8 @@ jsonScrapers:
       Details:
         selector: "{title.english,title.japanese}.@values"
         concat: "\n\n"
+      Code: 
+        selector: id
       URL: 
         selector: id
         postProcess:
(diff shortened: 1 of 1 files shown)


scrapers/nhentai.yml
```diff
@@ -39,22 +41,24 @@ jsonScrapers:
...
+        selector: 'tags.#(type=="group")#|0.name'
       Tags:
-        Name: tags.#(type=="tag")#.name
+        Name: '[tags.#(type=="tag")#.name,tags.#(type=="parody")#.name,tags.#(type=="category")#.name]|@flatten'
       Performers:
-        Name: tags.#(type=="artist")#.name
+        Name: 'tags.#(type=="character")#.name'
       Studio:
-        Name: tags.#(type=="group").name
+        Name: '[tags.#(type=="artist")#,tags.#(type=="group")#]|@flatten.0.name'
```
feederbox826, line 51: why the dupe of group between photographer and artist?

feederbox826:
And yes, it looks like they did lift the APIkey limit.

> Auth: Public (optional User Token or API Key for personalization)

scrapers/nhentai.yml
spaceyuck: "artist" as well as "group" are both optional, and are used somewhat inconsistently on the site. The goal was to have the studio field always filled, with a preference on "artist" to get works by the same author grouped under the same "studio" name, if at all possible (since multi-work / multi-part series are somewhat of a regular thing).

The photographer is just to retain the group name at all if it doesn't make it into the "studio" field.

feederbox826 (approved):

feederbox826:
Thanks!

PR #2734, opened 2026-05-06
Fix SexLikeReal scraper to find studio name

tj2222:
I guess they changed an h3 to an h2.

Checked against https://www.sexlikereal.com/scenes/pmv-sit-get-fcked-sitting-cowgirls-compilation-85687

Also verified that the other selectors that include `h3`s still work as-is.


feederbox826:
thanks!

PR #2739, opened 2026-05-08
KMM scraper fixes and improvements for 8K/5K sites

tj2222:
**Sites impacted**: 5KPorn, 5Kteens, 8K Milfs, 8K Teens

-----

**Functional changes**:
* Makes details text extraction work on 8K URLs again.
* Uses the 8K scraper for 5K URLs, therefore...
  * 5K sites now get tags!
  * 5K sites now get male performers!
  * Scraping via 5K domains will no longer yield a three digit zero-padded episode number in the studio code, e.g. `5KP #015` would now scrape as `5KP #15`. IMO this is fine.
  * 5K and 8K studio scenes can be scraped from any 5K or 8K domain and will get correct studo names, codes, and URLs.
* Alias tag "Amazon" to "Amazon Position". There's [a thread](https://discourse.stashapp.cc/t/new-tag-for-amazon/7093) about adding an Amazon tag to StashDB, which would be a different concept from Amazon Position, but KMM uses Amazon to mean Amazon Position. This replacement nets us an additional tag and prevents mis-tagging in the future.

-----

**Implementation notes**:
* Deletes `5kSite` scraper.
* Pulls cover image from `<meta>` tag because 5K studio scenes don't have a `<video>` trailer element to pull from.
* There are a couple spots where it seems like I should be able to use `map` rather than `replace` with regexes (Amazon tag, Studio name), but it just didn't work when I tried it ¯\\\_(ツ)\_\/¯

### DIFF
--- a/scrapers/KellyMadisonMedia.yml
+++ b/scrapers/KellyMadisonMedia.yml
@@ -107,19 +107,21 @@ xPathScrapers:
                 with: $1
 
   8ksite:
+    common:
+      $url: //link[@rel="canonical"]/@href
     scene:
       Title: //strong[text()="Title:"]/following-sibling::a
       Code:
-        selector: //strong[text()="Title:"]/following-sibling::a/@href
+        selector: $url
         postProcess:
           - replace:
-              - regex: .+?(8\w+)/(\d+)
-                with: "$1 #$2"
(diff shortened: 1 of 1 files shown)


feederbox826:
Woo

PR #2741, opened 2026-05-09
[EmbeddedMetadata] fix keywords error and add test coverage

### LINKED ISSUE #2738: Can't import tags because `value_parts` variable is referenced before it's declared
### Scraper name

EmbeddedMetadata

### Scraper version

f66d1941

### Which scraper type(s) are affected?

imageByFragment

### Specific examples

The main script references `value_parts` under an `if else` statement, but the variable is only declared under the `if` statement, so the script aborts whenever it detects entries in any keyword related EXIF tag.

### Logs

```shell

```

### Additional details

I've submitted a draft pull request with my simple fix for the problem (#2737). It can probably be improved by someone who actually knows python, unlike me.

### Have you looked at existing bug reports?

- [x] I have searched the existing bug reports.

### Are you in a country that may be geoblocked by the site being scraped?

- [x] I am not in a country that may be geoblocked.

spaceyuck:
Fixes an error in the keywords to tags logic, based on report by @NoNameHeath in [the forum](https://discourse.stashapp.cc/t/embeddedmetadata/5091/2).

closes #2738
closes #2737

## Scraper type(s)

- [X] imageByFragment

## Examples to test

included in folder test-data

## Short description

Aside from the small fix, this also adds unit test coverage (which necessitated a python file rename, and some small structural changes to be able to force both supported source library modes).

Here's a question: is it OK to place unit tests and test data within the same directory as the scraper? I couldn't find any other example with a unit test.


feederbox826:
> also adds unit test coverage (which necessitated a python file rename

?

DogmaDragon:
> Here's a question: is it OK to place unit tests and test data within the same directory as the scraper? I couldn't find any other example with a unit test.

I see no value in it for the end user. I would exclude it.

spaceyuck:
> > also adds unit test coverage (which necessitated a python file rename
> 
> ?

python importable module names can't contain '-', which the old file name did

> I see no value in it for the end user. I would exclude it.

I have moved the tests and test data into a subfolder for the scraper under a new top folder `scraper-tests`, this should keep all tests code and test data away from the actual scrapers while still mirroring the directory structure, so associations remain clear

PR #2742, opened 2026-05-09
[GalleryDLInfo] gallery-dl downloaded metadata scraper

### LINKED ISSUE #2240: Gallery.dl integration / metadata with stashapp
Hello, I was wondering if there's a way to integrate the two together, since gallery.dl is a popular data hoarding software. Alternatively, the gallery.dl has an option for metadata downloads, is there a way to import them? the json/txt files have the same name as the images but contain all the metadata

spaceyuck:
## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [X] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL
- [X] galleryByFragment
- [ ] galleryByURL
- [X] imageByFragment
- [ ] imageByURL

## Examples to test

Includes a test suite using example data based on a few different structures encountered in the wild

## Short description

gallery-dl has a few interesting options (`--write-info-json` and `--write-metadata`), that produce per gallery and per file JSON metadata files while downloading. This scraper implements processing these files.

These could be very helpful, if the source is for whatever reason no longer scrapable with existing scrapers (cloudflare, API ban, content removed by uploader, ...).

There seems to be somewhat of a consensus on the structure, but there are a lot of site-specific differences, so it's once again a best effort approach, which may or may not work well for any of the hundreds of sites supported by gallery-dl.

closes #2240

This scraper includes unit tests and test data, right now these are placed right besides the scraper, I don't know if they should maybe go somewhere else.


scrapers/GalleryDLInfo/gallerydl_info.py
```diff
@@ -0,0 +1,354 @@
...
+                })
+        elif key.lower() == 'category':
+            match value:
+                case 'rule34' if context == 'image' or context == 'scene':
+                    ret['URLs'] = ['https://rule34.xxx/index.php?page=post&s=view&id=' + ret['Code']]
+                case 'nhentai' if context == 'gallery':
+                    ret['URLs'] = ['https://nhentai.net/g/' + ret['Code'] + '/']
+                case 'exhentai' if context == 'gallery' and 'token' in metadata:
+                    ret['URLs'] = ['https://e-hentai.org/g/' + ret['Code'] + '/' + metadata['token'] + '/']
+                case 'exhentai' if context == 'image' and 'image_token' in metadata and 'num' in metadata and 'gid' in metadata:
```
feederbox826: This is hardcoded to these websites. This and explicit references to `ehentai` and test data is more unsettling, as well as title_jp and title_jpn makes me think that it wasnt tested against gdl source code directly. That and test case explicitly referring to the  site, alongside recent DMCA takedowns makes this a very very difficult accept.

If you can prove that it is generic enough that it is only based on GDL and not specific sites, i'd be more open to it, but the references to sites as well put us into the position of downloaders, not just of metadata

spaceyuck: Hm, fair concern, with the kerfuffle around it a bit back, and these specific sites, didn't think of that.

If hard-coding sites is a problem, then the URL generation has to go, as these JSON files only contain site-specific identifiers and a site identifier, not the actual source URLs. And I can make the examples names more generic too no problem.

As for genericness, GDL only defines the basic mechanism for their site-specific extractors to populate these files. So there is no GDL metadata standard to implement, just a guess and hope for a consensus among implementers on their side kind of approach.

I'll try to make it less specific

spaceyuck:
OK, reworked some stuff, no more site-specific code paths or references.

But as already said, it's only as generic as far as there is a consensus between implementations on their side. Some fields have a few different guesses of what they might be named, but it remains a bit of whack-a-mole.

DogmaDragon:
I don't think including test and test data serves any purpose for end users.

spaceyuck:
> I don't think including test and test data serves any purpose for end users.

I have moved the tests and test data into a sub-folder for the scraper under a new top folder `scraper-tests`, this should keep all tests code and test data away from the actual scrapers while still mirroring the directory structure, so associations remain clear.

PR #2743, opened 2026-05-09
Sheer update

robbydurano:
This completely redoes the Sheer scraper.  The old one doesn't process dates properly.  Sometimes you get a scene that was posted 1 hour/s ago, and you'd have to manually fix it in stash.  Or you get a date that was March 31 and the current year was hardcoded, and you still have to manually update in stash.  This update fixes all that.

Additionally, the old scraper did not scrape all tags.  One test scene I ran extracted 14 tags originally, but updated scraper now gets 33 tags.

Old scraper would also miss image extraction constantly.


feederbox826:
Stash has a built-in relative date parser as well as goja. What can't be done in yml?

robbydurano:
i bet the date could be done in yml, but i mostly redid it to get all the tags

PR #2752, opened 2026-05-17
[FileMetadata] Parse json from stdin buffer instead of text to prevent mojibake

decisive72:
`sys.stdin` is a text-mode stream that decodes bytes using the locale encoding (e.g. CP1252 on Windows), which corrupts non-ASCII characters in scene data (e.g. file names with Japanese characters). Reading from `sys.stdin.buffer` instead retrieves raw bytes, and `json.loads()` always interprets these as UTF-8 regardless of platform or locale settings.


Maista6969:
Thank you, that's a good fix!

PR #2772, opened 2026-06-06
Add additional cookies for Kink sites

pupmaxx:
Adds additional age_gate_accepted cookies for other domains supported by this scraper. The cookie already exists for the base kink.com domain; however, this scraper declares support for the other domains but doesn't add this cookie.

### DIFF
--- a/scrapers/Kink.yml
+++ b/scrapers/Kink.yml
@@ -279,4 +279,12 @@ driver:
           Domain: "www.kink.com"
           Value: "1"
           Path: "/"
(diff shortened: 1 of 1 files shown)


feederbox826:
https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/KinkTrans.yml#L168
https://github.com/stashapp/CommunityScrapers/blob/master/scrapers/KinkMen.yml#L168

handled in their individual scrapers

PR #2776, opened 2026-06-10
[Pornhub] Use TLS impersonation

Larsluph:
Their API recently changed and now blocks all requests without impersonation with 403 Forbidden.

Add TLS impersonation according to recent change: stashapp/stash#6806

Tested and it does work after adding impersonation.


DogmaDragon:
`surf` driver is only available on develop build at the moment.

PR #2779, opened 2026-06-12
Fix: Ladyboy Gold network - matching scene from search result chosen from file cache; update IDs for API searches

nrg101:
## Scraper type(s)
- [x] sceneByName
- [x] sceneByQueryFragment
- [x] sceneByFragment

## Examples to test

- search by text: Soft Lactating Tits
- pick one of the results
- the studio should match your choice

## Short description

The LadyboyGold network sometimes releases the scenes at other sites/substudios, so there can be more than one search hit for the same scene with basically just the studio name differing in the search results shown in Stash.

If the API search result does not have a main website value, then the URL is generated based on which site API was used.

However it is created, the scene URL will be different for each search result, and therefore in the resulting fragment of the search result chosen, so that is now used to pick from the file cache results in the by query fragment part of the process.

Some of the IDs needed updating for the API searches to work, so that has been done now and each site can be searched for scenes.

### DIFF
--- a/scrapers/LadyboyGold/island_dollars.py
+++ b/scrapers/LadyboyGold/island_dollars.py
@@ -28,10 +28,10 @@ CONFIG = {
         }
     },
     "ladyboygold": {
-        "cms_area_id": "a1fbacce-2340-45ef-9576-129e766e63f9",
+        "cms_area_id": "cd9a5600-5cda-4ed0-b356-f62af1887d96", # from homepage call to /config.json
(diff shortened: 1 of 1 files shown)


nrg101:
I ended up needing the fixes above for improving the scraper further by adding `galleryByFragment`, i.e. photo set search by text, so I will close this PR when I open the other one

nrg101:
duplicates #2801

PR #2786, opened 2026-06-26
javdb: extract JAV code from filename in sceneByFragment

MayhemDoggo:
## Problem
`javdb.yml`'s `sceneByFragment` only strips the file extension, then passes the **raw filename** to javdb's search:

```yaml
queryURLReplace:
  filename:
    - regex: \..+$
      with: ""
```

javdb's search expects the **hyphenated** code (e.g. `DDT-246`), so any filename that isn't already exactly that fails to match. Common real-world variants that currently miss:

| filename token | should search |
|---|---|
| `DDT246` (dashless) | `DDT-246` |
| `DDT 246` / `DDT_246` | `DDT-246` |
| `[GTJ--010]` (double dash) | `GTJ-010` |
| `JAV24 ... DDT284` (junk first) | `DDT-284` |
| `DDT-246 ... -CD1` (trailing disc token) | `DDT-246` |

## Fix
Two added `queryURLReplace` steps:

```yaml
- regex: (?i)\b(CD|DVD|DISC|PART|JAV)[-_ ]?\d{1,3}\b   # drop disc/site junk tokens
  with: " "
- regex: ^.*?([A-Za-z]{2,6})[-_ ]*(\d{2,5}).*$          # extract first code, normalize
  with: $1-$2
```

- Pristine `DDT-246` → unchanged
- The junk list is conservative (none collide with real JAV label prefixes)
- A matching fix for `JavBus.yml` (same class of filenames) will follow in a separate PR.

Verified against dashless, double-dash, junk-prefix, and trailing-disc-token filenames — all now resolve to the correct code.

### DIFF
--- a/scrapers/javdb.yml
+++ b/scrapers/javdb.yml
@@ -6,6 +6,15 @@ sceneByFragment:
     filename:
       - regex: \..+$
         with: ""
+      # Drop disc/site junk tokens that look like codes (CD1, DISC2, JAV24, ...) so the real
+      # code is picked even when such junk appears before it in the filename.
+      - regex: (?i)\b(CD|DVD|DISC|PART|JAV)[-_ ]?\d{1,3}\b
+        with: " "
+      # Extract the first LETTERS-DIGITS code and normalize it: dashless (DDT246),
+      # spaced/underscored, and double-dash ([GTJ--010]) filenames all become the
+      # hyphenated form (e.g. DDT-246) that javdb's search requires.
+      - regex: ^.*?([A-Za-z]{2,6})[-_ ]*(\d{2,5}).*$
+        with: $1-$2
   scraper: sceneQueryScraper
 sceneByURL:
   - action: scrapeXPath
--- a/scrapers/javdb.yml
+++ b/scrapers/javdb.yml
@@ -10,11 +10,17 @@ sceneByFragment:
       # code is picked even when such junk appears before it in the filename.
       - regex: (?i)\b(CD|DVD|DISC|PART|JAV)[-_ ]?\d{1,3}\b
         with: " "
-      # Extract the first LETTERS-DIGITS code and normalize it: dashless (DDT246),
-      # spaced/underscored, and double-dash ([GTJ--010]) filenames all become the
-      # hyphenated form (e.g. DDT-246) that javdb's search requires.
-      - regex: ^.*?([A-Za-z]{2,6})[-_ ]*(\d{2,5}).*$
-        with: $1-$2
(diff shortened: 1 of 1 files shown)


javstash:
Please test for the following edge cases:

`25ID-051`
`T38-065`
`3DSVR-1968`
`MBR-AA048`

MayhemDoggo:
Thanks — those four are exactly the cases the original single-pass regex got wrong. Confirmed all four against Go's `regexp` (RE2, the same engine stash uses), then reworked the extractor and re-verified.

Root cause: `^.*?([A-Za-z]{2,6})[-_ ]*(\d{2,5}).*$` requires the label to be **2–6 letters only**, so it silently dropped any digits in the label and couldn't see a letter-prefixed number:

| input | old result | new result |
|---|---|---|
| `25ID-051` | `ID-051` ❌ | `25ID-051` ✅ |
| `T38-065` | `T38-065` (passed by luck) | `T38-065` ✅ |
| `3DSVR-1968` | `DSVR-1968` ❌ | `3DSVR-1968` ✅ |
| `MBR-AA048` | `AA-048` ❌ | `MBR-AA048` ✅ |

The fix replaces the one regex with three ordered branches that pick the leftmost real code:

- **1a** `label + dash/underscore + number` — number may carry a short letter prefix (`MBR-AA048`)
- **1b** `label + space + number` — number must be pure digits, so a junk word after the label isn't swallowed
- **2** dashless `label` immediately followed by digits (`DDT246` → `DDT-246`)

`label` now matches letter-then-digits (`T38`) or optional-leading-digits + ≥2 letters (`25ID`, `3DSVR`). The "≥2 letters" rule also keeps resolution tokens like `720p`/`1080p` from being mistaken for a code.

Full regression matrix (all green), including the original PR cases:

| input | output |
|---|---|
| `DDT-246` / `DDT246` / `DDT 246` / `DDT_246` | `DDT-246` |
| `[GTJ--010]` | `GTJ-010` |
| `JAV24 foo DDT284` | `DDT-284` |
| `DDT-246-CD1` | `DDT-246` |
| `720p-DDT246` | `DDT-246` |
| `[2024] SSIS-698 [1080p]` | `SSIS-698` |
| `25ID-051` / `T38-065` / `3DSVR-1968` / `MBR-AA048` | unchanged ✅ |

Known out-of-scope limitation (pre-existing — the old regex behaves identically): leech-site domain prefixes like `hhd800.com@SSIS-698` still resolve to `hhd-800`, since `hhd800` is the leftmost code-shaped token. Stripping known site domains is a separate change I'm happy to follow up with if you'd like it in this scraper.

Pushed as `af22313`.

PR #2790, opened 2026-07-04
Adds a scraper for GloryHole Swallow and Cumpsters logged-in scene pages

tj2222:
## Scraper type(s)
- [X] sceneByURL

## Examples

[GloryHole Swallow](https://stashdb.org/edits/019f2583-e553-7da6-85d7-5418c53d1879), [Cumpsters](https://stashdb.org/edits/019f257e-2008-73a4-9c81-eb903e1d71fd)

## Short description

This Python scraper is in addition to the existing XML-based scraper. The latter continues to work for logged-out pages (note that URL matching for that one has been narrowed down a bit so as to not try to handle `/members` URLs). Scraping a members' URL with the new scraper will always yield additional data, but requires an active membership cookie, to be put in `config.ini`. See the long comment near the top of `GloryHoleSwallowMembers.py` for details of what additional data is scraped.

## LLM use

Written with some assistance from Antigravity / Gemini 3.5 Flash (High). I have read and understand all of the code, and wrote a decent amount of it "by hand".

### DIFF
--- a/scrapers/GloryHoleSwallowMembers/GloryHoleSwallowMembers.py
+++ b/scrapers/GloryHoleSwallowMembers/GloryHoleSwallowMembers.py
@@ -71,15 +71,6 @@ def get_cookie_name(domain: str) -> str:
     b64_name = base64.b64encode(name.encode('utf-8')).decode('utf-8')
     return f"pcar%5f{b64_name.replace('=', '%3d')}"
 
-
-def has_valid_cookie(url: str) -> bool:
-    domain = urlparse(url).netloc.lower()
-    cookie_val = get_pcar_cookie(domain)
-    if cookie_val:
-        return True
-    return False
-
-
 def get_presumed_public_url(url: str) -> str:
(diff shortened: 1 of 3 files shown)


scrapers/GloryHoleSwallowMembers/GloryHoleSwallowMembers.py
```diff
@@ -0,0 +1,374 @@
...
+                    is_estimate = True
+            except ValueError:
+                pass
+    
+    return (release_date, is_estimate)
+
+
+def get_details(tree) -> str:
+    paragraphs = tree.xpath("//div[@class='objectInfo']/div[@class='content']/p")
+    if paragraphs:
```
feederbox826, line 190: no return if no details_parts?

feederbox826:
LGTM at a glance, unable to validate

scrapers/GloryHoleSwallowMembers/GloryHoleSwallowMembers.py
tj2222: Failing the check on line 192 results in returning empty string on line 194

PR #2791, opened 2026-07-04
Add DMM doujin gallery scraper

korafu:

## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [ ] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [ ] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [x] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples to test

https://www.dmm.co.jp/dc/doujin/-/detail/=/cid=d_760708/

## Short description

This PR adds a gallery scraper specifically for DMM doujin. I made the changes in the existing DMM scraper and used CDP as there is an age verification and login wall for DMM, and it should resolve the issue for the existing scene scraper as well. But if it is too messy I can separate it out into its own scraper

### DIFF
--- a/scrapers/DMM.yml
+++ b/scrapers/DMM.yml
@@ -5,21 +5,21 @@ sceneByFragment:
   queryURLReplace:
     filename:
       - regex: \..+$
-        with: ""
+        with: ''
   scraper: sceneQueryScraper
 sceneByURL:
   - action: scrapeXPath
     url:
       - dmm.co.jp/
     scraper: sceneScraper
-    queryURL: "{url}"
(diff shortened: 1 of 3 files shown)


Maista6969:
Because Stash applies the CDP-setting across the entire scraper definition I'd prefer if this was split into its own scraper so current users do not need to set up CDP to keep using this

korafu:
@Maista6969 done, I added a directory to house the 2 scrapers

PR #2857, opened 2026-08-30
feat(CockyBoys): add search, scene code and full-size images

stashdbcorrode248:
Adds search, scene Code and full-size images to the CockyBoys scraper, converting it to Python and moving it into `scrapers/CockyBoys/`.

### Why this is not XPath

CONTRIBUTING prefers XPath, so this needs justifying. **The existing XPath scraper is not broken** - I checked every selector against the live site and all of them still match. It was replaced because search cannot be made to work in XPath on the current scraper schema, not because URL scraping needed fixing.

`sceneByName` does not accept `queryURLReplace`; it is not a field on `ByNameDefinition`, and Stash refuses to load the scraper if you add it:

```
Error loading scraper CockyBoys.yml: yaml: unmarshal errors:
  line 14: field queryURLReplace not found in type scraper.ByNameDefinition
```

Without query rewriting the site search is unusable, because the queries Stash sends never match verbatim:

- **Filename-derived queries** carry a typographic apostrophe, which the site does not match, plus a trailing `- Performer A & Performer B` that is not part of any title. `Max's Revenge on Tommy's Ass` returns 1 result; the same string with typographic apostrophes and the performer suffix returns 0.
- **The tagger** sends `<date> <studio> <performers> <title> <performers>`, which matches nothing verbatim. The date and studio name have to be stripped and shorter fragments tried before there is a hit.

Two further limits rule out XPath even if the query could be rewritten:

- A `queryURL` fetches one page, but the site paginates at 24 results. A common performer name buries the wanted scene on page 2 or later. The tagger example above resolves to a scene that is on page 2.
- Once several pages are collected they need ranking against the original query, which has no XPath equivalent.

### What changed

Everything the XPath version scraped is preserved. Added:

- `sceneByName`, `sceneByFragment` and `sceneByQueryFragment`, with query normalisation, pagination up to 5 pages, and results ranked by overlap with the original query.
- `Code`, taken from the `setid` in the player configuration.
- Full-size images. `og:image` is a 475x350 thumbnail; the CDN also serves a `-full` variant that is 1800x1200 for most scenes and the same size for the rest, so it is never worse. The URL is checked before use and falls back to the thumbnail.

Requires `requests` and `beautifulsoup4`, declared in `requirements.txt`.

### Testing

Run against Stash 0.31.1. URL scraping returns all previous fields plus `Code`. Search was checked with a plain performer name, a filename with typographic apostrophes and a performer suffix, and a tagger string that previously returned nothing and now ranks the correct scene first out of 26.


stashdbcorrode248:
Closing while I rework this; will reopen once it is ready.

gimmeliina:
Scene https://cockyboys.com/scenes/kai-reynolds-oliver-carter.html?type=vids was the last one with proper images; that worked best with 4x version, but new scenes doesn't have that version anymore. `-full` gives slightly different img for those and if i remember right, older scenes (before 2020) doesnt have `-full` at all...

That's why i used og:image

PR #2859, opened 2026-08-31
Add Bait Buddies scraper

he69man69:
## Summary
- Adds a scene and performer scraper for Bait Buddies (baitbuddies.com)

## Scraper type(s)
- [ ] performerByName
- [ ] performerByFragment
- [x] performerByURL
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL
- [ ] groupByURL
- [ ] galleryByFragment
- [ ] galleryByURL
- [ ] imageByFragment
- [ ] imageByURL

## Examples used in testing:
- https://www.baitbuddies.com/video/bb980-m-adam-scott-mp4-1998.html
- https://www.baitbuddies.com/video/bb985-m-rio-riley-mp4-1997.html
- https://www.baitbuddies.com/video/337-liam-toby-final-mp4-325.html
- https://www.baitbuddies.com/video/bb1013-m-beau-marcus-2072.html
- https://www.baitbuddies.com/models/scott-demarco-668.html
- https://www.baitbuddies.com/models/owen-429.html

## Notes
- Full rationale here: https://discourse.stashapp.cc/t/baitbuddies-com-scraper-proposal/15304
- Written as a Python script scraper because Date isn't present on the scene page. It only shows up on a performer's profile page, next to each of their scenes. The scraper fetches that second page (the first credited performer's profile) and matches the row linking back by the scene's numeric id.
- The site has no real scene titles, so Title is built from the credited performers, in the order the video's own internal filename lists them, and joined with "and" rather than "&", to stay consistent with GEVI-style scrapes.
- Sequel scenes with the same cast (flagged inconsistently in the site's own description like "Part/Round 2 of Tom and Jack!") get that appended to the title so otherwise-identical titles stay distinguishable. It's a deduping by name as with other studios like Falcon who release multiple scenes under one title. It helps avoid StashDB fingerprint mismatches, and this studio has a few cases of two different scenes with identical casts getting their fingerprints cross-contaminated.
- Image uses the video player's poster frame (`.../media/thumbs/custom_thumbs/{id}.jpg`) rather than the tiny `og:image` thumbnail. Same as the usual standard for this studio's images on StashDB.
- The "straight" tag is blacklisted, every scene on this site carries it, describing the performer rather than the scene.
- Performer scrape pulls Height/Weight/HairColor/EyeColor/PenisLength/Circumcised off each model's profile page.

### DIFF
--- a/scrapers/BaitBuddies/BaitBuddies.py
+++ b/scrapers/BaitBuddies/BaitBuddies.py
@@ -1,0 +1,284 @@
+import json
+import re
+import urllib.parse
+import urllib.request
+
+from py_common import log
+from py_common.deps import ensure_requirements
+from py_common.util import feet_to_cm, lb_to_kg, scraper_args
+
+ensure_requirements("lxml")
+from lxml import html  # noqa: E402
+
+HAIR_COLORS = {
+    "blond": "Blonde",
+    "blonde": "Blonde",
+    "brown": "Brown",
+    "brunette": "Brown",
+    "black": "Black",
+    "red": "Red",
+    "auburn": "Auburn",
+    "grey": "Grey",
+    "gray": "Grey",
+    "bald": "Bald",
+}
+
+EYE_COLORS = {
+    "blue": "Blue",
+    "brown": "Brown",
+    "green": "Green",
+    "grey": "Grey",
+    "gray": "Grey",
+    "hazel": "Hazel",
+    "red": "Red",
+}
+
+# Removing conflict with existing StashDB "Straight" tag, which don't apply
+# to these scenes.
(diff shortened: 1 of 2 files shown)


he69man69:
One extra note (when in Python territory): I've got the secondary lookup on GEVI, WAYBIG and IAFD to pull those link sources in too (but not using their metadata, just attaching the links). If that's allowed within the scope of a scraper, I can commit that too.

Illustration of scrape results, where the scene link provides three other links:
Screenshot 2026-09-01 at 09 12 30

Overall, adds 3-5 seconds to the overall process, but adds the benefit of more metadata sources without having to go look for them.

PR #2863, opened 2026-09-03
Affect3DStore: use python cloudscraper with proxies to circumvent Cloudflare

nrg101:
## Scraper type(s)
- [ ] sceneByName
- [ ] sceneByQueryFragment
- [ ] sceneByFragment
- [x] sceneByURL

### sceneByURL

- https://affect3dstore.com/aveline-s-quest-episode-1
- https://affect3dstore.com/aveline-s-quest-episode-2
- https://affect3dstore.com/girlfriends-4ever-dlc-02.html

### sceneByName / sceneByFragment / sceneByQueryFragment

TBC

## Short description

The site now has Cloudflare protection. It is possible to use [cloudscraper](https://github.com/venomous/cloudscraper) to bypass Cloudflare protection. This has worked well for other scrapers when used together with [free-proxy](https://pypi.org/project/free-proxy/), so is used here too.

There is a text search on the site, so sceneByName and sceneByFragment/sceneByQueryFragment are probably possible too (so I'll look into that)


Maista6969:
I'll merge this, but keep in mind that Stash v0.32 is adding support for TLS impersonation that makes it possible to keep using XPath for this :slightly_smiling_face:

nrg101:
> I'll merge this, but keep in mind that Stash v0.32 is adding support for TLS impersonation that makes it possible to keep using XPath for this 🙂

It will be nice to use a pure xPath scraper again if it can work with Cloudflare-protected sites.

I tried to use the xPath selectors in the python code as much as possible, so it should be pretty easy to bring it back over to an xPath scraper (when Stash can work with Cloudflare protected sites)

PR #2866, opened 2026-09-03
Reimplement Falcon Studios Scene number logic for Altwolia, 

gimmeliina:
Reimplement old scene numbering from algoliaAPI (see: #2682 )
## Scraper type(s)
- [X] sceneByFragment
- [X] sceneByURL

## Examples to test

https://www.falconstudios.com/en/video/falconstudios/Turn-It-Up/67578

## Short description

Falcon studios (And HotHouse, raging stallion etc) uses movie title for scenes leading identical titles. This PR brings back `,Scene X` addition to title.

### DIFF
--- a/scrapers/AdultTime/AdultTime.yml
+++ b/scrapers/AdultTime/AdultTime.yml
@@ -11,13 +11,18 @@ sceneByURL:
       - adulttimepilots.com/en/video/
       - agentredgirl.com/en/video/
       - ageandbeauty.com/en/video/
+      - alettaoceanempire.com/en/video/
       - allgirlmassage.com/en/video/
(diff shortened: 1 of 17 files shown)


Maista6969:
Whoops, I apologize for breaking this in the migration! I've recreated your PR with a slight architectural change (I prefer that Altwolia only contain generic logic that apply to all Algolia sites) and will close this PR in favor of 95908b6c

gimmeliina:
Yes, I thought that it will be easier to maintain if there is no duplicated code here and there :D
(First I thought arg based approach is the best, but as there is studio specific `.py` files I come up with this.)

gimmeliina:
@Maista6969 #2872

PR #2881, opened 2026-09-14
Fix BeefcakeHunter Details truncation on malformed scene markup

he69man69:
## Summary
- Fixes BeefcakeHunter's `Details` scraper silently dropping the opening paragraph(s) on scenes where the site emits malformed markup (text left unwrapped in `<p>`, followed by a stray closing `</p>`)

## Examples used in testing
- [x] Verified scene scraping against:
  - https://beefcakehunter.com/video/playing-with-beefcake-barron/ (the malformed-markup case; Details now includes the full opening text)

## Notes
- Root cause: the site occasionally leaves the opening paragraph(s) as bare text/inline markup inside the content `<div>` instead of wrapping them in `<p>`, with a stray unmatched `</p>` where the real `<p>` paragraphs start. `//p` can never match text that was never inside a `<p>` element.
- Fix selects the whole wrapping `<div>`'s text content instead of only `<p>` descendants, so it's resilient regardless of whether a given paragraph is `<p>`-wrapped. A `postProcess` regex restores the `\n\n` paragraph breaks previously provided by `concat`.


Maista6969:
Thanks for the PR! I noticed that this version is mangling descriptions in its own way, mostly because the regex replace `\n+` is a no-op in every case: Stash strips the newlines out of a text block before postprocessing ever runs.

Due to how inconsistent this site is I do not think we can solve this without resorting to Javascript. Can you find any example scenes where the following does not work? I do not have a lot of URLs to test across :sweat_smile: 

```yml
        selector: >
          //div[@id="dslc-theme-content-inner"]
          |//div[@id="dslc-theme-content-inner"]//p
          |//div[@id="dslc-theme-content-inner"]//br/following-sibling::node()[1]
        concat: "\x01"
        postProcess:
          - javascript: |
              if (!value) return "";
              var parts = value.split("\x01");
              var full = "";
              for (var i = 0; i < parts.length; i++) {
                if (parts[i].length > full.length) full = parts[i];
              }
              var cuts = [0];
              for (var i = 0; i < parts.length; i++) {
                var anchor = parts[i].trim();
                if (!anchor || anchor === full) continue;
                var at = full.indexOf(anchor);
                if (at > 0 && at === full.lastIndexOf(anchor)) cuts.push(at);
              }
              cuts.sort(function (a, b) { return a - b; });
              var out = [];
              for (var i = 0; i < cuts.length; i++) {
                var end = i + 1 < cuts.length ? cuts[i + 1] : full.length;
                var block = full.slice(cuts[i], end).trim();
                if (block) out.push(block);
              }
              return out.join("\n\n");
```

he69man69:
Thanks @Maista6969! I'll throw a bunch into testing with the above fix as soon as I get back to a computer. I'll compare scrape results line by line and report back soon 😁

he69man69:
Yeah, that works much better @Maista6969, thank you! Tested with two scenes per year and compared the results over the different styling, and they call came back 1-to-1. Added that in now 🙏

PR #2887, opened 2026-09-18
Implement a scene URL scraper for cathyscraving.com

tj2222:
## Summary
* Does what it says on the tin, basically.
* Requires membership cookie. Includes instructions for placing cookies in a `config.ini` file.
* Subjective decisions (open to pushback on these):
  * Use 'Scene 1234: Full Title Here' for scene titles. Technically, this exact string may never occur on the site. This is because the scene list pages show 'Scene 1234: Abbreviated Title Here' and scene pages show 'Scene 1234`\n`Full Title Here', with the scene number portion in a different color than the rest.
  * Use the large top images on scene pages as cover images. There are smaller thumbnails that are used on the scene list pages and shown just above the file download links on the scene page, which arguably are at least as canonical as the large top image. However, 1) The smaller thumbnails are significantly lower resolution; 2) Subjectively, I think that the semantic quality of the large and thumbnail images are equal on average (e.g. how well they convey the scene content, how aesthetically appealing they are); 3) TPDB uses this image strategy as well, so there's some precedent.

## Example scrapes (NSFW)
[Example 1](https://stashdb.org/edits/01a0a8fa-9991-773c-aeb5-c3c3da06913f)
[Example 2](https://stashdb.org/edits/01a0a8f8-5955-7194-8d03-e76df9423ea1)
[Example 3](https://stashdb.org/edits/01a0a8ef-638f-7818-b73a-e102db8bcdb3)

## LLM usage disclosure
This was written in the Google Antigravity IDE. I primarily used the ExploitedTeens community scraper as a reference to base this off of, but did get some LLM help as well. LLM help came in the forms of: tab completion for some logging/xpath/date-formatting-string completion, "agentic" (am I using that right?) discussions about xpath functions and how to implement `extract_text_with_linebreaks()`. I have completely read and understand all of the code :)


Maista6969:
Thanks for the contribution! I can't test it since I don't have a subscription, but the scenes you've submitted seem good :slightly_smiling_face:

